PR #2, opened 2015-01-23
Support finding QGIS in QGIS_PREFIX_PATH too

timlinux:


m-kuhn:
How/Where is this env-var set?

timlinux:
QGIS_PREFIX_PATH is recognised by QGIS (see https://github.com/qgis/QGIS/blob/master/src/core/qgsapplication.cpp#L108). The user would be responsible to set this. In our work we use a small bash script to set it e.g. https://github.com/AIFDR/inasafe/blob/develop/run-env-linux.sh so that we can test aganist multiple installs of QGIS

m-kuhn:
So it's a runtime variable?

It won't be possible to run QField against multiple versions of QGIS (unless we convince everybody that ABI compatibility is important...)

So alternatively this could be a CMake variable to make it easier to configure cmake. Or do you think that it also makes sense in this scenario to use the env var?

timlinux:
Yeah adding it as a CMAKE var would probably be nice...I'll tweak it

m-kuhn:
Merged in https://github.com/opengisch/QField/commit/46303d56ce5cfb890707bf23896ed22fcfb26f77 with a CMAKE variable.
Thank you very much

timlinux:
Cool thanks! I'll push the same change the the upstream cmake file in QGIS too.

PR #26, opened 2016-01-27
Insert missing spaces into Welcome Screen message

riggsd:
This is a tiny fix that will improve users' first impressions of the app. Bug caused by literal string concatenation without whitespace.


m-kuhn:
Thanks a lot!

PR #27, opened 2016-02-02
changed URL to submodule repositories

rigid:
Using public https:// instead of private git@ enables everyone to clone --recursive / submodule update. Even if they have configured a public key for github but don't have that key registered to the submodules.
Previously, "git submodule update" would try to authenticate with my public key and deny access.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "test/QGIS-Sampledata"]
 	path = test/QGIS-Sampledata
-	url = git@github.com:opengisch/QGIS-Sampledata.git
+	url = https://github.com/opengisch/QGIS-Sampledata.git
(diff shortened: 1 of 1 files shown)


m-kuhn:
The downside of this is that it makes pushing to submodules harder...

rigid:
The docs are quite clear on this:

> Since the URL in the .gitmodules file is what other people will first try to clone/fetch from, make sure to use a URL that they can access if possible. For example, if you use a different URL to push to than others would to pull from, use the one that others have access to. You can overwrite this value locally with git config submodule.DbConnector.url PRIVATE_URL for your own use. When applicable, a relative URL can be helpful.

from https://git-scm.com/book/en/v2/Git-Tools-Submodules

m-kuhn:
Sounds good, thanks

PR #174, opened 2017-10-11
buildable with qgis gui lib/includes

PeterPetrik:
I have QGIS built with CMake flag "-DWITH_GUI=FALSE", see https://github.com/qgis/QGIS/pull/5338, but I am unable to build QField with such QGIS build.


m-kuhn:
Thanks!

PR #196, opened 2017-12-04
fix QgsDistanceArea::measureLine using QVector rather than QList

3nids:


m-kuhn:
@signedav does this look good to you? can you approve?

signedav:
Yes. I got the same and it works perfectly.

PR #198, opened 2017-12-20
QgsDatumTransform no longer needs to be initialized

m-kuhn:
Read: the method has been removed.


m-kuhn:
no idea, you tell me 😀

3nids:
I don't really see where it is use from, but I think it needs to receive a context to work properly.
can you do `setContext(mProject->transformContext())` in place of the old initialize ?

PR #205, opened 2018-01-11
multilevel visibilityexpressions concatinate

### LINKED ISSUE #125: Expression controlled visibility
When using drag'n'drop designer to build up a form QGIS Desktop also offers tabs and group containers whose visibility can be controlled by expressions. This somehow works also on QField but as soon as there is more than one container the expressions defining the visibility are somehow mixed up.

### How it works perfectly fine on QGIS 2.18.6

This video quickly shows how the form should work (mainly second part of the video):
https://youtu.be/K7Bca1u_b1k

### How QField fails
This video shows the same project fails to behave the same way on QField. The project is set offline by QFieldSync but the same behaviour is also observed for "online" project (directly copy'n'pasting of QGIS-Project and data to QField):
https://youtu.be/ZedWQGd-VTI

**QField version:** [0.9.10]

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: [I am not sure, but I never noticed it on previous versions]
* Problem happens with all files and projects, not only some files or projects: [Yes, we noticed it already with another project

Here you find the data and project in Case you want to check out the settings in detail.
[steinschmaetzer.zip](https://github.com/opengisch/QField/files/925176/steinschmaetzer.zip)

signedav:
instead of concatinating all visibilityexpression of the same level in attributeform what leaded misbehaviour

Fix #125

### DIFF
--- a/src/attributeformmodelbase.cpp
+++ b/src/attributeformmodelbase.cpp
@@ -214,16 +214,15 @@ void AttributeFormModelBase::updateAttributeValue( QStandardItem* item )
   }
 }
 
-void AttributeFormModelBase::flatten( QgsAttributeEditorContainer* container, QStandardItem* parent, const QString& visibilityExpressions, QVector<QStandardItem*>& items )
+void AttributeFormModelBase::flatten( QgsAttributeEditorContainer* container, QStandardItem* parent, const QString& parentVisibilityExpressions, QVector<QStandardItem*>& items )
 {
-  QString visibilityExpression = visibilityExpressions;
-
   Q_FOREACH( QgsAttributeEditorElement* element, container->children() )
   {
     switch ( element->type() )
     {
       case QgsAttributeEditorElement::AeTypeContainer:
       {
+        QString visibilityExpression = parentVisibilityExpressions;
         QgsAttributeEditorContainer* container = static_cast<QgsAttributeEditorContainer*>( element );
         if ( container->visibilityExpression().enabled() )
         {
--- a/src/attributeformmodelbase.h
+++ b/src/attributeformmodelbase.h
@@ -75,7 +75,7 @@ class AttributeFormModelBase : public QStandardItemModel
 
     void updateAttributeValue( QStandardItem* item );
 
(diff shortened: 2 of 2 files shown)


src/attributeformmodelbase.cpp
```diff
@@ -214,16 +214,15 @@ void AttributeFormModelBase::updateAttributeValue( QStandardItem* item )
...
   Q_FOREACH( QgsAttributeEditorElement* element, container->children() )
   {
     switch ( element->type() )
     {
       case QgsAttributeEditorElement::AeTypeContainer:
       {
+        QString visibilityExpression = parentVisibilityExpressions;
         QgsAttributeEditorContainer* container = static_cast<QgsAttributeEditorContainer*>( element );
         if ( container->visibilityExpression().enabled() )
         {
```
3nids: where is `visibilityExpression` initialized then? (line 229)

src/attributeformmodelbase.cpp
```diff
@@ -214,16 +214,15 @@ void AttributeFormModelBase::updateAttributeValue( QStandardItem* item )
   }
 }
 
-void AttributeFormModelBase::flatten( QgsAttributeEditorContainer* container, QStandardItem* parent, const QString& visibilityExpressions, QVector<QStandardItem*>& items )
```
3nids: would need to rename the arg in the header too

signedav: Yes. Thanks.

src/attributeformmodelbase.cpp
signedav: With the `parentVisibilityExpression` - and this is initialized on [call of the function](https://github.com/opengisch/QField/pull/205/files/de6c6738911ce289f44ec899cb5cb16c10e2e4ce#diff-e41b3ba57e866b91157e3b015593a8caL158)
Can't we do that?

signedav:
I tested that successfully in the debugging mode on the computer.

signedav:
Thanks @3nids

Issue #87, opened 2016-12-07
Constraint across many fields not updated on changes in other fields


saemiw:
### Observed behavior:
Example: I have 3 fields (e.g. f1, f2, f3) and define for field 3 a constrain like this in QGIS ("f1"+"f2"+"f3")=100 in order to check if the 3 fields sum up to 100%. If entering a new feature in QGIS Desktop the constraint is checked (as expected) as soon as I type in a number in any one of the 3 fields. On QField on the other hand, the constraint is only checked if I type in a number in field 3.

### Expected behavior:

The constraint should continously be validated (upon every touch on the screen).

**QField version:** [0.8.8]

## Additional information:

* Problem can be reliably reproduced, doesn't happen randomly: [Yes]

saemiw:
2 videos illustrating the above explained:

Constraints across many fields not updated satisfactory on QField 0.10.1: 
https://youtu.be/2odN7VPwhFE

Constraints across many fields not updated satisfactory on QGIS 2.18.13: 
https://youtu.be/atLOqJNIPh4

m-kuhn:
> When implementing this, can you make sure that it also works when the layers are set to offline-editing (i.e. go through the qfieldsync plugin).

Are you currently experiencing extra troubles with constraints in combination with offline editing? If yes, can you provide more information?

saemiw:
No at the moment I don't  experience extra problems (used to be a problem in early versions). Just a reminder (also to myself) to check this once its implemented.

signedav:
Fixed in Version 0.10.5 (in Playstore for Beta-Testers or [here](https://github.com/opengisch/QField/releases/tag/release-0.10.8_5))

saemiw:
I can confirm that this is fixed and works nicely unter 0.10.11. Thanks.

PR #207, opened 2018-01-15
Evaluate all the field constraints on every change

signedav:
On every change of any field ALL constraints had to be evaluated. Because the constraint can depend on different fields. 

This is a fix of https://github.com/opengisch/QField/issues/87

And now it looks like this (constraints on every field is: "EinsS"+"ZweiS"+"DreiS"=100):
qfield_constraints_fixed

### DIFF
--- a/src/attributeformmodelbase.cpp
+++ b/src/attributeformmodelbase.cpp
@@ -317,16 +317,13 @@ void AttributeFormModelBase::updateVisibility( int fieldIndex )
   for ( ; constraintIterator != mConstraints.constEnd(); ++constraintIterator )
   {
     QStandardItem* item = constraintIterator.key();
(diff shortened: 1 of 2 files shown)


gaby-h:
I tested this issue (also related to issue no 87). A new point can only be saved wenn the values respect the constraint set (in this example "p1"+ "p2"+"p3" = 100).
BUT for an exiting point, it remains possible to save changes afterwards which don't respect the constraint! It's only shown visualy (column-names appear red instead of black), that the constraint isn't respected, but it's possible to save changes (f.ex. p1 =50; p2 = 50 and p3 =20 -> p1 + p2 + p3 ≠ 100):
grafik
Test project and gpkg: 
[constraints.zip](https://github.com/opengisch/QField/files/1716705/constraints.zip)

signedav:
Thanks @gaby-h We will care about.

PR #208, opened 2018-01-15
Evaluate all the field constraints on every change

signedav:
On every change of any field ALL constraints had to be evaluated. because the constraint can depend on different fields. So I removed the check to evaluate only on the current field.

It's a fix for #87 

It looks now like this (Contraint is: "EinsS"+"ZweiS"+"DreiS"=100)
qfield_constraints_fixed


signedav:
Thanks @3nids

PR #212, opened 2018-01-17
Open the valuemap data as QList

signedav:
Instead of QMap because it's stored since 3.0 as QList in config('map').

Means it needed another loop to get all the data for the checkbox and some changes because the structure is slightly different. 

This change only makes sense when this is merged: https://github.com/qgis/QGIS/pull/6091

Possible that we could make some parts more elegant. I was not sure about everything but it works now perfectly with QGIS 2.18 and QGIS 3.00 projects.
qfield valuemap


signedav:
Thanks @3nids

Issue #199, opened 2017-12-25
Categorized Renderer layer crash QField


giedrioks26:
I'm trying to expand the layer with Categorized Renderer style layer properties, QField crashes.
QField 0.10.4
Android 7.0, 5.0
Device: Blackview BV6000s, sony xperia tablet z1

previous qfield version (0.10.1) works fine

m-kuhn:
What is e "expand"?

giedrioks26:
> What is e "expand"?

that's it drop down categorized layer
On QGIS called "expand"
![ekranvaizdis_2017-12-25_10-39-05](https://user-images.githubusercontent.com/6436791/34336552-bc442088-e960-11e7-8e1c-87fa62f7790c.png)

m-kuhn:
Can you extend a minimal sample project? E.g. the one here (no need to include rastras)

giedrioks26:
[categorized.zip](https://github.com/opengisch/QField/files/1585813/categorized.zip)

giedrioks26:
In QField double touch on layer

gaby-h:
version 0.9.10, displayind categorized symbols also works, no crash

giedrioks26:
I think this is related to the new feature in version 0.10.4 - "Change layer visibility on the fly"

signedav:
Fixed in Version 0.10.5 (in Playstore for Beta-Testers or [here](https://github.com/opengisch/QField/releases/tag/release-0.10.8_5))

giedrioks26:
works perfectly. Thank you!

Issue #181, opened 2017-10-25
Feature count causes hidden symbol in legend


saemiw:
If you check the "Count Feature" Option in QGIS and then transfer this project to qfield, qfield will not show the symbology in the table of content. The following images illustrate the problem:

**Situation without Feature count:**
on QGIS:
![qgis without feature count](https://user-images.githubusercontent.com/20398621/31986435-c4ffc86c-b968-11e7-85dc-509ada6cebc7.PNG)

on QField
![qfield without feature count](https://user-images.githubusercontent.com/20398621/31986438-c8ee354e-b968-11e7-86ac-9c1441d1ad28.png)

**Situation with Feature count for layer radiotagging:**
on QGIS
![qgis with feature count](https://user-images.githubusercontent.com/20398621/31986449-d72219c8-b968-11e7-9794-ba0865e98bd9.PNG)

on QField
![qfield with feature count](https://user-images.githubusercontent.com/20398621/31986453-db36cd06-b968-11e7-9b93-7ae70e52362c.png)

And here the project if you want to check yourself:
[qgis-project.zip](https://github.com/opengisch/QField/files/1413682/qgis-project.zip)

**QField version:** [0.10.1]

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: [No]
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* Problem happens with all files and projects, not only some files or projects: [Not sure but I think it depends on the file, with geopacke it is certainly an issue]

saemiw:
Additional info. According to my coworker this problem disappears when using qfieldsync. So maybe the issue is really specific to the geopackage format.

signedav:
I tried to reproduce. I didn't use qfieldsync but copied the project from your zip to my tablet directly instead. There is still the symbology. I've got version 0.10.4 - probably it's solved already there.
Did you tested it with 0.10.4 as well or only 0.10.1 @saemiw ? 
@m-kuhn where there changes effecting a solution on that?

m-kuhn:
Nothing I am aware of.
@saemiw, can you attach (or send us) a sample project that demonstrates the issue?

saemiw:
I'll check tomorrow (under 0.9.10 I can still reproduce the error with).

saemiw:
[...]

PR #213, opened 2018-01-24
Care about legend nodes

signedav:
Functionality to set to visible / invisible and get the visible status for legend nodes - means in categorized etc. layers.

This avoids the crash and is a fix for #199 and mentioning in #181 and let's us display the subcategory as well. And set them visible and invisible like on QGIS.

image

### DIFF
--- a/src/layertreemodel.cpp
+++ b/src/layertreemodel.cpp
@@ -90,7 +90,8 @@ QVariant LayerTreeModel::data( const QModelIndex& index, int role ) const
 
     case Visible:
     {
-      if ( QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) ) )
+      QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) );
+      if ( sym )
       {
         return sym->data( Qt::CheckStateRole ).toBool();
       }
@@ -109,12 +110,10 @@ bool LayerTreeModel::setData(const QModelIndex& index, const QVariant& value, in
 {
   if ( role == Visible )
   {
-    if ( QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) ) )
+    QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) );
+    if ( sym )
     {
-      QVariant checked = Qt::Unchecked;
-      if( value.toBool() )
-        checked=Qt::Checked;
-
+      QVariant checked = value.toBool() ? Qt::Checked : Qt::Unchecked;
       sym->setData( checked, Qt::CheckStateRole );
     }
     else
--- a/src/layertreemodel.cpp
+++ b/src/layertreemodel.cpp
@@ -90,10 +90,16 @@ QVariant LayerTreeModel::data( const QModelIndex& index, int role ) const
 
     case Visible:
     {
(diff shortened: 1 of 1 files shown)


src/layertreemodel.cpp
```diff
@@ -90,10 +90,16 @@ QVariant LayerTreeModel::data( const QModelIndex& index, int role ) const
 
     case Visible:
     {
-      QgsLayerTreeNode* node = mLayerTreeModel->index2node( mapToSource( index ) );
-      return node->isVisible();
+      if ( QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) ) )
```
3nids: I think this will always return true, no?
Don't you have to do:

```
QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) );
if ( sym )
```

src/layertreemodel.cpp
```diff
@@ -103,8 +109,19 @@ bool LayerTreeModel::setData(const QModelIndex& index, const QVariant& value, in
 {
   if ( role == Visible )
   {
-    QgsLayerTreeNode* node = mLayerTreeModel->index2node( mapToSource( index ) );
-    node->setItemVisibilityCheckedRecursive( value.toBool() );
+    if ( QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) ) )
+    {
+      QVariant checked = Qt::Unchecked;
+      if( value.toBool() )
+        checked=Qt::Checked;
```
3nids: the code would a little nicer IMHO with

```
Qt::CheckState checked = (Qt::CheckState) value.toBool();
sym->setData( checked, Qt::CheckStateRole );
```

src/layertreemodel.cpp
```diff
@@ -103,8 +109,19 @@ bool LayerTreeModel::setData(const QModelIndex& index, const QVariant& value, in
 {
   if ( role == Visible )
   {
-    QgsLayerTreeNode* node = mLayerTreeModel->index2node( mapToSource( index ) );
-    node->setItemVisibilityCheckedRecursive( value.toBool() );
+    if ( QgsLayerTreeModelLegendNode *sym = mLayerTreeModel->index2legendNode( mapToSource( index ) ) )
```
3nids: same remark as above

src/layertreemodel.cpp
signedav: No, this makes no difference. It returns false if sym is null in both cases.

But wouldn't mind to change it, if it's ugly or mistakable.

src/layertreemodel.cpp
signedav: But the `Qt::CheckState` has to be `Qt::Checked` what is like 2..

src/layertreemodel.cpp
3nids: interesting!
I found it confusing though.

src/layertreemodel.cpp
3nids: and it convert it to PartiallyChecked if true?

signedav: yes

3nids: too bad
```
Qt::CheckState checked = value.toBool() ? Qt::CheckState : Qt::UnChecked ;
```
but it makes no difference, sorry for the noise

signedav: Thanks. Agree on your inputs anyway. Will make the code cuter.

PR #214, opened 2018-01-24
use transform context in coordinate transform

3nids:


signedav:
Makes sense. But cannot test. The feature zoomToSelected seems not to work. Neither before than after. Let's see on the next release...

3nids:
ok, thanks!

PR #216, opened 2018-01-24
Removed quit option in setting

signedav:
Removed this Menu Item part, because usually a mobile app has not the functionality to quit.
When there is a "quit" the user expects something different than when pressing the back key.


signedav:
It's only about to remove the "Hide" - an option is planned with a message before closing like "Press back again to close" that the user doesn't close it accidentally. I opened an issue for that #217

PR #240, opened 2018-03-14
fix resources path

3nids:


3nids:
see also https://github.com/qgis/QGIS/pull/6604 for a working env

3nids:
superseded by #244

PR #244, opened 2018-03-15
correct use of QgsApplication::setPrefixPath

3nids:


3nids:
`QGIS_PLUGIN_DIR` variable could be removed but it's use in QgisTestApp. I think it requires fixes too.

PR #249, opened 2018-03-19
QgsQuick usage

PeterPetrik:


m-kuhn:
Thanks a lot for this. To make the tests pass we'll need to create an sdk with quick included, I'll be happy to do that.

Are the classes which are ported in this pull request 1:1 copies or are there changes within them that you'd like me to review?

PeterPetrik:
I am going to spend another half day on this to port mapcanvas(settings), but unfortunately I will not be able to work more on QField now. 

Classes should be identical, where they are not, I have just inherited the class from QgsQuick (see featuremodel.cpp)

m-kuhn:
@PeterPetrik thanks for that, the work on this is much appreciated.
On my main focus right now is the smooth integration of QgisQuick into master.

One of my main challenges there is to know which code is copied from QField 1:1, which is modified and which is completely new, this question was very much targeted on that.

PeterPetrik:
related to https://github.com/opengisch/OSGeo4A/pull/21

PeterPetrik:
Note in mapcanvasmap.cpp Martin Dobias implemented some fixes, best to check with QField base implementation

Not ported (not 1-1), but probably worth if you have time
   - qgsquickfeatureform.qml  + editor/*.qml
   - qgsquickpositionkit.cpp (GPS location)
   - qgsquickidentifykit.cpp (replaces identifytool.cpp)
   - qgsquickscalebarkit.cpp 
   - QgsQuick.Utils.dp -> replaces dp context property

Also in qgsquick we have different implementations for 
   - photo capture (qgsquickprotopanel.qml)

PeterPetrik:
@m-kuhn best to use kdiff3 or something to check, but except mapcanvasmap it should be 1-1 (almost). The difference is that QgsQuick does not use QgsProject::instance() so there is a new code to implement this....

Issue #173, opened 2017-09-25
Double values can be entered in integer fields


maxmentha:
When the user creates a new feature and fills the attribute table, he can enter data that does not have the correct format without being warned about it, and without the app to prevent this from happening. For example, if a field is defined as a double real number, but with a scale of 0, the user can still enter numbers with decimals when digitizing. However, they will be saved with the scale that was defined at the creation of the field. This leads to a loss of information that goes unnoticed at the moment of the digitalization.

### Reproduction Steps:

1. create a shapefile with a real/double field with scale 0
2. enter data with decimals
3. the data gets lost...

### Expected behavior:

warning or impossibility to save invalid data

### Observed behavior:

data gets saved with defined scale, without any warning

### Screenshots and GIFs

**QField version:** 0.10.1
## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: ?
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* Problem happens with all files and projects, not only some files or projects: [Yes]

saemiw:
I can confirm this. It happend during some field work of us where we entered double values into an accidentally as integer define field. All the decimal information was lost. It would of course be nice if QField could somehow show a warning when this is the case.

m-kuhn:
> When the user creates a new feature and fills the attribute table, he can enter data that does not have the correct format without being warned about it, and without the app to prevent this from happening. For example ...

Is it only the double/integer issue or are there other ones as well?

saemiw:
I'll have list of potential bugs that I will check hopefully mid-october. I will then also check if this issue is relevant for other data types. For the moment I am only aware of the above mentioned.

saemiw:
There are some other issues (date for example does not save integer, decimal values). This is probably not so problematic. BUT it would of course be nice to have a warning for all the fields if the entry doesn't correspond to the datatype defined.

PR #253, opened 2018-03-25
Allow entering decimals in all double fields

### LINKED ISSUE #197: Problem entering decimals in spatialite layers
QField doesn't offer decimal point when entering Data in a Field with format 'DOUBLE'.

### Reproduction Steps:

1. Create Offline Project from Postgres
2. Fieldtype in Sqlite is  'DOUBLE'
3. Enter Data in QField
4. Data entry with decimal point not possible (Android doesn't offer decimal point in keypad).

### Expected behavior:

Decimal Point should be displayed / usable

### Observed behavior:

No decimal Point displayed / usable

**QField version:** 0.10.4

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: [Yes]

We had no Problems with this in 0.10.3 (was fixed around 0.10.2).
Might be related to #173 and its related fix?

m-kuhn:
The new approach is based on the field type. The previous approach relied on precision which was reported differently, in particular ogr and spatialite would happily return 0 precision for double values.

Fix #197

### DIFF
--- a/src/platformutilities.cpp
+++ b/src/platformutilities.cpp
@@ -63,3 +63,8 @@ void PlatformUtilities::open( const QString& data, const QString& type )
   Q_UNUSED( type )
 }
 
+QString PlatformUtilities::fieldType( const QgsField &field ) const
+{
+  return QVariant( field.type() ).typeName();
+}
+
--- a/src/platformutilities.h
+++ b/src/platformutilities.h
@@ -20,6 +20,7 @@
 #define PLATFORMUTILITIES_H
 
 #include <QObject>
+#include <qgsfield.h>
 #include "picturesource.h"
 
 class PlatformUtilities : public QObject
@@ -47,5 +48,10 @@ class PlatformUtilities : public QObject
 
     Q_INVOKABLE virtual void open( const QString& data, const QString& type );
 
+    /**
+     * Returns the QVariant typeName of a field.
+     * This is a stable identifier (compared to the provider field name).
+     */
+    Q_INVOKABLE QString fieldType( const QgsField& field ) const;
 };
 #endif // PLATFORMUTILITIES_H
--- a/src/qml/editorwidgets/TextEdit.qml
+++ b/src/qml/editorwidgets/TextEdit.qml
@@ -18,7 +18,19 @@ Item {
 
     text: value || ''
 
-    inputMethodHints: field.isNumeric || widget == 'Range' ? platformUtilities.fieldType( field ) === 'double' ? Qt.ImhFormattedNumbersOnly : Qt.ImhDigitsOnly : Qt.ImhNone
+    validator: {
+      if (field.isNumeric || widget == 'Range') {
+        if (platformUtilities.fieldType( field ) === 'double')
+          doubleValidator;
+        else
+          intValidator;
+      }
+      else {
+        null;
+      }
+    }
+
+    inputMethodHints: field.isNumeric || widget == 'Range' ? Qt.ImhFormattedNumbersOnly : Qt.ImhNone
 
     background: Rectangle {
       y: textField.height - height - textField.bottomPadding / 2
@@ -50,5 +62,13 @@ Item {
   FontMetrics {
     id: fontMetrics
     font: textField.font
+  }
+
+  IntValidator {
+    id: intValidator
+  }
+
+  DoubleValidator {
+    id: doubleValidator
   }
 }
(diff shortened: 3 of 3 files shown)


m-kuhn:
@saemiw do you have a chance to test this?

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-253-cb0d1a855dd6192d778ad8b55994ec27a934d117-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-253-cb0d1a855dd6192d778ad8b55994ec27a934d117-x86.apk)

EulenspiegelTill:
@m-kuhn It works fine when testet in one of our field projects. 
The tested sqlite layer displays precision 0 for double values in properties.

m-kuhn:
@EulenspiegelTill do I understand correctly, the download from in here works better than the last release, i.e. issue fixed?

saemiw:
I'll test this afternoon

EulenspiegelTill:
@m-kuhn In our case, yes. With the latest release i was not able to type decimal values for double fields (latest possible was 0.10.3). and with this test apk it works.

saemiw:
I confirm that with the test apk downloaded from this issue by qfield_fairy. It was possible to enter decimal values in a real field. Also it is possible to enter negative values.

Three problems remain (I think only the last one needs "urgent" attention):
- it is possible to enter non-sense things in QField like for example "10-6" into a real field. At first sight QField seems to save this value. When later checking the same point again (after inspecting the attribute table for another point) and also when using qfieldsync the "10-6" entry is saved as "10". This is not really a big problem, but it would be nicer if QField could prevent wrong entries like this one. 

- the precision defined in QField (I used 2.18.17) is however not implemented. If you define precision 2 you can still enter a number like 10.111. I did notice that already QGIS is not storing the precision information (correctly). If I rember correctly this is a known error. Will there be a fix in QGIS by the way (I did not check with QGIS 3)?

- Negative integer values (for example -5) are still not possible to enter, when the numeric field is defined as integer. Would this be possible to add, i.e. allow to choose the "-" symbol?

saemiw:
@m-kuhn I prefer to be able to enter data (like minus temperaturs. like -5 or -5.9) over avoiding the possibilty to enter stupid things (like "9-6"). So the comma and the minus should be available for real (minus and comma) and integer (ideally only the minus) fields. Do I get you right that it is not possible to only offer the minus (but not the comma) if the field is defined as integer?

m-kuhn:
> I prefer to be able to enter data (like minus temperaturs. like -5 or -5.9) over avoiding the possibilty to enter stupid things (like "9-6"). So the comma and the minus should be available for real (minus and comma) and integer (ideally only the minus) fields.

Yep, we just have to make sure we don't introudce https://github.com/opengisch/QField/issues/173 again.

> Do I get you right that it is not possible to only offer the minus (but not the comma) if the field is defined as integer?

Correct, that's what I've found.

m-kuhn:
Let's see what our dear fairy is baking this time.

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-253-734f030642d956b2f20860d97c2efcf2a41ba791-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-253-734f030642d956b2f20860d97c2efcf2a41ba791-x86.apk)

saemiw:
I love the latest cake of the fairy very much. No, jokes aside, this is an excellent solution. On my screen I can type in it wrongly but then you see immidately that qfield changes the value to integer or real-conform format (depending on the field definition). I also checked the transfer with QFieldSync. Works smoothly. Can you implement this in an official release soon? Many thanks, mille grazie, merci beaucoup, danke vöu mou!

m-kuhn:
Perfecto. Está integrado y lo vamos a tener en el proximo release!

saemiw:
I just noted that this fix by this test apk ist not yet implemented in 0.10.11. Could you incorporate that soon into the next release? Thanks!

m-kuhn:
This is in 0.10.12 which is the current release. Does your device not upgrade to this release?

saemiw:
oh shit, sorry my mistake. I always donwload the apk from github and I overlooked that there is a newe version. Sorry for the hassle...all good!

Issue #254, opened 2018-03-25
Settings not saved after exit


mrtwister3142:
With 0.10.9-0.10.11 my settings (e.g. full-screen attribute table) are not saved after quitting QFIELD,

### Reproduction Steps:

1. Open a project
2. Settings -> check show attibute form in full screen
3. close settings, quit QFIELD.

### Expected behavior:

Settings are saved.

### Observed behavior:

Settings are not saved, and last used project is not remembered.

**QField version:** 0.10.9-0.10.11 on Samsung Galaxy A3 2016 and Android 7.0

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: Yes
* Problem can be reliably reproduced, doesn't happen randomly: Yes
* Problem happens with all files and projects, not only some files or projects: Yes

VxTedxV:
Same kind of issue with 0.10.11 for me (but not 0.10.9 as I remember).
QField doesn't open the last project. It shows the starting page with the icon to open a project.
Updated from PlayStore.
Samsung Galaxy XCover4 Android 7.0

mrtwister3142:
also on Samsung Galaxy Tab A with Android 7.0

giedrioks26:
also on blackview BV6000s with Android 7.0

giedrioks26:
same problem with Sony xperia tablet z1, 2, 3, Lenovo tablets, Samsung galaxy tab.
Android 5, 6, 7

bartabbas:
Also on XCover3 under android6 Qfield 0.10.11
Updated from Playstore.

fahrwegdienste:
also on Glalaxy Tab A, Android 7, QField 0.10.11

m-kuhn:
Just a random thought, that this might be related https://github.com/opengisch/QField/issues/259#issuecomment-378885619

> Also starting to notice more and more QField- folders when choosing 'open project'.

PR #263, opened 2018-04-09
Set QSettings format to NativeFormat

signedav:
Then Qt handles the whole background with path, prefix and conf-file.

Issue was, that the settings haven't been stored and read. This leaded to the issue, that the project cannot be opened automatically again on reactivate QField from standby mode. With the NativeFormat instead of IniFormat it says in the [QSettings](http://doc.qt.io/qt-5/qsettings.html) specification.

> On Unix, NativeFormat and IniFormat mean the same thing, except that the file extension is different (.conf for NativeFormat, .ini for IniFormat).

And with NativeFormat there is no problem with the prefix path and so the issue #254 should be fixed.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -46,8 +46,10 @@ int main( int argc, char ** argv )
     QGuiApplication::setAttribute( Qt::AA_EnableHighDpiScaling );
   delete dummyApp;
 #endif
+
   QgsApplication app( argc, argv, true );
   QSettings settings;
+
   app.setThemeName( settings.value( "/Themes", "default" ).toString() );
 
   // load providers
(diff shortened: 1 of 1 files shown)


m-kuhn:
Good job, @signedav 
Let's wait for the apk upload to test and then release as quick as possible

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-263-3ba6ddf6b1117f7e849613514beac1f96be54f4e-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-263-3ba6ddf6b1117f7e849613514beac1f96be54f4e-x86.apk)

PR #316, opened 2018-08-09
[digitizing] remove pen icon

3nids:
I'd like to propose these changes for digitizing to make it more coherent with desktop but also within the app: merge it with the "+" icon.

The button has the same effect as the add vertex button and I find current mode a bit confusing.

[screencast](https://www.youtube.com/watch?v=CeVu2Y64cww "screencast")

### DIFF
--- a/src/qml/DigitizingToolbar.qml
+++ b/src/qml/DigitizingToolbar.qml
@@ -43,19 +43,12 @@ Row {
   Button {
     id: confirmButton
     iconSource: {
-      if ( Number( rubberbandModel.geometryType ) === 0 )
-      {
-        Style.getThemeIcon( "ic_create_white_24dp" )
-      }
-      else
-      {
-        Style.getThemeIcon( "ic_save_white_24dp" )
-      }
+      Style.getThemeIcon( "ic_save_white_24dp" )
     }
     visible: {
       if ( Number( rubberbandModel.geometryType ) === 0 )
       {
-        true
+        false
       }
       else if  ( Number( rubberbandModel.geometryType ) === 1 )
       {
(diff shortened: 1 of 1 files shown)


qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-316-aed1e968481e76ed52efd9f5efca9423e6a46f4d-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-316-aed1e968481e76ed52efd9f5efca9423e6a46f4d-x86.apk)

signedav:
+1 

I like the idea. The "pen" symbol is actually more for an "edit" than an "add" button.

3nids:
yes, but it was adding a vertex, so I found it confusing.
I'll wait for @m-kuhn opinion too, thanks for the input

signedav:
An additional idea:

Do we need the X to cancel the digitization? 
Or would minus, minus, minus, minus make it as well?

image

If it's a big lack of comfort I would keep the X - but I think we could live without X as well. Or make a long press on the minus to delete all corners...

3nids:
not sure about this one, the long press is nice, but it's kind of a hidden feature.
We could display a message "long press to cancel" if there is some vertices removed at a high speed. That could do the job.

signedav:
Don't you think it's intuitive like backspacing text?
But message would be okay for me as well...

3nids:
I'm maybe not a heavy user enough to know....

signedav:
Me neither :-)

m-kuhn:
Looks great to me!

Long press on + -> was proposed for going into "autocapture" mode (new vertex every x seconds/meters), so I'd rather not change to this. But maybe changing the x to a trashbin?

signedav:
Or remove it completely. If you digitize a feature and then you want to cancel it, it's a small part of the effort to press minus, minus, minus, minus... But it's not that relevant for me.

3nids:
@m-kuhn it was long press on - and not +, so it could work, no?

m-kuhn:
> @m-kuhn it was long press on - and not +, so it could work, no?

good point, yes :)

Issue #268, opened 2018-04-19
Save/Cancel options


signedav:
With the Save and Canceling of **Settings** and after **editing features** there appeared some problems I don't know since when. Anyway I think we should think about it in general. Since we will have a solid focus-handling we could use sometimes the back-key instead of cancel.

Save and Cancel is used in the following situations:
- Settings
- Edit Feature
- Add Feature 

### Settings
![01_qfield_settings](https://user-images.githubusercontent.com/28384354/38996469-8f5e300a-43eb-11e8-8e4e-d1869ae0fcde.gif)
**Problem:**
It doesn't matter if I click Save or Cancel or back-key. It's saved anyway.
**Suggestion:**
Remove the option Save/Cancel and let it be saved always on closing (with back-key). IMHO the UX is, that it's saved then. Anyway I don't really know apps where we use Save/Close on settings...

### Edit Feature
![02_qfield_featureedit](https://user-images.githubusercontent.com/28384354/38996635-e6ece14a-43eb-11e8-8adf-53c2598085a8.gif)
**Problem:**
It doesn't matter if I click Save or Cancel or back-key. It's saved anyway.
**Suggestion:**
I see three sollutions:
1. Fix it. Then it will be saved on Save and not saved on Cancel or back-key.
2. Remove Cancel. Then we have Save for save and back-key for cancel (because I suppose that a user anyway never uses cancel but back-key instead)
3. Remove Cancel and Save. Then it will be always saved when closing (pressing back-key)

### Add Feature
![03_qfield_digitize](https://user-images.githubusercontent.com/28384354/38996789-39ebc4ec-43ec-11e8-9ea7-0c612d132c9a.gif)
**Problem:**
No problem. This works like expected.
**Suggestion:**
I would make it like the Edit Feature. Means if we remove any Save/Cancel option, I would remove it here as well.

What do you think @m-kuhn @3nids @mbernasocchi ?

I suppose it's a small thing to do that. And I propose to do it before 1.0.

mbernasocchi:
specifically: https://stackoverflow.com/questions/4988224/how-to-handle-save-and-cancel-buttons-and-the-back-key/4989368#4989368

> I highly recommend you following the model used by the standard Android and apps, where editing always happens in place. That is, there is no "save" button -- the changes you are making are effectively being saved as they are being done.

signedav:
Thanks @mbernasocchi Yes, that's my opinion as well. For **settings** it's everyone's I guess.
[...]

PR #320, opened 2018-08-14
Save/Cancel option

signedav:
Regarding the discussion here #268

Sorry for the GIF frenzy ;-)

### Add-Session
Save-Button is removed. It saves on pressing "back"-key. It cancels on X. 
Message "Changes saved" / "Changes discarded" appears.
add-save-cancel

Before the Save-Button was greyed out if constraints where not valid yet. I did it with orange color and message "Constraints not valid"

add-save-constraint

***Open question: Should it be saved or canceled on tipping on canvas? Currently it saves.***

### Edit-Session
Save-Button is removed. It saves on pressing "back"-key. It cancels on X.
Message "Changes saved" / "Changes discarded" appears.
edit-save-cancel

I did the constraint handling here with orange color and message "Constraints not valid" as well.
edit-save-constraints

***Open question: Should it be saved or canceled on tipping on canvas? Currently it cancels.***

### Settings
Settings are always saved. No canceling here. Additionally I changed the checkboxes to switch-buttons.
settings

### Other Fixes in this implementation:
- Reset of values on cancel in form on edit-session (before the have been only reseted in the model)
- Care about constraints in edit-session (before you were able to save even with invalid conastraints)
- Hide of sidebar (FeatureList) when no Feature is selected (before there was an empty sidebar)


signedav:
@3nids @m-kuhn Probably it's good, if you test it on the device to feel the handling.

Please give a feedback on the two open questions. It should be the same on both (edditing and adding) - but I want your opinion before I change the one or the other.

Thanks.

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-320-0721f67aa271178617e54bc0af3d324e48a87e1b-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-320-0721f67aa271178617e54bc0af3d324e48a87e1b-x86.apk)

m-kuhn:
I like the background colors for the constraints :+1: 

I'm not yet convinced by the cross icon for saving changes. I find myself looking for the "save" button, which shouldn't be if it's natural and intuitive. I have just had a look at the android stock contacts app, there's a button with the label (text) "save", the calendar I use (acal) has a cross (dismiss) and a checkmark (save) to accept new agenda entries. They also allow aborting an edit operation.

Did you look at other apps as examples?

VxTedxV:
according to me :
- cross should not save
- back button should not save (as it's used as cancel button in many apps)
- tap on the map to exit should save entries 
- a "save" icon or text should valid entries

signedav:
Just for information: The X cross is for cancel, not for saving.

m-kuhn:
Not sure if we are talking about the same thing. I was talking about the feature form where I cannot see any other button which would save then?

signedav:
The Back-Key on the device is to save: 
backkey

m-kuhn:
That was not obvious to me at least :/

signedav:
Ok, thanks. I like it like this pretty much, but I wasn't sure if everyone does. That's why I started the discussion in #268

For me it's intuitive on editing a feature that it's saved on back key. On adding a feature I thought it is as well because we add the feature (after digitizing) and then kind of edit it's attributes. I think it makes sense, that the behavior is the same on editing and adding features.

On other apps the back key is used diverse. It depends on the workflow. 
Eg. Some Google Map functionality (setting labels) save on back-key. Acalendar for example has the hook (like m-kuhn said). Evernote has a hook and saves on back...

VxTedxV:
I guess it's just a matter of getting used to it. I'm personally not against any choice in the end ;)
Just be sure not to make a choice and go back... for the user to have time to get the habit !

m-kuhn:
Modern devicec don't have physical buttons any more and from what I can see, the trend is to go to software buttons everywhere. Especially if we consider iOS apps in the future as well, where there does not even a software button exist (and none is emulated like on all the android devices without a hardware button).

Note: I am not against saving with the back key, but I think we need a button on the form as well. I did not check, but I guess even the mentioned google maps functionality does have one?

signedav:
VxTedxV is right, that if we do it, it has to be definitive and we don't change it back on the next release. That's why it's important that this solution is okay for everyone.

@m-kuhn Google has a back arrow icon that does the same like back key. Actually on Evernote it's weird: On adding it's saved with back arrow and on edit it's not. That's not what I expected.

So should I make an update - working the same like now but including a checkmark or floppy icon?

What is the opinion of my questions with the tip on the canvas?

3nids:
yep, makes sense to add the button then

signedav:
- Reinserted save buttons
- Save on tip on the canvas (like with back-key) and avoid saving with invalid constraint

savecancel_points

savecancel_poly

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-320-d68ffb02c50d8daa3bf273297ac1cc47edeb0025-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-320-d68ffb02c50d8daa3bf273297ac1cc47edeb0025-x86.apk)

signedav:
I made some design changes for the icons:

1. Changed + button for digitizing and saving points from green to yellow. Because this button combines digitizing (in polygon the green plus) and saving (in polygon the yellow floppy). So it has to be coherent to both. We decided to make a yellow plus.
selection_028

2. Changed for adding the cancel-key (X) to a trashcan, while it's still the X on editing. This because you vanish there everything, while on editing you just cancel your changes.
selection_029

3. Replaced the floppy on adding and editing with a checkmark, because it looks more modern and most apps uses checkmarks to accept changes.
selection_030
Hereby follows the question, if we should replace the yellow floppy on digitizing as well? What do you think @VxTedxV and @3nids? Are you happy with these changes (and the ones before)?

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-320-5d058e11555de9d8b7edf92bf41902706cd778c7-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-320-5d058e11555de9d8b7edf92bf41902706cd778c7-x86.apk)

3nids:
sorry I meant for editing.
But looking at it again, I think it's fine as is!
good job!

PR #324, opened 2018-08-21
Print to PDF

signedav:
When a project has print layouts, they are available on QField as well.

It's tested on a mobile phone, that's why the small screen:
print2

On a large amount of layouts, it would be still comfortable to handle.

printing

So the basic works. If the print layouts looks the same on device than on the computer needs to be validated still.

Nice things would be: 
- Busy indicator (because old devices are slow) - the implementation in this PR is not yet working.
- Temporary file handling (that files are deleted afterwards)

### DIFF
--- a/src/printlayoutlistmodel.cpp
+++ b/src/printlayoutlistmodel.cpp
@@ -36,10 +36,10 @@ QgsProject* PrintLayoutListModel::project() const
 
 void PrintLayoutListModel::setProject( QgsProject* project )
 {
-  if( mProject != project )
-    mProject = project;
+  if( mProject == project )
+    return;
 
-  reloadModel();
+  mProject = project;
 }
 
 void PrintLayoutListModel::reloadModel()
--- a/src/printlayoutlistmodel.cpp
+++ b/src/printlayoutlistmodel.cpp
@@ -1,0 +1,75 @@
+/***************************************************************************
+  printlayoutlistmodel.cpp
+ ---------------------
+ begin                : 20.8.2018
+ copyright            : (C) 2018 by david
+ email                : david at opengisch dot ch
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+#include "printlayoutlistmodel.h"
+#include <qgsproject.h>
+
+PrintLayoutListModel::PrintLayoutListModel(QObject* parent)
+  : QAbstractListModel( parent )
+{
+}
+
+QHash<int, QByteArray> PrintLayoutListModel::roleNames() const
+{
+  QHash<int, QByteArray> roles = QAbstractListModel::roleNames();
+  roles[TitleRow]  = "Title";
+  roles[IndexRow] = "Index";
+
+  return roles;
+}
+
+QgsProject* PrintLayoutListModel::project() const
+{
+  return mProject;
+}
+
+void PrintLayoutListModel::setProject( QgsProject* project )
+{
+  mProject = project;
+
+  reloadModel();
+}
+
+void PrintLayoutListModel::reloadModel()
+{
+  beginResetModel();
+  mPrintLayouts.clear();
+
+  for ( const auto &layout : mProject->layoutManager()->printLayouts() )
+  {
+    mPrintLayouts.append( PrintLayout( layout->name() ) );
+  }
+  endResetModel();
+
+}
+
+int PrintLayoutListModel::rowCount( const QModelIndex& parent ) const
+{
+  Q_UNUSED( parent )
+  int test = mPrintLayouts.size();
+  return test;
+}
+
+QVariant PrintLayoutListModel::data( const QModelIndex& index, int role ) const
+{
+  if ( index.row() >= mPrintLayouts.size() )
+    return QVariant();
+
+  if ( role == TitleRow )
+    return mPrintLayouts.at( index.row() ).title;
+  else if ( role == IndexRow )
+    return  index.row();
+
+  return QVariant();
+}
--- a/src/printlayoutlistmodel.cpp
+++ b/src/printlayoutlistmodel.cpp
@@ -36,7 +36,8 @@ QgsProject* PrintLayoutListModel::project() const
 
 void PrintLayoutListModel::setProject( QgsProject* project )
 {
-  mProject = project;
+  if( mProject != project )
+    mProject = project;
 
   reloadModel();
 }
@@ -58,8 +59,10 @@ void PrintLayoutListModel::reloadModel()
 
 int PrintLayoutListModel::rowCount( const QModelIndex& parent ) const
 {
-  Q_UNUSED( parent )
-  return mPrintLayouts.size();
+  if ( !parent.isValid() )
+    return mPrintLayouts.size();
+  else
+    return 0;
 }
 
 QVariant PrintLayoutListModel::data( const QModelIndex& index, int role ) const
--- a/src/printlayoutlistmodel.cpp
+++ b/src/printlayoutlistmodel.cpp
@@ -46,7 +46,9 @@ void PrintLayoutListModel::reloadModel()
   beginResetModel();
   mPrintLayouts.clear();
 
-  for ( const auto &layout : mProject->layoutManager()->printLayouts() )
+  const QList< QgsPrintLayout * > layouts( mProject->layoutManager()->printLayouts() );
+
+  for ( const auto &layout : layouts )
   {
     mPrintLayouts.append( PrintLayout( layout->name() ) );
   }
@@ -57,13 +59,12 @@ void PrintLayoutListModel::reloadModel()
 int PrintLayoutListModel::rowCount( const QModelIndex& parent ) const
 {
   Q_UNUSED( parent )
-  int test = mPrintLayouts.size();
-  return test;
+  return mPrintLayouts.size();
 }
 
 QVariant PrintLayoutListModel::data( const QModelIndex& index, int role ) const
 {
-  if ( index.row() >= mPrintLayouts.size() )
+  if ( index.row() >= mPrintLayouts.size() || index.row() < 0 )
     return QVariant();
 
   if ( role == TitleRow )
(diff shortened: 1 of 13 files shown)


signedav:
@m-kuhn should the print()-function be in the printlayoutmodel or is the qgismobileapp the correct place?

src/printlayoutlistmodel.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  reloadModel();
+}
+
+void PrintLayoutListModel::reloadModel()
+{
+  beginResetModel();
+  mPrintLayouts.clear();
+
+  for ( const auto &layout : mProject->layoutManager()->printLayouts() )
```
m-kuhn: Take a const copy of the list to avoid a detach

src/printlayoutlistmodel.cpp
```diff
@@ -0,0 +1,75 @@
...
+  }
+  endResetModel();
+
+}
+
+int PrintLayoutListModel::rowCount( const QModelIndex& parent ) const
+{
+  Q_UNUSED( parent )
+  int test = mPrintLayouts.size();
+  return test;
```
m-kuhn: Why the test? :)

src/printlayoutlistmodel.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+QVariant PrintLayoutListModel::data( const QModelIndex& index, int role ) const
+{
+  if ( index.row() >= mPrintLayouts.size() )
```
m-kuhn: or < 0?

src/printlayoutlistmodel.h
```diff
@@ -0,0 +1,68 @@
...
+#include <qgslayoutpagecollection.h>
+
+class QStringList;
+class QgsProject;
+
+class PrintLayoutListModel : public QAbstractListModel
+{
+    Q_OBJECT
+
+    Q_PROPERTY( QgsProject* project READ project WRITE setProject )
```
m-kuhn: If in doubt, also do a NOTIFY

src/printlayoutlistmodel.h
```diff
@@ -0,0 +1,68 @@
...
+
+    enum Roles
+    {
+      TitleRow,
```
m-kuhn: Can you add some (very short) docstrings

src/qgismobileapp.cpp
```diff
@@ -268,6 +278,36 @@ void QgisMobileapp::loadProjectFile( const QString& path )
   emit loadProjectEnded();
 }
 
+void QgisMobileapp::print( int layoutIndex )
+{
+  const QList<QgsPrintLayout *> projectLayouts( mProject->layoutManager()->printLayouts() );
+
+  const auto &layoutToPrint = projectLayouts.at( layoutIndex );
```
m-kuhn: This will crash if the index is out of bounds. Not sure if you caught that somewhere else

src/qgismobileapp.cpp
```diff
@@ -268,6 +278,36 @@ void QgisMobileapp::loadProjectFile( const QString& path )
...
+
+  emit printingStarted( layoutToPrint->name() );
+
+  if ( layoutToPrint->pageCollection()->pageCount() == 0 )
+    return;
+
+  QPrinter printer;
+  printer.setOutputFormat(QPrinter::PdfFormat);
+  printer.setPaperSize(QPrinter::A4);
```
m-kuhn: Is this required? I think that's defined by the layout itself

src/qgismobileapp.cpp
```diff
@@ -268,6 +278,36 @@ void QgisMobileapp::loadProjectFile( const QString& path )
...
+  printer.setOutputFileName( mProject->homePath() + "/" + layoutToPrint->name() + ".pdf" );
```
m-kuhn: QStringLiteral for the ".pdf", '/' for the "/"

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-324-cafedc20caac2274ac7d7fad07818d10f31a9249-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-324-cafedc20caac2274ac7d7fad07818d10f31a9249-x86.apk)

m-kuhn:
Good job :tada: Love it!!

m-kuhn:
> Nice things would be:
> 
>    Busy indicator (because old devices are slow) - the implementation in this PR is not yet working.

Right, that would be nice. I don't know the API for this.

>    Temporary file handling (that files are deleted afterwards)

I think you convinced me that it is even better without this

signedav:
Regarding the Busy Indicator I have some trouble.

On the current implementation, where the `QgisMobileapp::print` emits signals (`printingStarted` on start and printingEnded` on end), there is nothing happening on the BusyIndicator `busyMessageIndicator` and it's rectangle `busyMessage`. Why? Because there seems to be no visual updates while something is happening in `QgisMobileapp::print`, like the Thread would be blocked (but the signals are received).
I belief that this is on loading Projects as well, so the old code would work neither. 
Do you know a sollution for this @m-kuhn or @3nids ?

I read about `WorkerScripts` but there we will need to call the `QgisMobileapp` from a `.js` File, and I'm not sure about all the consequences there...

PS. The BusyIndicator is mainly important on slow devices. While my Galaxy S6 Mobile Phone is quite fast on our Active Tap it's taking around 12 seconds to print a layout...

src/qml/qgismobileapp.qml
```diff
@@ -809,8 +808,8 @@ ApplicationWindow {
   Controls.BusyIndicator {
     id: busyIndicator
     anchors { right: alertIcon.left; top: parent.top }
-    width: 36 * dp
-    height: 36 * dp
+    width: 96 * dp
```
m-kuhn: Is this related to this change?
I think it's unfortunately not that simple, it looks good on low-res devices and bad on high-res devices, so I think we need to do something else than just increase the factor. Can we just keep it out for now?

signedav: Yes. Actually it has been more for size testing. We can remove it.

src/printlayoutlistmodel.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+QgsProject* PrintLayoutListModel::project() const
+{
+  return mProject;
+}
+
+void PrintLayoutListModel::setProject( QgsProject* project )
+{
+  mProject = project;
```
m-kuhn: Better add a guard here

```if ( mProject == project ) 
return;
?```

File: src/printlayoutlistmodel.cpp [hunk 11]
```diff
@@ -0,0 +1,76 @@
...
+    mPrintLayouts.append( PrintLayout( layout->name() ) );
+  }
+  endResetModel();
+
+}
+
+int PrintLayoutListModel::rowCount( const QModelIndex& parent ) const
+{
+  Q_UNUSED( parent )
+  return mPrintLayouts.size();
```
m-kuhn:
```
if ( !parent.isValid() )
   return layouts.size()
else
  return 0;
```

src/printlayoutlistmodel.h
```diff
@@ -0,0 +1,82 @@
...
+    Q_PROPERTY( QgsProject* project READ project WRITE setProject NOTIFY projectChanged )
+
+     /*!
+      * Roles to get the data of the model.
+      * To get the printLayout attributes TitleRow is used.
+      * To get position in the list of PrintLayouts IndexRow is used.
+     */
+    enum Roles
+    {
+      TitleRow,
```
m-kuhn: You can add docstrings to individual items

```
  TitleRow, //!< The title of a printLayout
```

src/printlayoutlistmodel.h
```diff
@@ -0,0 +1,82 @@
...
+      IndexRow
+    };
+
+    /*!
+     * Struct of printLayout attributes (at the moment only for the title (name) used)
+    */
+    //! at the moment only contains the title string - probably in future more attributes needed
```
m-kuhn: I think you can remove this line, that should be obvious ;)

m-kuhn:
> BusyIndicatorbusyMessageIndicatorand it's rectanglebusyMessage. Why? Because there seems to be no visual updates while something is happening inQgisMobileapp::print`, like the Thread would be blocked (but the signals are received).

I guess the printing code is not safe to be sent to a background thread. I'd just accept the deadlock for now. Without digging a bit deeper I would be worried to send it to a background thread.

signedav:
Hm... Okay, I understand. 
Are you aware that the BusyIndicator on opening files does not work neither or is it working and I missed something?

m-kuhn:
Yes, makes sense, that it's the same issue there.
Can you have a look at the last comments, I'd like to merge this soon!

src/printlayoutlistmodel.cpp
signedav: I've seen it in the QAbstractItemModel, that it should return 0. But I don't understand why...

src/printlayoutlistmodel.cpp
signedav: Because it's used for refreshing model as well, I would prefer:
```
  if( mProject != project )
    mProject = project;

  reloadModel();
}
```
What do you think?

m-kuhn: Why not call `reloadModel()` explicitly in such cases?

src/printlayoutlistmodel.cpp
m-kuhn: models are capable of providing tree-structures. So you can call this for an item at row 1 and it will give you the number of nodes http://doc.qt.io/qt-5/qtwidgets-itemviews-simpletreemodel-example.html#design-and-concepts

signedav: Okay, I understand. Because it request of a table the amount of nonexistent children of an entry. Thanks for clarification.

src/printlayoutlistmodel.cpp
signedav: I thought because it needs to reload after loading project anyway. But I can do that and then there are two calls onLoadProjectEnded. No problem.

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-324-b4a268c42d10930fba7c9a77864332a882d29e8e-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-324-b4a268c42d10930fba7c9a77864332a882d29e8e-x86.apk)

PR #346, opened 2018-08-29
Buttons in the feature form

signedav:
After reporting of weird display of the buttons I replaced the `ToolButton` with the Buttons we use usually.
Could confirm the misplaced issue on my tablet and the width with my mobile.

Looks now like this:
screenshot_20180829-073331


signedav:
@mbernasocchi okay now?
@m-kuhn why did we used `ToolButton` originally? Because there has been no own implementations of Butttons when it was implemented?

m-kuhn:
@giedrioks26 does a thumbs up mean "tested and works"?
@signedav it was one of the first buttons done ever, so that's just old code.

PR #359, opened 2018-09-07
Update French translation

florida63:
a little contribution


m-kuhn:
Thank you.
These will in the future be overwritten by [translations on transifex](https://www.transifex.com/opengisch/qfield-for-qgis/), can you add them there so they will not be lost?

florida63:
I look this.

PR #376, opened 2018-10-03
Fix of remembering wrongly date

### LINKED ISSUE #319: Edited date is visible on all other objects
Dear all

Once a date field has been edited, it will be displayed on all other objects, even if the date for these objects is not saved.
It sounds a bit confusing, but I'll try to explain it below.

### Reproduction Steps:

I have a tree layer with a felling-date attribute. By editing this attribute, for example tree number 100 has now felling-date 2018-08-14, the same date is shown for all other trees as well.

Only by closeing and reopening the application, the felling-date for the other trees is reset to their initial state again.

### Expected behavior:

The modification of the date of one attribute mustn't be displayed for all the other objects as well.

### Observed behavior:

The felling-date wasn't set for all trees, just for tree number 100. Therefore it must be a display problem I guess.

**QField version:** [0.10.13]

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: [No]
I tried several projects and different data sources (shape, gepackage, sqlite).

Regards and thanks
Christoph

### LINKED ISSUE #360: Bug with dates and default values
When defining a field as date, QField automatically sets a default value after having entered a first value in the date field. Reproduced on last vers 0.11.6 and also on old versions 0.10. To reproduce : 

1°/ Open attached project on QField 
2°/ Create a new feature and enter a "date_mesure" -> checkbox beside field is automatically checked, but we do not want it 
3°/ Save the feature
4°/ Create a second new feature -> "date_mesure" already contains a date, but we do not want it (field is not defined in QGIS as "not null") and it is not possible to set the "date_mesure" to null

This issue is really annoying when dealing with dates in a production environnement...

[test.zip](https://github.com/opengisch/QField/files/2370238/test.zip)

### LINKED ISSUE #369: Display of date/time widget doesn't correspond to what it saves
When you have a date field QField will present a calendar popup. When you choose a date upon the first entry, QField will show you the same date in the second entry. The apprearance in the form suggests that you are saving that date. In fact QField doesn't save a date at all. 

### Reproduction Steps:

1. On Qfield enter a date by chosing a value in the calendar popup
2. Save this first record
3. Start a new (second) entry. In the date field the previously chosen date will appear. Save that records
4. check the date in the second record - it will be empty despite the appearence of the form when you entered the data.

### Expected behavior:

I'd expect that the date widget upon entry of the second records is empty so I have to deliberately choose a date. The chosen date has then to be saved to the data.

### Observed behavior:

This video shows the behaviour: https://youtu.be/xs1qwSnGLyY

## Additional information:

I used QGIS 3.2 and QField 0.11.6

signedav:
... and still not saving it. 

Additionally there has been a cross when entering a date that has been unluckily placed exactly on the RememberValue box. So it's placed correctly now, because it's **a clear-cross to erase the date-value (if allowNull)**...
Looks like this:
datum

The other thing has been, on "Add Feature" the date of the last edited feature was always displayed - though not saved. So this is fixed now as well.

fix #369
fix #360
fix #319

Would be nice, if you could test that @gnerred @saemiw :slightly_smiling_face: Thanks.

### DIFF
--- a/src/qml/editorwidgets/DateTime.qml
+++ b/src/qml/editorwidgets/DateTime.qml
@@ -52,7 +52,8 @@ Item {
 
         Image {
           source: Style.getThemeIcon("ic_clear_black_18dp")
-          anchors.left: parent.right
+          anchors.right: parent.right
+          anchors.verticalCenter: parent.verticalCenter
           visible: main.currentValue !== undefined && config['allow_null']
 
           MouseArea {
--- a/src/qml/editorwidgets/DateTime.qml
+++ b/src/qml/editorwidgets/DateTime.qml
@@ -43,8 +43,8 @@ Item {
         anchors.fill: parent
         verticalAlignment: Text.AlignVCenter
 
-        text: value === undefined ||  value === '' ?  qsTr('(no date)') : new Date(value).toLocaleString(Qt.locale(), config['display_format'] )
-        color: value === undefined || value === '' ? 'gray' : 'black'
+        text: value === undefined ?  qsTr('(no date)') : new Date(value).toLocaleString(Qt.locale(), config['display_format'] )
+        color: value === undefined ? 'gray' : 'black'
 
         MouseArea {
           anchors.fill: parent
@@ -57,7 +57,7 @@ Item {
           source: Style.getThemeIcon("ic_clear_black_18dp")
           anchors.right: parent.right
           anchors.verticalCenter: parent.verticalCenter
-          visible: ( value !== undefined || value === '' ) && config['allow_null']
+          visible: ( value !== undefined ) && config['allow_null']
 
           MouseArea {
             anchors.fill: parent
--- a/src/qml/editorwidgets/DateTime.qml
+++ b/src/qml/editorwidgets/DateTime.qml
@@ -43,6 +43,9 @@ Item {
         anchors.fill: parent
         verticalAlignment: Text.AlignVCenter
 
+        text: value === undefined ||  value === '' ?  qsTr('(no date)') : new Date(value).toLocaleString(Qt.locale(), config['display_format'] )
+        color: value === undefined || value === '' ? 'gray' : 'black'
+
         MouseArea {
           anchors.fill: parent
           onClicked: {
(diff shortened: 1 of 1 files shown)


qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-376-2ba4cb70e2a2478524614669be1140e542b1bdf8-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-376-2ba4cb70e2a2478524614669be1140e542b1bdf8-x86.apk)

src/qml/editorwidgets/DateTime.qml
```diff
@@ -52,8 +55,9 @@ Item {
 
         Image {
           source: Style.getThemeIcon("ic_clear_black_18dp")
-          anchors.left: parent.right
-          visible: main.currentValue !== undefined && config['allow_null']
+          anchors.right: parent.right
+          anchors.verticalCenter: parent.verticalCenter
```
m-kuhn: are both placements required? To me it looks like it should be either one or the other, since they can be contradicting?

src/qml/editorwidgets/DateTime.qml
```diff
@@ -52,8 +55,9 @@ Item {
...
+          visible: ( value !== undefined || value === '' ) && config['allow_null']
```
m-kuhn: `( value !== undefined || value === '' )` isn't that equal to `value !== undefined`?

signedav: Seemed like not. But I will recheck.

src/qml/editorwidgets/DateTime.qml
signedav: Why contradicting? One is vertical alignment and the other horizontal. If I make no `verticalCenter` it's on the right top corner...

m-kuhn: oops sorry

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-376-7b37815c5a6e3e6f99dd510e0750b0dbfce2cc93-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-376-7b37815c5a6e3e6f99dd510e0750b0dbfce2cc93-x86.apk)

gnerred:
I installed test apk armv7, but I get an error on my project loading : "Picture expression eval error" written two times upper left on a white screen. If I touch, the map appears and the date problem is solved. Thanks.

signedav:
The project you attached on the issue @gnerred ?

gnerred:
Yes I think, but I do not remember exactly which version I gave you

PR #381, opened 2018-10-10
Remember values over layers

### LINKED ISSUE #177: Remember "Remember checkboxes" when switching between layers
The "remember checkboxes" make data entry faster/easier by remembering values that are used over and over. Our this years field work has shown that people often have to switch between different tables (so: digitize a tree in the table trees and then digitize a bird in the table birds). When you check a "remember checkbox" this checked state is lost as soon as you start digitizing on another layer. It would be nice if the "remember checkbox" state would be remembered until QField is closed and not until the next layer is edited.

**QField version:** [0.10.1]

## Additional information:

* The data are saved in geopackages
* The project is set up on QGIS 2.18.13

signedav:
It's done with a `QMap` storing the rememberAttributes and the last feature.

rememberlayers

Fix #177

### DIFF
--- a/src/core/featuremodel.cpp
+++ b/src/core/featuremodel.cpp
@@ -43,18 +43,29 @@ void FeatureModel::setFeature( const QgsFeature& feature )
 
 void FeatureModel::setCurrentLayer( QgsVectorLayer* layer )
 {
+  //remember the feature of old layer
+  if( mRememberings.contains( mLayer ) )
+  {
+    mRememberings[mLayer].rememberedFeature = mFeature;
+  }
+
   if ( layer == mLayer )
     return;
 
   mLayer = layer;
   if ( mLayer )
   {
-    mFeature = QgsFeature( mLayer->fields() );
-
-    mRememberedAttributes.resize( layer->fields().size() );
-    mRememberedAttributes.fill( false );
+    //load remember values or create new entry in mRememberings
+    if( mRememberings.contains( mLayer ) ){
+      mFeature = mRememberings[mLayer].rememberedFeature;
+    }else{
+      mFeature = QgsFeature( mLayer->fields() );
+      mRememberings[mLayer].rememberedAttributes.resize( layer->fields().size() );
+      mRememberings[mLayer].rememberedAttributes.fill( false );
+    }
   }
 
+  //dave: und schliesslich auch grad befüllen, denn es wird nur reset gemacht wenn nicht remember und nicht wiederhergestellt bei remember...
   emit currentLayerChanged();
 }
 
@@ -122,7 +133,7 @@ QVariant FeatureModel::data( const QModelIndex& index, int role ) const
       break;
 
     case RememberAttribute:
-      return mRememberedAttributes.at( index.row() );
+      return mRememberings[mLayer].rememberedAttributes.at( index.row() );
       break;
   }
 
@@ -155,7 +166,7 @@ bool FeatureModel::setData( const QModelIndex& index, const QVariant& value, int
 
     case RememberAttribute:
     {
-      mRememberedAttributes[ index.row() ] = value.toBool();
+      mRememberings[mLayer].rememberedAttributes[ index.row() ] = value.toBool();
       emit dataChanged( index, index, QVector<int>() << role );
       break;
     }
@@ -222,7 +233,7 @@ void FeatureModel::resetAttributes()
   beginResetModel();
   for ( int i = 0; i < fields.count(); ++i )
   {
-    if ( !mRememberedAttributes.at( i ) )
+    if ( !mRememberings[mLayer].rememberedAttributes.at( i ) )
     {
       if ( fields.at( i ).defaultValueDefinition().isValid() )
       {
@@ -321,5 +332,5 @@ void FeatureModel::applyVertexModelToGeometry()
 
 QVector<bool> FeatureModel::rememberedAttributes() const
 {
-  return mRememberedAttributes;
+  return mRememberings[mLayer].rememberedAttributes;
 }
--- a/src/core/featuremodel.h
+++ b/src/core/featuremodel.h
@@ -38,6 +38,13 @@ class FeatureModel : public QAbstractListModel
     Q_PROPERTY( QString positionSourceName READ positionSourceName WRITE setPositionSourceName NOTIFY positionSourceChanged )
     Q_ENUMS( FeatureRoles )
 
+
+  struct RememberValues
+  {
+      QgsFeature rememberedFeature;
+      QVector<bool> rememberedAttributes;
+  };
+
   public:
     enum FeatureRoles
     {
@@ -126,9 +133,9 @@ class FeatureModel : public QAbstractListModel
     QgsFeature mFeature;
     VertexModel *mVertexModel = nullptr;
     Geometry* mGeometry;
-    QVector<bool> mRememberedAttributes;
     std::unique_ptr<QGeoPositionInfoSource> mPositionSource;
     QString mTempName;
+    QMap<QgsVectorLayer*, RememberValues> mRememberings;
 };
 
 #endif // FEATUREMODEL_H
--- a/src/core/attributeformmodelbase.cpp
+++ b/src/core/attributeformmodelbase.cpp
@@ -284,6 +284,10 @@ void AttributeFormModelBase::flatten( QgsAttributeEditorContainer* container, QS
       case QgsAttributeEditorElement::AeTypeInvalid:
         // todo
         break;
+
+      case QgsAttributeEditorElement::AeTypeQmlElement:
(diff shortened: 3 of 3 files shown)


src/core/featuremodel.cpp
```diff
@@ -46,15 +46,24 @@ void FeatureModel::setCurrentLayer( QgsVectorLayer* layer )
...
-
-    mRememberedAttributes.resize( layer->fields().size() );
-    mRememberedAttributes.fill( false );
+    //load remember values or create new entry
+    if( mRememberings.contains( mLayer ) ){
+      mFeature = mRememberings[mLayer].rememberedFeature;
+    }else{
+      mFeature = QgsFeature( mLayer->fields() );
+      mRememberings[mLayer].rememberedAttributes.resize( layer->fields().size() );
+      mRememberings[mLayer].rememberedAttributes.fill( false );
```
3nids: two lines can be merged, `fill` also have a resize param.

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-a1f8743a587bf07032010a04232c58e18cb21fbd-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-a1f8743a587bf07032010a04232c58e18cb21fbd-x86.apk)

src/core/featuremodel.cpp
```diff
@@ -46,15 +46,23 @@ void FeatureModel::setCurrentLayer( QgsVectorLayer* layer )
...
   if ( mLayer )
   {
-    mFeature = QgsFeature( mLayer->fields() );
-
-    mRememberedAttributes.resize( layer->fields().size() );
-    mRememberedAttributes.fill( false );
+    //load remember values or create new entry
+    if( mRememberings.contains( mLayer ) ){
+      mFeature = mRememberings[mLayer].rememberedFeature;
+    }else{
```
m-kuhn: Can you enable the pre-commit hook to keep coding style consistent with QGIS?

src/core/featuremodel.h
```diff
@@ -38,6 +38,13 @@ class FeatureModel : public QAbstractListModel
     Q_PROPERTY( QString positionSourceName READ positionSourceName WRITE setPositionSourceName NOTIFY positionSourceChanged )
     Q_ENUMS( FeatureRoles )
 
+
+  struct RememberValues
```
m-kuhn: API docs

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-2972f1f7d2d3c359f6940df777c24315cb6ad14d-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-2972f1f7d2d3c359f6940df777c24315cb6ad14d-x86.apk)

giedrioks26:
Test apk works perfectly. I've been waiting for this for a long time! Thank you!

src/core/featuremodel.cpp
```diff
@@ -46,15 +46,26 @@ void FeatureModel::setCurrentLayer( QgsVectorLayer* layer )
   if ( layer == mLayer )
     return;
 
+  //remember the last feature of the old layer
```
m-kuhn: Do we need something like this?

```
slot FeatureModel::onLayerRemoved( layer )
{
  mRememberings.remove( layer );
}

connect ( mLayer, destroyed, ..., Qt::UniqueConnection );
```

signedav: I made it now in a different way. After project is loaded I reset the map. I liked your approach with the layers, but when the model never is destroyed the layer is never destroyed as well, right?

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-e988d1f1bab73cac38f78bbd0e0aa6f3231ea9b0-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-e988d1f1bab73cac38f78bbd0e0aa6f3231ea9b0-x86.apk)

src/core/featuremodel.cpp
m-kuhn: The layer should be destroyed on project unload, it's owned by the project, not by this model.
To be honest, I am not a fan of giving the responsibility for clearing this to QML code.

signedav: you are right. than it's even simpler. thanks.

src/core/featuremodel.cpp
```diff
@@ -53,6 +53,9 @@ void FeatureModel::setCurrentLayer( QgsVectorLayer* layer )
   }
 
   mLayer = layer;
+
+  connect(mLayer,&QgsVectorLayer::destroyed,this,[=]{ mRememberings.clear(); }, Qt::UniqueConnection);
```
m-kuhn: Qt::UniqueConnection does not work with lambdas unfortunately... It really needs to be a slot in this case.

m-kuhn: And just to be on the safe side for the future if we decide to load/unload layers individually (and not always in complete projects) I'd just remove the layer which has been destroyed.

m-kuhn: And check your pre-commit hook ;)

signedav: Haven't knew that. I'll change it.

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-b916c03cc0efef3f2aed843e85bf296ec78696f4-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-b916c03cc0efef3f2aed843e85bf296ec78696f4-x86.apk)

src/core/featuremodel.cpp
signedav: @m-kuhn Probably you have a quick input for my problem because I guess you used destroyed signals before, otherwise tell me to investigate more myself, but I'm feeling to loose time here:
```
void FeatureModel::removeLayer( QObject* layer )
{
  mRememberings.remove( qobject_cast< QgsVectorLayer * >( layer ) );
}

connect( mLayer, &QgsVectorLayer::destroyed, this, &FeatureModel::removeLayer );
```
This does not remove the layer from the map. I don't know why. It's hard to find out in the debugger what `layer` does contain. `mRememberings.contains( qobject_cast< QgsVectorLayer * >( layer ) ) ` is false as well. Means the  `layer`  is not the  `mLayer` we stored in remember map.

It's possible that this is because the derived object (QgsVectorLayer) is actually destroyed before the signal is emitted. But not sure. I tried another possiblity, hoping it keeps the layer, but it's not even arriving in the slot (possibly not emitting the signal:
```
void FeatureModel::removeLayer( QgsVectorLayer* layer )
{
  mRememberings.remove( layer );
}
  connect( mLayer, SIGNAL(destroyed(QgsVectorLayer*)), this, SLOT(removeLayer(QgsVectorLayer*)));
```

signedav: Seems to work with `static_cast`, thanks.
If I do understand it correctly, the `qobject_cast` only gives an object, if there as been a proper cast and `static_cast` gives us just a converted zombie-object that is usually not usable at all but for this case it is, because the pointer address stays the same. Correct?

m-kuhn: `qobject_cast` is pretty much like `dynamic_cast`, they do actually check (at runtime) if the object can be converted to a given subclass.

`static_cast` checks at build time if this is something you can do, but doesn't do any checks at runtime.

E.g.

```
QObject *myobject = new QWidget();

// this is returning a nullptr, because it looks at myobject and realizes that myobject is actually a QWidget which cannot be converted to a QgsVectorLayer
QgsVectorLayer *layer = qobject_cast<QgsVectorLayer*>(myobject);

// QgsVectorLayer is a subclass of QObject, so static_cast will just do the conversion. Even if in this case it's a Widget which is not safe to cast to a QgsVectorLayer. Any attempt to use layer-specific methods on QgsVectorLayer result in undefined behavior (in most cases a crash)
QgsVectorLayer *layer = static_cast<QgsVectorLayer*>(myobject);
```

signedav: Thanks @m-kuhn

qfield-fairy:
Uploaded test apks for [armv7](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-7e0addafe8e25f5a7cfbf1b14b590187aff8b209-armv7.apk) and [x86](https://opengis.ch/download/qfield/ci-builds/qfield-dev-381-7e0addafe8e25f5a7cfbf1b14b590187aff8b209-x86.apk)

m-kuhn:
Thanks @signedav

PR #404, opened 2018-12-07
finer tuning of QML date/time edit

3nids:
* do not open calendar when option is not checked in QGIS project
* use inputMethodHints whenever easily possible, also use regex validator

fix #9
fix #215
fix #371

PR #421, opened 2018-12-18
correctly convert date/time format to respect field and display formats

### LINKED ISSUE #9: DateTime field edit type
I have installed QField 0.3.2 on a Samsung Galaxy Note 8.0. I have a project with a shapefile layer. One of the fields, called "Time" is defined as Date (QDAte) and widget is set to display "yyyy-MM-dd  HH:mm:ss", two other fields are called respectively Start_Date and End_Date, are defined as Date (QDate) and widget is set to display "yyyy-MM-dd" only.
However all fields do not automatically update when opened in QField editor, so that theoretically I'd have to insert all times and dates manually. On the other hand in QGIS for desktop everything is working fine.
When I use a spatialite layer the behavior is the same.

It may be useful to say that if a field was edited in QGIS (2.8.1), when opened in QField it displays as 2015-03-25T00:00:00, even if time was correct in QGIS. May be there a problem when copying the project file? Should I export projects and shapefiles differently, instead of just copying and pasting?

### LINKED ISSUE #215: Calendar widget
In QGis (2.18.16), whether calendar option is checked or not in Date/Time widget, it pops up in QField (0.10.4) when clicking on the field.

### Expected behavior:

Calendar pops up if checked in date/time option in QGis ; doesn't pop up if not checked

## Additional information:

Maybe related to answer given in issue #9 (which would be nice to solve too ;) )

Thanks to all the team : you make great work !

### LINKED ISSUE #371: Entry of time only (without date)
I'd like to only save time into a geopackage. As geopackage only knows date/time format I chose a text as the field type in order to only save time (and not also date). Then I define the attribute form the following way: 
zeittext

On QGIS 3.2 I the data entry is presented like this.
zeittextqgis

On QField I am shown an calendar widget upon data entry, which is of course useless for storing time. I would expect something like this as well:
zeittextqgis

**QField version:** [0.11.6]
**QQGIS version:** [3.2]

## Additional information:

3nids:
these formats are defined in QGIS and can differ from each other
more over, the type of value from the field can be either a date or a string

by correctly constructing a date/time object when getting/setting the value from
the widget, the same behavior than in QGIS is allowed

also:
* do not open calendar when option is not checked in QGIS project
* define inputMask whenever easily possible

fix #9
fix #215
fix #371 

supersedes #404

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
   - gem --version
 
 script:
-  - export QFIELD_SDK_VERSION=20190123
+  - export QFIELD_SDK_VERSION=20190128
   - echo "travis_fold:start:docker-pull"
   - docker pull opengisch/qfield-sdk:${QFIELD_SDK_VERSION}
   - echo "travis_fold:end:docker-pull"
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,21 +27,27 @@ script:
   - ./scripts/upload-artifacts.sh
 
 jobs:
+  allow_failures:
+    - testing
   include:
     - stage: test
+      name: testing
       script:
         - echo "travis_fold:start:build\n$(tput bold)Build QField $(tput sgr0)"
         - docker-compose -f .docker/testing/docker-compose-travis.yml run qgis /usr/src/.docker/testing/build-test.sh
       if: type = pull_request OR tag IS present OR branch = master
     - stage: test
+      name: build ARMv7
       env:
         - ARCH=armv7
       if: type = pull_request OR tag IS present OR branch = master
     - stage: test
+      name: build X86
       env:
         - ARCH=x86
       if: type = pull_request OR tag IS present OR branch = master
     - stage: deploy
+      name: "Deploy 🍺"
       if: type = pull_request OR tag IS present OR branch = master
       script:
         - pip install pyopenssl
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
 
 jobs:
   allow_failures:
-    - testing
+    - name: testing
   include:
     - stage: test
       name: testing
@@ -58,3 +58,6 @@ jobs:
       install:
         pip install -r requirements.txt
       script: ./scripts/ci/update-translations.sh
+
+matrix:
+  fast_finish: true 
--- a/src/core/featuremodel.cpp
+++ b/src/core/featuremodel.cpp
@@ -22,7 +22,7 @@
 #include <qgsmessagelog.h>
 #include <qgsvectorlayer.h>
 #include <QGeoPositionInfoSource>
-#include <QDebug>
+
 
 FeatureModel::FeatureModel( QObject *parent )
   : QAbstractListModel( parent )
@@ -128,7 +128,6 @@ QVariant FeatureModel::data( const QModelIndex& index, int role ) const
       break;
 
     case AttributeValue:
-      qDebug() << mFeature.attribute( index.row() );
       return mFeature.attribute( index.row() );
       break;
 
--- a/src/core/featuremodel.cpp
+++ b/src/core/featuremodel.cpp
@@ -128,6 +128,7 @@ QVariant FeatureModel::data( const QModelIndex& index, int role ) const
       break;
 
     case AttributeValue:
+      qDebug() << mFeature.attribute( index.row() );
       return mFeature.attribute( index.row() );
       break;
 
--- a/src/core/attributeformmodel.h
+++ b/src/core/attributeformmodel.h
@@ -26,7 +26,7 @@ class AttributeFormModel : public QSortFilterProxyModel
 {
     Q_OBJECT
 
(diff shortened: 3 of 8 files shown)


3nids:
@saemiw would be good to get your approval here

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-421-89e69c5f9f6e53a6837984734267ba54a38b944c-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-421-89e69c5f9f6e53a6837984734267ba54a38b944c-x86.apk)

saemiw:
@3nids everything looks fine upon data entry. The data are however not saved. When I revisit a feature that I already entered before I can't see the entered time. AND when I try to modify the time it will not accept my entries (field remains empty). The only time that is accepted and written to the geopackage file is when it is entered via a now()-default-value.

*Entry before saving anything*
screenshot_20181219-083829_qfield

*This is what is saved when re-visiting the feature*
screenshot_20181219-083852_qfield

3nids:
@saemiw thanks a lot for the report. I had only tested on editing on current features, and I thought it was working while not.
It seems the widget doesn't update itself properly (keep values of the first open feature).
Still on progress

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-421-21452e7896ffd47386136fb4faea199b329b314d-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-421-21452e7896ffd47386136fb4faea199b329b314d-x86.apk)

saemiw:
I did a quick test on the latest test apk. Data entry now seems to work fine. But when I modifie (change) a time-entry the attribute value is changed in all features instead of just the one that I wanted to modify.

src/core/featuremodel.cpp
```diff
@@ -128,6 +128,7 @@ QVariant FeatureModel::data( const QModelIndex& index, int role ) const
       break;
 
     case AttributeValue:
+      qDebug() << mFeature.attribute( index.row() );
```
m-kuhn: Remove

3nids:
Here is a new version that should work.

There is one case I fail to solve easily or properly, when using a shapefile, which has date type only.

QGIS passes a QDate object which is translated to a JS Date object, which has no distinction for dates only.
Hence, the new date gets a time zone.
So a date coming from the shapefile as `2001-01-01` will become `2000-12-31 19:00:00 -05` (in the carribeans time zone ;) ) in QML/JS.
And when formatting this to a date only, this is shown as `2000-12-31`.

This issue doesn't occur with gpkg as field has a date/time type.
I am not really sure how and where to handle this.

@saemiw could you give another try at this one?

3nids:
@saemiw sorry, this will need an update of the APK before testing

saemiw:
@3nids now we're talking! Everything works as expected. Great. I'll ask my work collegue to test it as well, as she was primiraly asking for that feature. But don't expect major problems. Cool.

3nids:
@saemiw thanks a lot for testing!

.travis.yml
```diff
@@ -19,29 +19,35 @@ before_install:
...
+  - export QFIELD_SDK_VERSION=20190128
   - echo "travis_fold:start:docker-pull"
   - docker pull opengisch/qfield-sdk:${QFIELD_SDK_VERSION}
   - echo "travis_fold:end:docker-pull"
   - docker run -v $(pwd):/usr/src/qfield -e "BUILD_FOLDER=build-${ARCH}" -e "ARCH=${ARCH}" -e "STOREPASS=${STOREPASS}" -e "KEYNAME=${KEYNAME}" -e "KEYPASS=${KEYPASS}" -e "VERSION=${TRAVIS_TAG}" opengisch/qfield-sdk:${QFIELD_SDK_VERSION} /usr/src/qfield/scripts/docker-build.sh
   - ./scripts/upload-artifacts.sh
 
 jobs:
+  allow_failures:
+    - name: testing
```
m-kuhn: what's this for?

3nids: test is currently failing because of a missing declare metatype in QGIS core (PR merged, waiting for new image to be built)

3nids:
merging this as is, we will try to fix the shapefile issue in a later step.

3nids:
merging as is, we will try to fix the shapefile issue in another step.

saemiw:
my collegue suggest also one more change (maybe also in another step/together with the shapefile issue-fix): When you enter non-sense like 30:00:00 in a hh:mm:ss-field QField does show these numbers in grey (instead of black) and then doesn't save them when you save the feature. I would be better when an non-sense entry like 30:00:00 would immediately disappear after pushing the o.k. button instead of showing the wrong entry in grey. Is this doable? Its still good to incorporate the current solution but maybe improve it later.

3nids:
fix or shapefiles in #458 

@saemiw good idea, I'll see if I can do it in a second step...

PR #440, opened 2019-01-17
fix crash in feature list model

### LINKED ISSUE #432: Qfield crashes on polyline select
Qfield crashes with polyline select

When doing the following steps I could repeat the behaviour on Samsung Galaxy TAB Active 2 with Android 8.1 installed

1. Open project attached
2. Zoom to polyline and select
3. Browse polyline information and observe crash

Qfield RC1 is installed. Happened with previous release also. If further information is required please advise how to gather it.

Best

Sebastian 

[qgep_en_201812291341.zip](https://github.com/opengisch/QField/files/2737430/qgep_en_201812291341.zip)

3nids:
fixes #432 

I am not sure this is the correct fix (i.e. is there an issue hidden behind) but it does prevent the crash.
My wild guess is that with the project in the issue report (QGEP), rendering takes more time than the timer (100ms) so everything is not correctly set when the timer ends, but I didn't look further.

@m-kuhn thoughts?

### DIFF
--- a/src/core/featurelistmodel.cpp
+++ b/src/core/featurelistmodel.cpp
@@ -25,7 +25,7 @@ FeatureListModel::FeatureListModel()
   connect( &mReloadTimer, &QTimer::timeout, this, &FeatureListModel::processReloadLayer );
 }
 
-QModelIndex FeatureListModel::index( int row, int column, const QModelIndex& parent ) const
+QModelIndex FeatureListModel::index( int row, int column, const QModelIndex &parent ) const
 {
   Q_UNUSED( column )
   Q_UNUSED( parent )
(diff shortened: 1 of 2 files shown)


src/core/featurelistmodel.cpp
```diff
@@ -147,6 +147,9 @@ void FeatureListModel::processReloadLayer()
 {
   mEntries.clear();
 
+  if ( !mCurrentLayer )
+    return;
```
3nids: my changes are actually here!

m-kuhn:
Yes, it's probably not the best fix, but I can't come up with a better one right now. It should probably be connected to a signal instead of the timer. Just a quick check: are there a lot of ValueRelations on this layer? They take a long time to load.

Maybe retrigger the timer in this case? If your suspicion is correct, that should make the behavior still work but slower.

schoeller:
I have downloaded and installed the apk from https://github.com/opengisch/QField/pull/440#issuecomment-455028295. When opening the dataset from https://github.com/opengisch/QField/issues/432#issue-396967243 on the Android 8.1 device I cannot register any more crashes. Thanks

3nids:
@m-kuhn thanks for the feedback
no every thing is ready at that time.
The issue is that the given project has invalid relations (i.e. missing layers).
I will provide a better fix asap

3nids:
here we go
screenshot 2019-01-21 at 20 55 35

PR #443, opened 2019-01-21
Use QML to render points

### LINKED ISSUE #156: Selection issues with point features
Hello, first of all we wanted to let you know how much we appreciate the lots of benefits Qfield delivers compared to other field data survey collection systems, despite some issues that are being constantly addressed in this thread. So really thank you for the great work!

Right now we are heavily testing Qgis & Qfield capabilities for field work, and therefore in the following weeks we can inform you on some of the main issues we have noticed.

To begin with, we wanted to inform you there is a very simple but quite useful command that doesn’t work properly: the selection on point data.

We are using Qgis v. 2.18.7 and Qfield v. 0.10.0 and we observed that selections attained by pressing point features (in Qfield):

>directly on screen or
>through attributes selection in the right area (where tabular data is shown) 

aren’t highlighted in the map, as it correctly occurs with polyline features. 
When you press a line or scroll in the attributes table the different selected lines, the chosen feature in the map rightly changes colour in association with the shown record, so that you can easily associate data to a specific location. 
For point features, instead, it is impossible to understand which record in the attribute table is associated to the specific points shown in the map, as no highlighting occurs.
We believe this is a very useful information when you operate in field; we didn’t try yet with polygon features, so we don't know if it works fine or not.

To solve the issue we tried to modify options in Qgis under Settings - Options menu (highlight color, buffer, minimun length) before uploading the project to Qfield, but it looks like these settings aren't being properly transmitted to Qfield or, maybe, we just didn't get it right and this is not the point to sort things out.

Did you also notice this issue or do you have any suggestions to solve this problem?

Thank you very much and Best Regards

The Tethys GIS-team

### LINKED ISSUE #435: Search functionality does not highlight point type layer element
Search functionality does not highlight point type layer element.

Performing a search on a specific feature of a point type layer, **only** centers the element on the equipment screen.

screenshot_20190109-165518

On the other hand, when the research occur in a  row-type layer, qfield, besides centering the element on the equipment screen, **increases the scale allowing an approximation to it** and **highlights the element** with the color of selection (defined in the workspace project: General Settings).

screenshot_20190109-161142

It would be convenient if the layer of the type of point would have the same potentialities in the level of its research..

Best

Carlos

3nids:
This adds a new `GeometryRenderer` item to render geometries.
Points are rendered using QML rectangles items, while line/polygons are forwarded to to cpp to be rendered using `QSGNode`. 

This solves issue of points being highlighted as 1 pixel only (on non-openGL only?)

It is now used both for the locator results and the multi selection highlighter.

Fixes #435 
Fixes #156


src/core/qgssggeometry.cpp
```diff
@@ -183,12 +185,22 @@ QSGGeometry *QgsSGGeometry::qgsPolygonToQSGGeometry( const QgsPolygonXY &polygon
...
+      double rad = i * 360 / n * M_PI / 180.0;
+      geometry->vertexDataAsPoint2D()[i].set( center.x() + std::cos( rad ) * radius,
+                                              center.y() + std::sin( rad ) * radius );
+    }
+  };
+
+
+  QSGGeometry *sgGeom = new QSGGeometry( QSGGeometry::defaultAttributes_Point2D(), n );
+  drawCircle( sgGeom, point, 5 * width );
+  sgGeom->setDrawingMode( QSGGeometry::DrawTriangleFan );
```
3nids: shall we drop this code to be sure that we never use this? and only use QML to render points?

src/core/linepolygonhighlight.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+void LinePolygonHighlight::setGeometry( QgsGeometryWrapper *geometry )
+{
+  if ( mGeometry == geometry )
+    return;
+
+  Q_ASSERT( !geometry || geometry->qgsGeometry().type() != QgsWkbTypes::PointGeometry );
+
+  if ( mGeometry )
+    mGeometry->deleteLater();
```
3nids: is this the way to go?

src/core/qgssggeometry.cpp
m-kuhn: yes

src/qgsquick/qgsquickmapcanvasmap.h
```diff
@@ -161,7 +161,7 @@ class QgsQuickMapCanvasMap : public QQuickItem
      * Refresh the map canvas.
      * Does nothing when output size of map settings is not set
      */
-    void refresh();
+    Q_INVOKABLE void refresh();
```
m-kuhn: Is this needed? Are slots not callable anyway?

src/core/qgsgeometrywrapper.cpp
```diff
@@ -0,0 +1,46 @@
+
```
m-kuhn: license header

m-kuhn:
Good stuff.

I am not sure I completely get, how the geometry wrapper works.
Is a new wrapper created from a new geometry and then passed on and deleted later on and the wrapped geometry never changes?
In this case I would assume we could just pass the geometry (gadget) itself around and not need a QObject based subclass like the wrapper.
I would imagine a wrapper to be instantiated once (in QML) and then the wrapped geometry property inside changed.

src/core/qgssggeometry.cpp
3nids: done

src/qgsquick/qgsquickmapcanvasmap.h
3nids: leftover, removed!

3nids:
Indeed, I am replacing the wrapper.

Mainly because in the locator filter (see https://github.com/opengisch/QField/pull/443/files#diff-40e6fb81d790b248689e34b3bc7724c6R170), I am doing this:

    QgsGeometryWrapper *geomwrapper = new QgsGeometryWrapper( geom, layer->crs() );
    mLocatorBridge->locatorHighlight()->setProperty( "geometry", QVariant::fromValue<QgsGeometryWrapper *>( geomwrapper ) );

I think we cannot use the basic QgsGeometry (gadget) because it needs to be a property of a QML item.

I started to implement the change to changing the properties of the wrapper instead of changing the wrapper, but it brings a bit more code and makes things less clear. I can go on this tomorrow, I'm just not sure what's the best approach.

3nids:
ok, now using the properties of the wrapper instead of resetting the wrapper itself

m-kuhn:
:tada:

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-443-1d06f861de957dc07ea7608c4469c1700c5b7a54-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-443-1d06f861de957dc07ea7608c4469c1700c5b7a54-x86.apk)

PR #445, opened 2019-01-21
avoid overlap of busy indicator and search icon

3nids:
move the busy indicator to the center of the map
it is not that much obtrusive

image


signedav:
Looks okay. But how is the size on the different devices? Did you checked?

3nids:
nope, but it shouldn't change from current situation, it's just moved in the center

PR #464, opened 2019-01-30
clear value in date/time if invalid when losing focus

3nids:
@saemiw this should fix the issue of leaving an invalid date/time after the widget loses focus.
Works for you?

### DIFF
--- a/src/qml/editorwidgets/DateTime.qml
+++ b/src/qml/editorwidgets/DateTime.qml
@@ -118,20 +118,30 @@ Item {
         }
 
         onActiveFocusChanged: {
-            if (activeFocus) {
-              var mytext = label.text
-              var cur = label.cursorPosition
-              while ( cur > 0 )
-              {
-                if (!mytext.charAt(cur-1).match("[0-9]") )
-                  break
-                cur--
-              }
-              label.cursorPosition = cur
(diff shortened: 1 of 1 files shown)


qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-464-ce9463bd5bde6a889e70f12a8734a2bc906ec7e4-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-464-ce9463bd5bde6a889e70f12a8734a2bc906ec7e4-x86.apk)

saemiw:
@3nids Yes that works. 

If it is easy to change it would however be better to change it the following:
Currently the non-sense time is removed as soon as you tap into the next attribute field (which is an improvement). If the time is however your last field you enter, you won't notice that non-sense entries will be removed. 
So it would be better to remove the non-sense entry directly after pushing the o.k. button (circled in blue in the pic below). Like this you would immediately notice that QField is not accepting the non-sense entry.

screenshot_20190131-095654_qfield

3nids:
it's a bit more complicated to handle this, I will try.
Merging it now to get it in RC4.

PR #498, opened 2019-03-04
fix crash when switching layer

### LINKED ISSUE #474: Replicable crash on QField RC3
I found a way to crash consistently QField RC3. Can provide sample project if necessary.

### Reproduction Steps:

1. open a project with two layers from the same GPKG datasource (possibly affected other datasources, not tested), one point, one polygon
2. (optional step) enable editing for polygons, add a feature, save --> OK
3. enable editing for the point layer
4. opsss... I wanted to edit the polygons --> enable editing for polygons
5. add a polygon, save --> crash

### LINKED ISSUE #477: Crash after layer switching
I do have a QGIS 3.4.4 Project, use QField Sync 3.1.1 and QField 1.0.0-rc3 to set it offline. With some layers that were used the QFieldSync action "copy" I have problems. Upon selecting that layer, then selecting an offline layer and trying to add a new feature in this offline layer QField crashes. 

First I thought I was able to very consistently (like 10 times in a row) reproduce the crash. But now suddenly it crashes and sometimes not. I first thought it was tied to the layer options "searchable" or "editable" but I get the crashes with and without this options.  I also noticed that it happens with geopackage and shapefile.  The only thing clear seems that it doesn't happen with layers that were set offline (by QFieldSync). I am happy to provide my project through wetransfer if you like testing your self. I keep you updated if I find out more.

3nids:
a crash was hapenning when switching between layers having tabbed and non-tabbbed attribute form
there was a count mismatch between swipe view and tab bar

fixes #474
fixes #477 

kudos to @m-kuhn for digging into this nasty crash
and thanks to reporters with sample projects!

### DIFF
--- a/src/qml/FeatureForm.qml
+++ b/src/qml/FeatureForm.qml
@@ -91,7 +91,7 @@ Page {
         }
 
         Repeater {
-          model: form.model
+          model: form.model && form.model.hasTabs ? form.model : 0
 
           TabButton {
             id: tabButton


gioman:
@3nids thanks! much appreciated.

PR #525, opened 2019-03-16
Legend padding

mbernasocchi:
a TreeView can take a `Styles.TreeViewStyle` as a property. `TreeViewStyle` has a `branchDelegate` which controls the timage that changes when a group is expanded. this is a branch to tryout a bit in that direction.

https://doc.qt.io/qt-5/qml-qtquick-controls-styles-treeviewstyle.html#branchDelegate-prop

### DIFF
--- a/src/qml/DashBoard.qml
+++ b/src/qml/DashBoard.qml
@@ -246,12 +246,16 @@ Controls.Pane {
     RowLayout {
         id: layout
         Layout.fillWidth: true
+        Layout.fillHeight: true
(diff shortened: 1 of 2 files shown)


mbernasocchi:
@m-kuhn just so you know and don't duplicate work. I'll play a bit mode Monday, but if you see an immediate solution, don't' hesitate, I'm super slow

mbernasocchi:
https://i.imgur.com/USXU38c.jpg

PR #529, opened 2019-03-18
Grant sd card

marioba:
Grant write permissions on sd-card

### DIFF
--- a/android/src/ch/opengis/qfield/QFieldProjectActivity.java
+++ b/android/src/ch/opengis/qfield/QFieldProjectActivity.java
@@ -1,6 +1,7 @@
 package ch.opengis.qfield;
 
 import java.io.File;
+import java.io.PrintWriter;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -29,6 +30,8 @@ import android.widget.AdapterView.OnItemLongClickListener;
 import android.graphics.Color;
 import android.graphics.drawable.ColorDrawable;
 import android.text.TextUtils;
+
+import ch.opengis.qfield.QFieldActivity;
 
 public class QFieldProjectActivity extends Activity {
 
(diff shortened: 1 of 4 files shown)


android/src/ch/opengis/qfield/QFieldProjectActivity.java
```diff
@@ -253,6 +296,21 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         Log.d(TAG, "onActivityResult ");
         Log.d(TAG, "resultCode: " + resultCode);
 
+        Log.d(TAG, "data: " + data.getDataString());
+        Log.d(TAG, "requestCode: " + requestCode);        
+
+        // This is the result of ACTION_OPEN_DOCUMENT_TREE 
+        if(requestCode == 666){
```
m-kuhn: :imp:

mbernasocchi: seriously!? they really don't want people to use files...

android/src/ch/opengis/qfield/QFieldProjectActivity.java
```diff
@@ -1,6 +1,7 @@
 package ch.opengis.qfield;
 
 import java.io.File;
+import java.io.PrintWriter;
```
mbernasocchi: Leftover?

android/src/ch/opengis/qfield/QFieldProjectActivity.java
marioba: It's a code I chose. But it seemed to me to be an appropriate number.

marioba:
@mbernasocchi @m-kuhn Please don't waste time checking this code now. At the moment it's just a not working draft.

mbernasocchi:
By @marioba in internal comunication:

> Update on the sd-card authorization problems:
I realized why I couldn't use correctly the intents `ACTION_OPEN_DOCUMENT` and `ACTION_OPEN_DOCUMENT_TREE`.

>The intents `ACTION_OPEN_DOCUMENT` and `ACTION_OPEN_DOCUMENT_TREE`, work and grant read-write access on the content of the returning URI, not on the possible real linked file. That is, if I call an intent `ACTION_OPEN_DOCUMENT` and select the file `proba.txt` on my sd-card, I will receive as an answer a URI like this: `content://com.android.externalstorage.documents/document/9C33-6BBD%3Aproba.txt`, at this point if I create a `DocumentFile` object from this uri, I can get an outputStream that allows me to write in the file. If instead I try to access the same file but through its real path (i.e. `/storage/9C33-6BBD/proba.txt`), I never have the permission to write to it. It seems that the permissions are managed on the URIs and not on the real files, which makes sense for how Android is structured but at the moment it is unusable with QField as we need to access the real file.

> At the moment, unless we find other ways to do it (I think very unlikely), and until we can get QField to work with URIs, I don't think we can do more than what is done currently in the file selector (read-write application directory on sd-card and the rest of the sd-card read only).

marioba:
For cleaning purpose, I delete the branch because I don't see the need to keep it open. Once understood what is indicated in the comment above, the implementation is trivial and the experiments of the branch are of little use.

m-kuhn:
Thanks for all the info @marioba 
Adapting QField (with gdal, ogr, Qt etc) to work on URI's will be a huge task, so sticking to what we have sounds like the way to go for now.

PR #542, opened 2019-03-26
Checkbox widget with boolean fields

signedav:
Checkbox has been not working for Boolean fields (only for other fields used as checkbox) I believe this issue existed since we used QGIS 3.0.

Because since then the widget needs to decide if the value is a boolean or we use a representing value.
This is done in the QGIS widget here: https://github.com/qgis/QGIS/blob/master/src/gui/editorwidgets/qgscheckboxwidgetwrapper.cpp#L29-L35

I did it now in the Checkbox widget here as well. 

Since I guess I'm not able to request a type in QML I did it like this. Another possibility would be to request the value over the model, but probably that would be an overkill as well. What do you think  @m-kuhn ?

### DIFF
--- a/src/qml/editorwidgets/CheckBox.qml
+++ b/src/qml/editorwidgets/CheckBox.qml
@@ -12,17 +12,18 @@ Item {
 
   CheckBox {
     property var currentValue: value
+    property bool isBool: value === true || value === false
 
-    checked: value == config['CheckedState']
+    checked: isBool ? value : value === config['CheckedState']
 
     onCheckedChanged: {
-      valueChanged( checked ? config['CheckedState'] : config['UncheckedState'], false )
+      valueChanged( isBool ? checked : checked ? config['CheckedState'] : config['UncheckedState'], false )
       forceActiveFocus()
     }
 
     // Workaround to get a signal when the value has changed
     onCurrentValueChanged: {
-      checked = currentValue == config['CheckedState']
+      checked = isBool ? currentValue : currentValue === config['CheckedState']
     }
 
     indicator.height: 16 * dp
--- a/src/qml/editorwidgets/CheckBox.qml
+++ b/src/qml/editorwidgets/CheckBox.qml
@@ -12,7 +12,8 @@ Item {
 
   CheckBox {
     property var currentValue: value
-    property bool isBool: value === true || value === false
+    //if the field type is boolean, ignore the configured 'CheckedState' and 'UncheckedState' values and work with true/false always
+    property bool isBool: field.type == 1
 
     checked: isBool ? value : value === config['CheckedState']
 
--- a/src/qml/editorwidgets/CheckBox.qml
+++ b/src/qml/editorwidgets/CheckBox.qml
@@ -13,7 +13,7 @@ Item {
   CheckBox {
     property var currentValue: value
     //if the field type is boolean, ignore the configured 'CheckedState' and 'UncheckedState' values and work with true/false always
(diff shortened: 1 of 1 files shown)


src/qml/editorwidgets/CheckBox.qml
```diff
@@ -12,17 +12,18 @@ Item {
 
   CheckBox {
     property var currentValue: value
+    property bool isBool: value === true || value === false
```
m-kuhn:
```suggestion
    readonly property bool isBool: field.type === 1
```

m-kuhn: And a comment (something like "If the field type is boolean, ignore the configured 'CheckedState' and 'UncheckedState' values and work with true/false always")

m-kuhn:
Good job!

src/qml/editorwidgets/CheckBox.qml
signedav: That's not working:

image

signedav: But yes - that's the part I don't like neither. 

But in the end it's a hack that is understandable enough that we can keep it IMO.

m-kuhn: `field.type == 1`?

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-542-1689ac75fb9895e1834469f8f38180311c49ad87-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-542-1689ac75fb9895e1834469f8f38180311c49ad87-x86.apk)

src/qml/editorwidgets/CheckBox.qml
m-kuhn: ... I think we need this approach or if the default value is NULL it will fail to detect the type.

signedav: that works

m-kuhn: great, with Qt 5.13 apparently enum support will be improved, whatever that means, let's keep it in mind

src/qml/editorwidgets/CheckBox.qml
```diff
@@ -12,17 +12,19 @@ Item {
 
   CheckBox {
     property var currentValue: value
+    //if the field type is boolean, ignore the configured 'CheckedState' and 'UncheckedState' values and work with true/false always
+    property bool isBool: field.type == 1
```
m-kuhn:
```suggestion
    readonly property bool isBool: field.type == 1
```

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-542-ffb4c5947e7e53f9028a756bfc83767bdb92bce8-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-542-ffb4c5947e7e53f9028a756bfc83767bdb92bce8-x86.apk)

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-542-e6b6dd759a85946e92833a5a74d154a526e2fe16-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-542-e6b6dd759a85946e92833a5a74d154a526e2fe16-x86.apk)

m-kuhn:
Tested and works

PR #551, opened 2019-03-27
Add day when it sets time back

signedav:
When we selected a day, it stored one day before. This because the time is set to 0:00:00.
Now with adding a day we have the correct date.

Not sure if this is the correct way to solve the issue. What do you think @3nids ?

### DIFF
--- a/src/qml/editorwidgets/DateTime.qml
+++ b/src/qml/editorwidgets/DateTime.qml
@@ -187,6 +187,8 @@ Item {
               newDate.setMinutes(0)
               newDate.setSeconds(0)
               newDate.setMilliseconds(0)
+              // because this sets us back one day we add one day
+              newDate.setDate(newDate.getDate()+1)
 
               if ( main.isDateTimeType )
               {
--- a/src/qml/editorwidgets/DateTime.qml
+++ b/src/qml/editorwidgets/DateTime.qml
@@ -183,12 +183,6 @@ Item {
             onClicked: {
               // weird, selectedDate seems to be set at time 12:00:00
               var newDate = calendar.selectedDate
-              newDate.setHours(0)
-              newDate.setMinutes(0)
-              newDate.setSeconds(0)
-              newDate.setMilliseconds(0)
-              // because this sets us back one day we add one day
-              newDate.setDate(newDate.getDate()+1)
 
               if ( main.isDateTimeType )
               {


3nids:
hmmm, that seems very strange.
Can you try by removing all the lines under the "weird" comment?
Maybe, I was confused and 12:00:00 is actually the very end of the previous day and it would work.

But this has to be very carefully checked (see comment around[ line 73](https://github.com/opengisch/QField/blob/bce240754925a57fd4030b1e7e6b4bbd888d3565/src/qml/editorwidgets/DateTime.qml#L73-L79)).

signedav:
It would work, when I remove them. One minute.

signedav:
That would work for my tests. But I didn't tested with shp etc. What tests are needed to be sure it does not break something? 

I tested with gpkg and the online pg db...

3nids:
sadly there are no available test suite for this...
I have a project I can test along.
Let me try later today

signedav:
I think, we need to release. Do you prefer to merge that without your tests or should we not?

3nids:
I don't have a testing env at the moment (no compiled QGIS).

Here are the possibilities:
- string field as date (no matter source type)
- string field as date AND time (no matter source type)
- date field on GPKG with time
- date field on shape (has no time)

and testing adding, manually editing and editing using calendar.

The main important one is shape and string as date and time.

3nids:
my general impression is that there should not be issue...but my experience with date time is that I am very ofter wrong.

signedav:
shp:
date: fine
date-time: fine 12.00.00
string: fine

gpkg:
date: fine
date-time: fine 11.00.00 (who knows why)
string: fine

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-551-abd2007afc2ffba74e3e9ca9bb4d05e7a2f27d82-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-551-abd2007afc2ffba74e3e9ca9bb4d05e7a2f27d82-x86.apk)

3nids:
@signedav @mbernasocchi this is not working as expected. It gets to 11:00:00 because of the UTC shift and the bad conception on QML side (there is no Date only object, it's always date and time)

this requires more work

3nids:
Hmmm, I guess this was working for me and not for you (a day before) because I'm on the other side of Greenwich.
I think we need to shift the value by the UTC offset.

signedav:
We definitely need to invest more work on it, but in the end it's better to get the correct date with a random time than a wrong date with the 0 time, no?

mbernasocchi:
yes, let's care about date first of all

3nids:
I cannot reproduce the issue you mentioned with gpkg and date/time.

PR #567, opened 2019-04-15
Measure tool

signedav:
Based on digitizing rubberband.

(Pink button is only for test - no worries)

measurepleasure

The digitizing rubberband and the measure rubberband are seperated, means you can continue with the digitizing after you did a measurement.

- [x] Mode switch button
- [x] Make color better
- [x] Fix jumping back to last vetex on cancel (same for digitizing)

### DIFF
--- a/src/qml/DashBoard.qml
+++ b/src/qml/DashBoard.qml
@@ -70,7 +70,6 @@ Controls.Pane {
         spacing: 4 * dp
 
         Controls.ToolButton {
-          // dummy button, is hidden behind main button
           height: 48 * dp
           width: 48 * dp
 
@@ -107,12 +106,6 @@ Controls.Pane {
           onClicked: showMenu()
         }
 
-        Rectangle {
-          //empty space
-          height: 56 * dp
-          width: 56 * dp
-          color: mainColor
-        }
          Controls.Switch {
           id: modeswitch
           height: 56 * dp
@@ -164,6 +157,26 @@ Controls.Pane {
             }
           }
         }
+
+         Controls.ToolButton {
+           height: 56 * dp
+           width: 56 * dp
+
+           contentItem: Rectangle {
+             anchors.fill: parent
+             color: "hotpink"
+             Image {
+               anchors.fill: parent
+               fillMode: Image.Pad
+               horizontalAlignment: Image.AlignHCenter
+               verticalAlignment: Image.AlignVCenter
+               source: Style.getThemeIcon( 'ic_settings_white_24dp' )
+             }
+           }
+
+           onClicked: changeMode( "measure" )
+         }
+
       }
     }
 
--- a/src/qml/DashBoard.qml
+++ b/src/qml/DashBoard.qml
@@ -8,7 +8,6 @@ Controls.Pane {
   objectName: "dashBoard"
 
   signal showMenu
-  signal changeMode( string mode )
 
   property alias allowLayerChange: legend.enabled
   property alias currentLayer: legend.currentLayer
@@ -48,10 +47,6 @@ Controls.Pane {
   onCurrentLayerChanged: {
     if ( currentLayer.readOnly && stateMachine.state == "digitize" )
       displayToast( qsTr( "The layer %1 is read only." ).arg( currentLayer.name ) )
-  }
-
-  onChangeMode: {
-      stateMachine.state = mode
   }
 
   Component.onCompleted: focusstack.addFocusTaker( this )
--- a/src/qml/DashBoard.qml
+++ b/src/qml/DashBoard.qml
@@ -100,59 +100,60 @@ Controls.Pane {
 
           onClicked: showMenu()
         }
+      }
 
-         Controls.Switch {
-          id: modeswitch
-          height: 56 * dp
-          width: 56 *2 * dp
-           indicator: Rectangle {
-            implicitHeight: 36 * dp
-            implicitWidth: 36 * 2 * dp
-            x: modeswitch.leftPadding
-            radius: 4 * dp
-            color:  "#80CC28"
-            border.color: "white"
+
+      Controls.Switch {
+        id: modeswitch
+        height: 56 * dp
+        width: ( 56 + 36 )  * dp
+        anchors.right: parent.right
+        indicator: Rectangle {
+          implicitHeight: 36 * dp
+          implicitWidth: 36 * 2 * dp
+          x: modeswitch.leftPadding
+          radius: 4 * dp
+          color:  "#80CC28"
+          border.color: "white"
+          anchors.verticalCenter: parent.verticalCenter
+          Image {
+            height: parent.height
+            width: parent.width / 2
+            anchors.left: parent.left
             anchors.verticalCenter: parent.verticalCenter
-             Image {
-              height: parent.height
-              width: parent.width / 2
-              anchors.left: parent.left
-              anchors.verticalCenter: parent.verticalCenter
-              source: Style.getThemeIcon( 'ic_map_white_48dp' )
-            }
-             Image {
-              height: parent.height
-              width: parent.width / 2
-              anchors.right: parent.right
-              anchors.verticalCenter: parent.verticalCenter
-              source: Style.getThemeIcon( 'ic_create_white_24dp' )
-            }
-             Rectangle {
-              x: modeswitch.checked ? parent.width - width : 0
-              width: 36 * dp
-              height: 36 * dp
-              radius: 4 * dp
-              color:  "#64B5F6"
-              border.color: "white"
-              Image {
-                height: parent.height
-                width: parent.height
-                anchors.right:  modeswitch.checked ? parent.right : undefined
-                anchors.left:  modeswitch.checked ? undefined : parent.left
-                anchors.verticalCenter: parent.verticalCenter
-                source:  modeswitch.checked ? Style.getThemeIcon( 'ic_create_white_24dp' ) : Style.getThemeIcon( 'ic_map_white_24dp' )
-              }
-            }
+            source: Style.getThemeIcon( 'ic_map_white_48dp' )
           }
-          onPositionChanged: {
-            if ( checked ) {
-              changeMode( "digitize" )
-            } else {
-              changeMode( "browse" )
+          Image {
+            height: parent.height
+            width: parent.width / 2
+            anchors.right: parent.right
+            anchors.verticalCenter: parent.verticalCenter
+            source: Style.getThemeIcon( 'ic_create_white_24dp' )
+          }
+          Rectangle {
+            x: modeswitch.checked ? parent.width - width : 0
+            width: 36 * dp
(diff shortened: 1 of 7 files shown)


qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-431ae83d8eac21ab9c3fc26e174c55a2a87c3c90-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-431ae83d8eac21ab9c3fc26e174c55a2a87c3c90-x86.apk)

src/qml/qgismobileapp.qml
```diff
@@ -176,6 +180,27 @@ ApplicationWindow {
         visible: stateMachine.state === "digitize"
       }
 
+      /** A rubberband for ditizing **/
```
3nids:
```suggestion
      /** A rubberband for measuring **/
```

src/qml/qgismobileapp.qml
```diff
@@ -258,15 +283,15 @@ ApplicationWindow {
...
+      rubberbandModel: stateMachine.state === 'measure' ? measuringRubberband.model : digitizingRubberband.model
       project: qgisProject
       crs: qgisProject.crs
     }
 
     x: mainWindow.width / 2 + 24 * dp
     y: mainWindow.height / 2 + 24 * dp
 
-    text: qfieldSettings.numericalDigitizingInformation && stateMachine.state === "digitize" ?
+    text: ( qfieldSettings.numericalDigitizingInformation && stateMachine.state === "digitize" ) || stateMachine.state === 'measure' ?
```
3nids: do we really want to dispay the coords all the time?
we could maybe display it before we actually have a line (i.e. you are measure a point until you start measuring distance/area).

src/qml/qgismobileapp.qml
```diff
@@ -189,8 +214,8 @@ ApplicationWindow {
     CoordinateLocator {
       id: coordinateLocator
       anchors.fill: parent
-      visible: stateMachine.state === "digitize"
-      highlightColor: digitizingToolbar.isDigitizing ? digitizingRubberband.color : "#CFD8DC"
+      visible: stateMachine.state === "digitize" || stateMachine.state === 'measure'
+      highlightColor: digitizingToolbar.isDigitizing ? stateMachine.state === "digitize" ? digitizingRubberband.color : measuringRubberband.color : "#CFD8DC"
```
3nids: it might be clearer to have a variable like `currentRubberand: stateMachine.state === "digitize" ? digitizingRubberband: measuringRubberband` and use the `currentRubberand` everywhere.

src/qml/qgismobileapp.qml
```diff
@@ -475,17 +500,26 @@ ApplicationWindow {
 
     onVertexAdded: {
       coordinateLocator.flash()
-      digitizingRubberband.model.addVertex()
+      if( stateMachine.state === 'measure' )
+        measuringRubberband.model.addVertex()
```
3nids: here we would then avoid testing and simply have
`currentRubberand.model.addVertex()`

signedav: I agree.

src/qml/qgismobileapp.qml
signedav: Hm... or having the coords in case of the `numericDigitizingInformation` setting?

3nids: one or the other yes, it might still be interesting to display them without having to change the setting. but no strong opinion

signedav: Hm... I like your idea as well. I'll think about it...

signedav:
1. Integrated the visibility of coordinates, like @3nids proposed. I agree, that on digitizing the coordinates can be helpful on "finding the next vertex to digitize" but on measure, you know the point already. So it makes sense to have the coordinates in case there is only one (none) vertex, and after it disappears.
2. I took black (half-transparent) as color, since measuring is more a "neutral" thing compared to digitize where it's red
measurepleasure2

3. I have still unfinished discussions with @m-kuhn regarding the mode switch. We agree that having a 3-state-switch on top all the time is very space-consuming. I made this carousel implementation you see in the following GIF. Clicking goes to the next state. @m-kuhn thinks, that when the user misses the state he wants he has to begin again from start and that's annoying. He proposed a menu. For me menu are too many clicks. He proposed as well to have "measure" in a menu and keep the existing switch between "digitze" and "browse" - but because these are states where one excludes the others, I would like to have them all at the same place. 
**So I propose this compromise:** This carousel and on "long press" a space-consuming switch appears. This switch can be replaced with a menu. Probably a menu is better.
What do you think @m-kuhn ?
measurepleasure_buttons

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-58f3b3ace1291e42be950019505ac0a6e3d20f68-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-58f3b3ace1291e42be950019505ac0a6e3d20f68-x86.apk)

signedav:
@3nids had the idea to keep the switch button (like you proposed as well @m-kuhn) and have the measure key on the canvas below the GPS button. But only in "browsing" mode (because the user gets confused) so it looks like a "substate" of the "browsing" mode.

Though it appears then as a "substate" of browsing (though it's not) I like this idea as well.
What do you think @m-kuhn ?

src/qml/qgismobileapp.qml
m-kuhn: What's the downside of showing it?
Would it be better to move it to the end instead of the beginning?

m-kuhn:
Thanks a lot @signedav that looks very good already!

In my imagination it still would give the best screen-space/exposure/availability tradeoff to put it into the menu and use it as an "override" mode. In explicit, just block whatever else was going on, measure with only a "Cancel" button somewhere which will revert to the state before when clicked.

src/qml/qgismobileapp.qml
signedav: It's to make it slimmer. Do you think the coordinates are requested? 

Like I said I assume it's not really needed 

> the coordinates can be helpful on "finding the next vertex to digitize" but on measure, you know the point already

But I dont mind to put it back. Just tell me.

signedav:
So you suggest to keep the switch but kind of "disabling" it while being in the special measure mode?
What kind of cancel button do you think about? Something on the digitizing toolbar?

src/qml/qgismobileapp.qml
m-kuhn: It's not a request, but I still think it could be nice. Maybe make them in gray, to reduce their visual weight?

m-kuhn:
> So you suggest to keep the switch but kind of "disabling" it while being in the special measure mode?

Exactly

>  What kind of cancel button do you think about? Something on the digitizing toolbar?

Technically not on the digitizing toolbar (it's an unrelated toolbar) but at the same position.

signedav:
And def. in the menu?
No button below GPS buttons?

I just dont like to turn something off on a different place than turning on. But when you insist, I'll do it. 
I don't go on the barricades for that 😉

signedav:
> Technically not on the digitizing toolbar (it's an unrelated toolbar) but at the same position.

Yes. I meant just the GUI placement.

m-kuhn:
See e.g. how map geo admin handles it

Peek 2019-04-17 14-19

3nids:
From geoadmin, I like:
* circle at start
* close polygon to get area (I think people are measuring distance 90% of the time)

m-kuhn:
:D I was mostly referring to the state management ;)

src/qml/DashBoard.qml
```diff
@@ -107,12 +113,7 @@ Controls.Pane {
           onClicked: showMenu()
         }
 
-        Rectangle {
-          //empty space
-          height: 56 * dp
-          width: 56 * dp
-          color: mainColor
-        }
+        /*
```
m-kuhn: What's that?
Is that going to be used again or can it be deleted?

src/qml/qgismobileapp.qml
```diff
@@ -59,6 +59,8 @@ ApplicationWindow {
     Component.onCompleted: focusstack.addFocusTaker( this )
   }
 
+  property Rubberband currentRubberband
```
m-kuhn: Can you add a docstring explaining that this is used as a proxy to the digitizing or the measuring rubberband, I think this will help us remember in the future :)

signedav:
@m-kuhn and me agreed on the following solution:
measurepleasure4ever

Having the "Measure Tool" in the menu, deactivating identify / search / etc. and having the exit button for it, where the dashboard button is. For me this has been a good idea, because I wanted the user to find in any case the "exit", and if he tries to deactivate measure where he activated it, he will trip over it...

signedav:
I think in case there will come more "tools" like that it would make sense to have kind of a "toolbox" where all of em work in a similar way.

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-ebe7d2e6dd5320a134278ef96fc819c381ff90a4-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-ebe7d2e6dd5320a134278ef96fc819c381ff90a4-x86.apk)

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-471f647de5446513ccaa414d27aa09c9008f2033-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-471f647de5446513ccaa414d27aa09c9008f2033-x86.apk)

m-kuhn:
Nice!
What do you think about some text on the button that indicates the current state. Like "Exit measuring" or "Exit feature digitizing"?

signedav:
Thanks.

> What do you think about some text on the button that indicates the current state. Like "Exit measuring" or "Exit feature digitizing"?

On clicking "Measure Tool" in menu or on clicking the X to leave the "Measure Tool"? At the moment a toast message appears telling the user in what mode he went. Do you mean something else?

signedav:
Hi @m-kuhn 
I made it "almost" like you wished. Actually I couldn't make it have the transparent border around the circle. I'm sure it should work with something like `OpacityMask` - I tried and got funny graphical effects, but not what it should be. So I made this in the end:

image

Tell me, if I should investigate more in `OpacityMask` stuff.

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-b83ac8fe8a667ed3df14fc8b6db06f99c577686e-armv7.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-567-b83ac8fe8a667ed3df14fc8b6db06f99c577686e-x86.apk)

PR #570, opened 2019-04-23
Added some newlines at the end of file

cgaldini-datek22:
Please don't merge


m-kuhn:
Thank you very much for your contribution, but we need some code, not just newlines

PR #577, opened 2019-05-15
Add missing type role

jhonalex06:


mbernasocchi:
hi @jhonalex06 could you update this to current master?

PR #578, opened 2019-05-15
Allow hidding legend nodes

jhonalex06:
configuration is to be done InvisibleLayersAndGroups Qgis Plug-in

### DIFF
--- a/src/core/layertreemodel.cpp
+++ b/src/core/layertreemodel.cpp
@@ -183,3 +183,28 @@ void LayerTreeModel::updateCurrentMapTheme()
     }
   }
 }
+
+bool LayerTreeModel::filterAcceptsRow( int source_row, const QModelIndex &source_parent ) const
+{
+  QgsLayerTreeNode *node = mLayerTreeModel->index2node( source_parent );
+  QgsLayerTreeGroup *parentgroup = qobject_cast < QgsLayerTreeGroup *>( node );
+
+  if ( parentgroup )
+  {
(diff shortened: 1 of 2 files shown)


mbernasocchi:
hi @jhonalex06 could you update this to current master?

3nids:
Hi @jhonalex06 
Any update on this?

gacarrillor:
This PR has not been updated to master because with @m-kuhn we plan to support Hidden nodes in a different way. 

This PR works with the "Invisible Layer and Nodes" [plugin](https://github.com/gacarrillor/InvisibleLayersAndGroups). Using such plugin we configure the project, including visibility of layer and group nodes, in a way that this PR make QField understand them and hide them accordingly.

The proposed solution is to port the plugin to QGIS Core, as described in this [QEP](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/155).

In that sense, I'd ask @m-kuhn whether we should discard this PR or not.

m-kuhn:
@gacarrillor 
I would very much like to see this functionality land in QField.
How high is the QGIS core functionality currently on your priority list?

gacarrillor:
Do you mean that we can merge this PR for now, and when the QEP has been implemented in QGIS core we sort of make QField switch to that way? If so, I'm fine with that. 

For the QEP implementation I'd like to have a mentor :)

m-kuhn:
Yes, I'm fine with that too.
I think it should be a minimal code overhead to keep both checks in place and provide a smooth upgrade path.

m-kuhn:
Superseeded by #725

PR #416, opened 2018-12-14
produce beta app in PRs

3nids:

PR #582, opened 2019-05-23
Test Version Build

signedav:
I'm only testing a little.

It's about  to have another app built for testing. Then the brave testers do not need to uninstall the original app and can use it for comparison etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ services:
   - docker
 
 before_install:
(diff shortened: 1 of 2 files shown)


3nids:
See #416 I had the same idea in mind, had a quick try but failed

signedav:
Yeah, has been a quick shot into the air from me.
Should we continue on your branch with that @3nids ?

signedav:
Closing this - still can reopen when we continue on that...

signedav:
Oops... Sorry haven't seen @m-kuhn that you apparently did something here as well. Feel free to reopen.

PR #630, opened 2019-09-27
GPS and digitize toolbars improvements

nirvn:
Here's a before vs. PR shot first:
image

On the GPS toolbar front, biggest change here is the relocation of the toolbar to sit at the bottom right corner of the screen. This is favorable for a couple of reasons:
- It'll likely come as a more 'natural' position for most users since a large number of mapping apps places the button in that area;
- More importantly, the GPS toolbar has strong relation to digitizing, it therefore makes sense to have it close to the digitizing toolbar;
- In its former position, below the hamburger menu button, it felt odd to see it moving when the dashboard expanded while the hamburger button didn't;
- On narrow screens, with the dashboard expanded, the GPS button was going beyond the right edge of the screen.

The commit also changes the background color of the enable/disable GPS button to be semi-transparent, which allows for map content to still be visible when the GPS isn't active.

On the digitizing toolbar front, the buttons were re-ordered so that the vertex addition (+) and
removal (-) buttons sit next to each other. This felt like a much more natural location, and reduces the confusion between the vertex removal and the cancel buttons. 

To further help distinguishing the vertex removal vs the cancel buttons, the latter is now colored in red. It should help giving more visual guidance to new users, and just provide a better visual indicator overall.

I've also changed the green color to match the QField green, and killed the overly bright orange/yellow color. The idea here is to ultimately settle on a limited color palette.

Finally, the green color is now attached to save feature button instead of the vertex addition button, the latter still green for point layers but for polygon and line layers will share the same dark gray as its sister vertex removal (-) button.
for line and polygon layers

### DIFF
--- a/src/qml/DigitizingToolbar.qml
+++ b/src/qml/DigitizingToolbar.qml
@@ -16,23 +16,11 @@ VisibilityFadingRow {
   signal confirm
 
   Button {
-    id: removeVertexButton
-    iconSource: Style.getThemeIcon( "ic_remove_white_24dp" )
-    visible: rubberbandModel.vertexCount > 1
-    round: true
-    bgcolor: "#616161"
-
-    onClicked: {
-      vertexRemoved()
-    }
-  }
-
-  Button {
     id: cancelButton
     iconSource: Style.getThemeIcon( "ic_clear_white_24dp" )
     visible: rubberbandModel.vertexCount > 1
     round: true
-    bgcolor: "#616161"
+    bgcolor: "#900000"
 
     onClicked: {
       cancel()
@@ -61,7 +49,7 @@ VisibilityFadingRow {
       }
     }
     round: true
-    bgcolor: "#FFD600"
+    bgcolor: "#80cc28"
 
     onClicked: {
       // remove editing vertex for lines and polygons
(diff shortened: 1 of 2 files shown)


qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-630-8d39e7e8a91f3257c37e3b2d8afaa7db9bf3a373-armv7-X.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-630-8d39e7e8a91f3257c37e3b2d8afaa7db9bf3a373-x86-X.apk) (Android 5: [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-630-8d39e7e8a91f3257c37e3b2d8afaa7db9bf3a373-armv7-5.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-630-8d39e7e8a91f3257c37e3b2d8afaa7db9bf3a373-x86-5.apk))

mbernasocchi:
looks really nice when digitizing and the GPS  position feels natural, the only "strange" thing is that the GPS icon is "floating" (a lot of margin-bottom) when the digitizing row is missing
Screenshot_20190927-155213_QField Beta 630

Emic37:
Looks really nice. Maybe, the solution of the problem described by @mbernasocchi is to move the mode button in the bottom of the localisation button, as a material action button? 
What du you think about that?

nirvn:
Re GPS button not moving down when the digitize toolbar is hidden, it was a deliberate choice. The reasoning behind it is to avoid messing with muscle memory. Ie, the GPS toggle's distance to your thumb will always be X centimeters. I added the semi transparent background to avoid the button being too much of a visual obstacle there.

So that's my 2 cents on why it should be there :) if others have stronger feelings, it's easy enough to have it move down when the digitize toolbar is hidden.

m-kuhn:
> In navigation mode, this button can display an eye for example or the actual map icon. A long press (to avoid any mishandling) enables the edition mode.

The toggle edit button is indeed something that is not yet completely satisfying. It's hard to reach for something that's used often.
I haven't found a completely satisfying option so far. The problem with the long click for example is that it's hard to discover (I notice that many people don't find the long click options on the existing buttons).
And having a top level button is taking a lot of screen space for something that's normally only used once in a session.

Emic37:
@m-kuhn , I understand your point of view. I though too the solution of the top level button but, as you, I'm afraid that it taking too much of screen space.
So I really think that the action button is the less bad solution. And to avoid users desn't find the long press, short press on action button can command the display of a toast which explains how to enable the edition mode. 

Furthermore, there is the documentation and this other solution : 
Screenshot_20190928-125053

3nids:
I am not a huge fan a the floating button but you argument sounds very valid and I have nothing valuable to propose. 
Good job!

PR #642, opened 2019-09-29
Upgrade dashboard to a Drawer to enable edge swipe to open

nirvn:
One major UX issue I had when first testing QField was the absence of a edge swipe gesture to open the dashboard, which has come to be a pretty standard behavior these days.

This PR implements that by upgrading the dashboard to become a drawer. The result is pleasing to the thumb:
Peek 2019-09-29 13-45

As the above GIF indicates, this move makes the dashboard modal, whereas once opened, other parts of QField are blocked until the dashboard is closed. IMHO, this is actually a nice change as it insures that there are no overlap between several drawers (identify feature list drawer comes to mind). 

Would anyone object to this new modal behavior?

There *is* a way to make drawers non modal, but AFAIK, you also need to set those drawers as non interactive and that means losing the edge swipe (which brings us back to the UX gap this PR looks at addressing :smile:  )

### DIFF
--- a/src/qml/DashBoard.qml
+++ b/src/qml/DashBoard.qml
@@ -4,7 +4,7 @@ import QtQuick.Controls 2.4 as Controls
 import QtQuick.Layouts 1.1
 import "js/style.js" as Style
 
-Controls.Pane {
+Controls.Drawer {
   objectName: "dashBoard"
 
   signal showMenu
@@ -14,14 +14,17 @@ Controls.Pane {
   property MapSettings mapSettings
 
   property color mainColor: "#80CC28"
-  padding: 0
 
-  anchors { left: parent.left; bottom: parent.bottom; top: parent.top; }
+  width: Math.min( 300 * dp, mainWindow.width)
+  height: parent.height
+  edge: Qt.LeftEdge
+  dragMargin: 10 * dp
+  padding: 0
 
   property bool preventFromOpening: overlayFeatureFormDrawer.visible
   readonly property bool open: dashBoard.visible && !preventFromOpening
 
-  visible: false
+  position: 0
   focus: visible
   clip: true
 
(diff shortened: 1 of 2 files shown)


nirvn:
BTW, one option could be to have a QField settings to toggle the dashboard's modal state. By default, I think QField would be better served with the modal set to true (with edge swipe). Users could set the dashboard to non modal in their settings, which would deactivate the edge swipe too.

qfield-fairy:
Uploaded test apks for [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-642-6584a72f64a2ab37c533694ff2e305d20f7db059-armv7-X.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-642-6584a72f64a2ab37c533694ff2e305d20f7db059-x86-X.apk) (Android 5: [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-642-6584a72f64a2ab37c533694ff2e305d20f7db059-armv7-5.apk) and [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-642-6584a72f64a2ab37c533694ff2e305d20f7db059-x86-5.apk))

3nids:
It makes sense to me to make it modal and I would avoid adding an option for this (keep coherent behavior across devices, avoid cluttering the config UI)

Emic37:
Just tested. It's a really big improvement. Maybe, is it possible to reduce the necessary length of the swipe to close the panel? Because, a fast short swipe doesn't close the panel. This only move it.

nirvn:
@Emic37 , glad you like the improvement. The length it takes to swipe close in - AFAIK - hard coded into Qt itself. You can also close the drawer by clicking / tapping outside of the drawer area, that's faster than a swipe, and always closer to your resting thumb (that is, for left-handed folks :) )

Emic37:
@nirvn I didn't know this possibility. Indeed, it simply works by clicking outside the drawer. Amazing ! Now, I dream the new bottom drawer have the same feature to open it in full screen or to close it. 😃 Thanks for these improvement !

PR #703, opened 2019-11-07
fix new SDK

3nids:


3nids:
let's wait for the split tool merge instead

PR #713, opened 2019-11-13
Add quick compiler to cmake

m-kuhn:


m-kuhn:
@3nids we don't have the possibility to get a qgis master for linux with Qt 5.11 or newer, do we?
It's currently using the docker with Qt 5.9

PR #739, opened 2019-12-12
Msvc build

elpaso:
Builds on windows with QtCreator and MS C compiler (but crashes immediately)


src/qml/qgismobileapp.qml
```diff
@@ -1342,6 +1342,7 @@ ApplicationWindow {
       }
     }
     onDropped: {
+      console.warn('Drop URL: ' + drop.urls[0])
```
elpaso: This can be removed....

m-kuhn:
Nice, thanks!
Does this supersede #732?

elpaso:
Yes, it adds the MSVC build changes on top of MXE build changes.

m-kuhn:
Superseded by #793

PR #824, opened 2020-02-12
Update translations from stable branch

3nids:


3nids:
> I just wonder how we can be sure that everything which is pushed to master is (almost) immediately ported to stable when no further patch releases are planned.
I'm tempted to follow your proposal on having version branches, but that will come with the same transifex problems.

I don't think so but we will have to branch resources on Transifex too (or require a string freeze for patches).
Because, if we follow the branch approach, master will stay the source for Transifex.

m-kuhn:
Or we just go back to the previous approach and hold back features until we are done with patching? (using *merge after thaw*)

3nids:
> Or we just go back to the previous approach and hold back features until we are done with patching?

I'm not very happy with the previous approach. You can have roughly 2 months of bugfix releases. Holding merges just increases the conflicts and prevent people running unstable to test the last devs They would need to test PR per PR which is very unlikely.

m-kuhn:
Do you have experience with branching on transifex? What will be required to get that to work in our env?

(How would string freeze for patches help? Wouldn't it be required to have a string freeze on features?)

qfield-fairy:
Uploaded test apks for
 - [**armv7**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-824-98694a70dda2262da696048ae988c224f2f85635-armv7.apk)
 - [**x86**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-824-98694a70dda2262da696048ae988c224f2f85635-x86.apk)
 - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-824-98694a70dda2262da696048ae988c224f2f85635-arm64_v8a.apk)
 - [**x86_64**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-824-98694a70dda2262da696048ae988c224f2f85635-x86_64.apk)

3nids:
> Do you have experience with branching on transifex? What will be required to get that to work in our env?

I am not an expert, but I think we will need to make a copy of the resource (one per minor release). 
So in the deploy script, if we are in tag, we would try to create the resource 1.4 from 1.3 (and skip if it already exists). Not a big deal but still some coding time and complexity added to the scripts.

> (How would string freeze for patches help? Wouldn't it be required to have a string freeze on features?)

I was thinking that if we don't modify the string we could stick to the same resource, but we would need to tag the resources too. So no benefit from above solution.

m-kuhn:
> we would try to create the resource 1.4 from 1.3

Not 1.4 from master?

3nids:
> Not 1.4 from master?

hmmm, yes ;)

m-kuhn:
Shall we try that, @3nids ?

3nids:
ok, I will give it a try either today or tomorrow

PR #832, opened 2020-02-13
Fix color on valid constraints

signedav:
should be green

PR #807, opened 2020-02-08
Geotag indicator for external resource (i.e. picture) widget

nirvn:
This PR adds a visual indicator floating over external resource pictures to indicate the presence of geotags:
Screenshot from 2020-02-08 16-03-44

This is achieved by making use of QGIS' QgsExifTools class (thanks to Nyall for that one). It requires for https://github.com/qgis/QGIS/pull/34367 to be merged and the QField SDK to be updated. It also means QField would add QGIS' analysis library, which on the short term might be a bit of an overkill but on the longer term is likely going to benefit QField in other areas.

PR #833, opened 2020-02-13
fix crash on setting screen to keep alive

signedav:
There is a crash on all the newly created APKs like this https://github.com/opengisch/QField/pull/832#issuecomment-585747637

The APK created right after the keep_screen_on implementation worked fine. It's something with the current master (osgeo4a etc.)

What I did here:
- add `libexiv2.so` to build
- run `addFlag` for `FLAG_KEEP_SCREEN_ON` in `runOnAndroidThread`

Honestly I don't see through it clearly.

First I received this message on crash:
```
F .opengis.qfiel: java_vm_ext.cc:542] JNI DETECTED ERROR IN APPLICATION: the return type of CallObjectMethodV does not match void android.view.Window.addFlags(int)
F .opengis.qfiel: java_vm_ext.cc:542]     in call to CallObjectMethodV
F .opengis.qfiel: java_vm_ext.cc:542]     from void org.qtproject.qt5.android.QtNative.startQtApplication()
F .opengis.qfiel: java_vm_ext.cc:542] "qtMainLoopThread" prio=5 tid=14 Runnable
```

Then I replaced the `callObjectMethod` with `callMethod<void>`.

Then I received this message on crash:
```
F .opengis.qfiel: java_vm_ext.cc:542] JNI DETECTED ERROR IN APPLICATION: JNI CallStaticObjectMethodV called with pending exception android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
```

What leaded me to use the  `runOnAndroidThread`.

And finally it worked. But please have a look @nirvn because I'm not sure at all...

### DIFF
--- a/qfield.pri
+++ b/qfield.pri
@@ -22,6 +22,7 @@ android {
     $${OSGEO4A_STAGE_DIR}/$$ANDROID_TARGET_ARCH$$/lib/libssl_1_1.so \
     $${OSGEO4A_STAGE_DIR}/$$ANDROID_TARGET_ARCH$$/lib/libcrypto_1_1.so \
     $${OSGEO4A_STAGE_DIR}/$$ANDROID_TARGET_ARCH$$/lib/libexpat.so \
+    $${OSGEO4A_STAGE_DIR}/$$ANDROID_TARGET_ARCH$$/lib/libexiv2.so \
     $${OSGEO4A_STAGE_DIR}/$$ANDROID_TARGET_ARCH$$/lib/libgeos.so \
     $${OSGEO4A_STAGE_DIR}/$$ANDROID_TARGET_ARCH$$/lib/libgeos_c.so \
     $${OSGEO4A_STAGE_DIR}/$$ANDROID_TARGET_ARCH$$/lib/libgslcblas.so \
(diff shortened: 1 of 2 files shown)


signedav:
Or maybe @marioba has an idea :slightly_smiling_face:

nirvn:
@signedav , when did crash occur? At startup? It's odd, things worked fine on my two devices. I'll test the alternative approach tomorrow.

signedav:
@nirvn yes on startup. This APK https://github.com/opengisch/QField/pull/832#issuecomment-585747637 works for you?

nirvn:
@signedav , no you're right, that APK crashes, just tested.

But as you're pointing out, the APK from the prevent screen lock itself is just fine, tried again just now. Odd.

nirvn:
The APK from the geotag PR crashes: https://github.com/opengisch/QField/pull/807#issuecomment-585047477

So very likely it's only having to do with that alone.

nirvn:
Can you try to only commit the exif lib addition and see if the build is okay?

signedav:
Did it. But I'm pretty sure it wont work. Since on my computer it did not make it while I use the same ndk / osgeo4a and master...

PR #836, opened 2020-02-14
use date instead of commits count to determine version code

3nids:
remove all git log based calculations since they might not be available on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,6 @@ script:
   - ./scripts/upload-artifacts.sh
 
 jobs:
-  # allow_failures:
-  #   - name: "🍳 Testing"
   include:
     - stage: test
       name: "🍳 Testing"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -26,8 +26,6 @@ script:
   - ./scripts/upload-artifacts.sh
 
 jobs:
-  # allow_failures:
-  #   - name: "🍳 Testing"
```
3nids: not sure, but in the last PR only a stage is built, I guess there is an issue here
https://travis-ci.org/opengisch/QField/builds/650273752?utm_source=github_status&utm_medium=notification

scripts/ci/travis_build.sh
```diff
@@ -24,16 +20,18 @@ if [ -n "${TRAVIS_TAG}" ]; then
 
 elif [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
   ARCH_NUMBER=$(arch_to_build_number ${ARCH})
+  # DATA as YYYY DDD HH MM (without spaces)
```
m-kuhn:
```suggestion
  # DATE as YYYY DDD HH MM (without spaces)
```

PR #838, opened 2020-02-14
remove matrix that has been deprecated in Travis

3nids:


3nids:
this works as expected: see https://travis-ci.org/opengisch/QField/builds/650345098?utm_medium=notification&utm_source=github_status

PR #854, opened 2020-02-24
Extract Embedded Feature Form

signedav:
For reuse.

Excluded from Relation Editor.

Can be done as well in Relation Reference.

### DIFF
--- a/src/qml/EmbeddedFeatureForm.qml
+++ b/src/qml/EmbeddedFeatureForm.qml
@@ -1,0 +1,61 @@
+import QtQuick 2.11
+import QtQuick.Controls 2.4
+
+import QtQuick.Layouts 1.0
+import "."
+import Theme 1.0
+
+import org.qfield 1.0
+import org.qgis 1.0
+
+Popup {
+    id: formPopup
+
+    //the add entry stuff
+    property var attributeFormModel: AttributeFormModel {
+        featureModel: FeatureModel {
+        }
+    }
+
+    property var state
+
+    signal featureSaved
+    signal featureCancelled
+
+    onStateChanged: {
+        //because the binding in form does not work
+        form.state = formPopup.state
+    }
+
+    parent: ApplicationWindow.overlay
+
+    x: 24 * dp
+    y: 24 * dp
+    padding: 0
+    width: parent.width - 48 * dp
+    height: parent.height - 48 * dp
+    modal: true
+    closePolicy: Popup.CloseOnEscape
+
+    FeatureForm {
+        id: form
+        model: formPopup.attributeFormModel
+
+        focus: true
+
+        embedded: true
+        toolbarVisible: true
+
+        anchors.fill: parent
+
+        onSaved: {
+            formPopup.featureSaved()
+            formPopup.close()
+        }
+
+        onCancelled: {
+            formPopup.featureCancelled()
+            formPopup.close()
+        }
+    }
+}
--- a/src/qml/EmbeddedFeatureForm.qml
+++ b/src/qml/EmbeddedFeatureForm.qml
@@ -6,10 +6,15 @@ Popup {
     id: formPopup
 
     property alias state: form.state
-    //featureModel needs to be created to be accessible from outside
-    property AttributeFormModel attributeFormModel: AttributeFormModel {
-        featureModel: FeatureModel {
-        }
+
+    property alias currentLayer: formFeatureModel.currentLayer
+    property alias linkedRelation: formFeatureModel.linkedRelation
+    property alias linkedParentFeature: formFeatureModel.linkedParentFeature
+    property alias feature: formFeatureModel.feature
+
+    onAboutToShow: {
+        if( state === 'Add' )
+           formFeatureModel.resetAttributes()
     }
 
     signal featureSaved
@@ -27,7 +32,11 @@ Popup {
 
     FeatureForm {
         id: form
-        model: formPopup.attributeFormModel
+        model: AttributeFormModel {
+            featureModel: FeatureModel {
+                id: formFeatureModel
+            }
+        }
 
         focus: true
 
--- a/src/qml/EmbeddedFeatureForm.qml
+++ b/src/qml/EmbeddedFeatureForm.qml
@@ -1,31 +1,19 @@
 import QtQuick 2.11
 import QtQuick.Controls 2.4
-
-import QtQuick.Layouts 1.0
-import "."
-import Theme 1.0
-
 import org.qfield 1.0
-import org.qgis 1.0
 
 Popup {
     id: formPopup
 
-    //the add entry stuff
-    property var attributeFormModel: AttributeFormModel {
+    property alias state: form.state
+    //featureModel needs to be created to be accessible from outside
+    property AttributeFormModel attributeFormModel: AttributeFormModel {
         featureModel: FeatureModel {
         }
     }
 
-    property var state
-
     signal featureSaved
     signal featureCancelled
-
-    onStateChanged: {
-        //because the binding in form does not work
-        form.state = formPopup.state
-    }
 
     parent: ApplicationWindow.overlay
 
--- a/src/qml/editorwidgets/RelationEditor.qml
+++ b/src/qml/editorwidgets/RelationEditor.qml
@@ -91,12 +91,12 @@ Rectangle{
                   if( buffer() ) {
                       //this has to be checked after buffering because the primary could be a value that has been created on creating featurer (e.g. fid)
                       if( relationEditorModel.parentPrimariesAvailable ) {
-                          embeddedFeatureForm.state = 'Add'
-                          embeddedAttributeFormModel.featureModel.currentLayer = relationEditorModel.relation.referencingLayer
-                          embeddedAttributeFormModel.featureModel.linkedParentFeature = relationEditorModel.feature
-                          embeddedAttributeFormModel.featureModel.linkedRelation = relationEditorModel.relation
-                          embeddedAttributeFormModel.featureModel.resetAttributes()
-                          embeddedFeatureForm.active = true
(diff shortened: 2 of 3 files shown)


src/qml/EmbeddedFeatureForm.qml
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    property var state
+
+    signal featureSaved
+    signal featureCancelled
+
+    onStateChanged: {
+        //because the binding in form does not work
```
signedav: @3nids 
That's something I don't like. I didn't invest more time on it, so you can continue your work.
Tell me if you have an idea why the binding does not work here...

qfield-fairy:
Uploaded test apks for
 - [**armv7**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-69d6cf04bd85e03679cf6958deaa075ee4d0c6d3-armv7.apk)
 - [**x86**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-69d6cf04bd85e03679cf6958deaa075ee4d0c6d3-x86.apk)
 - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-69d6cf04bd85e03679cf6958deaa075ee4d0c6d3-arm64_v8a.apk)
 - [**x86_64**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-69d6cf04bd85e03679cf6958deaa075ee4d0c6d3-x86_64.apk)

src/qml/EmbeddedFeatureForm.qml
signedav: @3nids Regarding your question to bind it directly. 
```
embeddedPopup.form.state = !readOnly ? 'Edit' : 'ReadOnly'
```
This leads to an issue, since it cannot connect to the `form`.
It would be possible to connect to it, when having it like this:
```
property FeatureForm form : FeatureForm {
```
But it leaded to issues with the data. An empty form appeared :thinking:

src/qml/EmbeddedFeatureForm.qml
```diff
@@ -0,0 +1,61 @@
...
+import Theme 1.0
+
+import org.qfield 1.0
+import org.qgis 1.0
+
+Popup {
+    id: formPopup
+
+    //the add entry stuff
+    property var attributeFormModel: AttributeFormModel {
```
m-kuhn: Is this property

- expected to be set from outside (in this case, do we need the default value in here)
- or only for reading (in this case should the property be readonly)
- or not for outside usage (in this case should it be in an embedded QtObject{})

src/qml/EmbeddedFeatureForm.qml
m-kuhn: ```embeddedPopup.form.state = !readOnly ? 'Edit' : 'ReadOnly'```

You cannot access properties of embedded objects.
You can however "proxy" them, by making a toplevel `property bool readonly isReadOnly: form.state == readOnly` parameter which can be accessed from outside (readonly is optional).

src/qml/EmbeddedFeatureForm.qml
signedav: expected to be set from the outside, but we do not need to set properties of the AttributeFormModel but of it's FeatureModel. That's why the default value.

src/qml/EmbeddedFeatureForm.qml
signedav: Why not the `embeddedPopup.state` and bind it internally to the `embeddedPopup.form.state` ?

src/qml/EmbeddedFeatureForm.qml
m-kuhn: Then do the model internally and add an alias (proxy) property https://doc.qt.io/qt-5/qtqml-syntax-objectattributes.html

src/qml/EmbeddedFeatureForm.qml
m-kuhn: That should also work.
I also remembered the right tool to use again: `property alias`. https://doc.qt.io/qt-5/qtqml-syntax-objectattributes.html

signedav: This works:
```
id: formPopup
property alias superstate: form.state
FeatureForm {
        id: form
}
```
This does not work:
```
id: formPopup
property var superstate

FeatureForm {
        id: form
	state: formPopup.superstate
```
Suddenly the binding is broken. Why?

Somewhere in the FeatureForm it says:
```
state = 'Edit'
```

So the conclusion: Since an alias binds on both sides a binding does not break when we set the "depending" object (like state = 'Edit'). @m-kuhn, does that makes sence?

src/qml/EmbeddedFeatureForm.qml
```diff
@@ -0,0 +1,61 @@
+import QtQuick 2.11
+import QtQuick.Controls 2.4
+
+import QtQuick.Layouts 1.0
+import "."
```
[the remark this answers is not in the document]
signedav: no

src/qml/EmbeddedFeatureForm.qml
signedav: I don't know how to work with an alias here. 

When I do `property alias featureModel:form.model.featureModel` and connect it from outside, and set e.g. the `embeddedForm.featureModel.currentLayer` it says that it cannot set a property of null. Of course, FeatureModel is not created.

So I see the possibility to make an alias for every FeatureModel-Property (currentLayer, feature etc.) or to create the FeatureModel from outside and set it via `alias`. I don't like these solutions.

But maybe I don't see the obvious.

qfield-fairy:
Uploaded test apks for
 - [**armv7**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-da52dfe365e2fd55568c11706de80a464c71af34-armv7.apk)
 - [**x86**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-da52dfe365e2fd55568c11706de80a464c71af34-x86.apk)
 - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-da52dfe365e2fd55568c11706de80a464c71af34-arm64_v8a.apk)
 - [**x86_64**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-da52dfe365e2fd55568c11706de80a464c71af34-x86_64.apk)

src/qml/EmbeddedFeatureForm.qml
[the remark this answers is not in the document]
3nids: It took me a while to figure it out....but it seems it was causing a crash when creating the popup using `Qt.createComponent` + `createObject`. But  this is not occurring all the time.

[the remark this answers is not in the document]
signedav: Ok. Because of which import?
I removed layout, theme, . and qgis...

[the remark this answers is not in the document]
3nids: the one you remove, the "import ."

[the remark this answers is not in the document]
signedav: Okay, strange.

src/qml/EmbeddedFeatureForm.qml
3nids: makes sense to me

signedav: Thanks. It's kind of the illumination I was looking for. I guess I will use more often aliases in future...

m-kuhn: Yes. This is called "breaking the binding" and there is a tool since Qt 5.10 to detect it

https://www.kdab.com/new-qt-5-10-diagnostics-breaking-qml-bindings/

It would be really good if we checked this at some point.
> # Are broken bindings a problem?
> 
> 
> 
> In general, the above example is 100% correct QML code, following the documented QML semantics. Therefore, one cannot claim that such code should be rejected by the QML engine or unconditionally raise warnings.
> 
> However, in my experience at KDAB, *I have never encountered one single occasion when breaking a binding was intended by the developer*. Broken bindings were always caused by accident (during refactorings, not realizing that the original property was bound to something, etc.), or by excessive usage of JavaScript to manage state, instead of using more declarative approaches.

signedav: >  I have never encountered one single occasion when breaking a binding was intended by the developer.

Yes. But in this case we need to set the `state` from different places and we need the proxy property to pass it to the child object. Means the `alias` Property is exactly what we need.

Would be nice to have it in Qt Creator to show when a binding is broken. Because as you say, it's usually not intended.

m-kuhn: The quote is from the link above, not even emphasis is mine.

src/qml/EmbeddedFeatureForm.qml
signedav: I did it now with aliases of the featureModel properties.

qfield-fairy:
Uploaded test apks for
 - [**armv7**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-aac94391c34e89b18c81ff7f55ff99e272a4bf35-armv7.apk)
 - [**x86**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-aac94391c34e89b18c81ff7f55ff99e272a4bf35-x86.apk)
 - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-aac94391c34e89b18c81ff7f55ff99e272a4bf35-arm64_v8a.apk)
 - [**x86_64**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-854-aac94391c34e89b18c81ff7f55ff99e272a4bf35-x86_64.apk)

3nids:
good to go?

m-kuhn:
@3nids your call

PR #947, opened 2020-04-09
wip Bump sdk version to 5.14.2

m-kuhn:


nirvn:
@m-kuhn , @3nids , travis is gone, anyway we can have it back to check whether this SDK update is valid now?

PR #955, opened 2020-04-16
Fix QGIS_PREFIX_PATH to work for QGIS_INCLUDE_DIR; Add forgotten assert for APP_VERSION when building docker image

suricactus:


scripts/docker-build.sh
```diff
@@ -18,6 +18,11 @@ source ${DIR}/version_number.sh
 
 set -e
 
+if [[ -z $APP_VERSION ]]; then
+  echo "*** ERROR MISSING APP_VERSION ENVIRONMENT VARIABLE"
```
3nids, line 22: APP_VERSION is not mandatory to build, at least that is what is explained above.
Why do you want to enforce it?

suricactus: When I try to build without `APP_VERSION` or `APP_VERSION_CODE`, if fails with the error below. This happens in case https://github.com/opengisch/QField/blob/master/scripts/version_number.sh does not manage to set the variable. That's why I put the check after calling that script.
```
suricactus@orange:~/work/qgis/QField$ docker run -v $(pwd):/usr/src/qfield --network="host" opengisch/qfield-sdk:20200305 /usr/src/qfield/scripts/docker-build.sh --rm
[...]
FAILURE: Build failed with an exception.

* Where:
Build file '/usr/src/qfield/build-docker/out/build.gradle' line: 86

* What went wrong:
A problem occurred evaluating root project 'out'.
> For input string: ""

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 4s
Building the android package failed!
  -- For more information, run this command with --verbose.
```
The reason is that the version file remains empty:
```
suricactus@orange:~/work/qgis/QField$ head -n 3 version.pri 

VERSIONCODE = 
VERSTR = ''
```

but if I run it like velow, all is good.
```
suricactus@orange:~/work/qgis/QField$ docker run -v $(pwd):/usr/src/qfield --network="host" -e APP_VERSION=v1.2.3 opengisch/qfield-sdk:20200305 /usr/src/qfield/scripts/docker-build.sh --rm
```
The documentation is right, I can pass either `APP_VERSION` or `APP_VERSION_CODE`, so check the new commit.

m-kuhn:
Good catch!

PR #981, opened 2020-05-01
hyperlinks or deep link from QFIELD forms

oznerol2:
Hello,

need to set forms to use value as hyperlink or deep link ( ie https://waze.com/ul?param_name=value[&param_name=value])   from QFIELD forms.

(HMTL multilines can be use into QGIS 3 forms )

Any solution ?


m-kuhn:
@oznerol2 please open an issue at https://github.com/opengisch/QField/issues

PR #999, opened 2020-05-08
Add option to change attachments on desktop build

suricactus:
Clicking on the gallery button opens a standard qt file dialog, instead of silently ignoring it.

### DIFF
--- a/src/core/picturesource.cpp
+++ b/src/core/picturesource.cpp
@@ -15,10 +15,24 @@
  ***************************************************************************/
 #include "picturesource.h"
 
-PictureSource::PictureSource( QObject *parent )
+#include <QFileDialog>
+
+PictureSource::PictureSource( QObject *parent, const QString &prefix, const QString &pictureFilePath )
   : QObject( parent )
+  , mPrefix( prefix )
+  , mPictureFilePath( pictureFilePath )
 {
 
+}
+
+void PictureSource::init()
+{
+  QString path { QFileDialog::getOpenFileName( nullptr, tr( "Select Media File" ), mPrefix, tr( "JPEG images (*.jpg *.jpeg)" ) ) };
+
+  if ( path.startsWith( mPrefix ) )
+    path = path.remove( mPrefix );
+
+  emit pictureReceived( path );
 }
 
 PictureSource::~PictureSource()
--- a/src/core/picturesource.cpp
+++ b/src/core/picturesource.cpp
@@ -15,24 +15,19 @@
  ***************************************************************************/
 #include "picturesource.h"
 
-#include <QFileDialog>
+#include <QTimer>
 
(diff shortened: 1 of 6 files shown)


suricactus:
@m-kuhn let's merge this, we are waiting for the next opportunity in the numbers :)

m-kuhn:
I agree there should be a desktop implementation for this, but I'm not convinced that the picture source is the right place for it. This handling is specific for the "gallery picture", whereas a different implementation could be envisioned for camera.
I think this logic/dialog should instead be in the `getGalleryPicture` implementation and no requirement should be there to call `init()` from QML.

suricactus:
I did not find a way to return a `PictureSource` and in the mean time avoid calling `init`, because the rest of the implementation depends on the `pictureReceived` signal. If I emit it from the constructor, the listeners does not get triggered, as they are attached at later stage.

I do agree that `getGalleryPicture` is the place to go, I will try to find a solution

src/core/picturesource.h
```diff
@@ -35,6 +35,9 @@ class PictureSource : public QObject
      * Emit this signal when a picture really has been received.
      */
     void pictureReceived( const QString &path );
+  protected:
```
m-kuhn, line 38:
```suggestion
  private:
```

src/core/picturesource.cpp
```diff
@@ -15,10 +15,19 @@
...
+
+PictureSource::PictureSource( QObject *parent, const QString &prefix, const QString &pictureFilePath )
   : QObject( parent )
+  , mPrefix( prefix )
+  , mPictureFilePath( pictureFilePath )
 {
+  if ( mPictureFilePath.startsWith( mPrefix ) )
+    mPictureFilePath = mPictureFilePath.remove( mPrefix );
 
+  QTimer::singleShot(0, this, [ = ] () {
```
m-kuhn, line 28: Nice solution

src/core/picturesource.h
suricactus: done!

src/core/picturesource.h
```diff
@@ -26,7 +26,7 @@ class PictureSource : public QObject
 {
     Q_OBJECT
   public:
-    explicit PictureSource( QObject *parent = nullptr );
+    explicit PictureSource( QObject *parent = nullptr, const QString &prefix = QString(), const QString &pictureFilePath = QString() );
```
m-kuhn, line 29: Can you add a docstring that explains the behavior with respect to pictureFilePath and emitting the signal? With explicit recommendations for subclasses, something like "subclasses which implement their own file dialog should provide an empty string for \a pictureFilePath and emit \a pictureReceived when appropriate."

suricactus: Done!

PR #1014, opened 2020-05-11
Translate interface into Bulgarian

suricactus:


m-kuhn:
Thanks.

/note to self, we should fix pull requests from other repositories (and disable test build downloads) -- or switch to github actions.

PR #1018, opened 2020-05-13
Return bool indication if create/update/delete was successfull

suricactus:
There was no way to check whether the datasource operation was successful or not. More importantly, the UI and data source state divert at some point. For example, a failed delete causes the feature to disappear, but on project reload appears again. 

This PR adds indication something went wrong both for the user and in the log.


src/core/featuremodel.cpp
```diff
@@ -351,58 +351,147 @@ void FeatureModel::featureAdded( QgsFeatureId fid )
...
-
-  startEditing();
+    return false;
 
-  connect( mLayer, &QgsVectorLayer::featureAdded, this, &FeatureModel::featureAdded );
-  if ( !mLayer->addFeature( mFeature ) )
+  if ( ! startEditing() )
   {
-    QgsMessageLog::logMessage( tr( "Feature could not be added" ), "QField", Qgis::Critical );
+    QgsMessageLog::logMessage( tr( "Cannot start editing" ), "QField", Qgis::Warning );
```
signedav, line 361: Maybe you could improve these log messages by adding additional info like "Cannot start editing on layer <name> to create feature" etc. 
Just an input. Not a requested change.

src/core/featuremodel.cpp
```diff
@@ -351,58 +351,147 @@ void FeatureModel::featureAdded( QgsFeatureId fid )
...
+      if ( childLayer->startEditing() )
       {
-        childLayer->deleteFeature( childFeature.id() );
+        QgsFeatureIterator relatedFeaturesIt = referencingRelation.getRelatedFeatures( mFeature );
+        QgsFeature childFeature;
+        while ( relatedFeaturesIt.nextFeature( childFeature ) )
+        {
+          if ( ! childLayer->deleteFeature( childFeature.id() ) )
+          {
+            QgsMessageLog::logMessage( tr( "Cannot delete feature from child layer" ), "QField", Qgis::Critical );
```
signedav, line 428: As well here maybe the child layer name in the message.

src/qml/FeatureForm.qml
```diff
@@ -300,7 +300,9 @@ Page {
             {
               AttributeValue = isNull ? undefined : value
               if ( qfieldSettings.autoSave && !dontSave ) {
-                save()
+                if ( ! save() ) {
+                  displayToast( qsTr( 'Unable to save changes') )
```
signedav, line 304: I would prefer a log message than a display toast here.
Since it saves "in the background" it's very confusing for the user to get a toast.

src/qml/FeatureForm.qml
```diff
@@ -347,7 +349,10 @@ Page {
       return
     }
 
-    save()
+    if ( ! save() ) {
+      displayToast( qsTr( 'Unable to save changes') )
```
signedav, line 353: Here a display toast is fine.

src/qml/FeatureForm.qml
```diff
@@ -347,7 +349,10 @@ Page {
...
+      return
```
signedav, line 354: Shouldn't it go back to "Edit" state anyway? I'm unsure.

src/qml/FeatureForm.qml
```diff
@@ -361,24 +366,24 @@ Page {
...
+    return isSuccess
   }
 
   function cancel() {
-    if( form.state === 'Add' && featureCreated && !qfieldSettings.autoSave )
-    {
-      model.deleteFeature()
+    if( form.state === 'Add' && featureCreated && !qfieldSettings.autoSave ) {
+      if ( ! model.deleteFeature() ) {
+        displayToast( qsTr( "Failed to reset feature" ) )
```
signedav, line 385: Log message instead of toast.

src/qml/Tracking.qml
```diff
@@ -39,12 +39,16 @@ Item{
...
+                  if ( ! featureModel.create() ) {
+                      displayToast( qsTr( "Failed to create feature!" ) )
+                  }
                   mainModel.feature = featureModel.feature
               }
               else
               {
-                  featureModel.save()
+                  if ( ! featureModel.save() ) {
+                      displayToast( qsTr( "Failed to save feature!" ) )
```
signedav, line 50: Log message instead of toast. Otherwise this would pop up all the time during the tracking.

src/qml/qgismobileapp.qml
```diff
@@ -810,22 +810,28 @@ ApplicationWindow {
...
                 {
                     digitizingFeature.resetAttributes();
                     if( overlayFeatureFormDrawer.featureForm.model.constraintsHardValid ){
                       //when the constrainst are fulfilled
-                      digitizingFeature.create()
-                      overlayFeatureFormDrawer.featureForm.featureCreated = true
+                      overlayFeatureFormDrawer.featureForm.featureCreated = digitizingFeature.create()
+
+                      if ( ! overlayFeatureFormDrawer.featureForm.featureCreated ) {
+                        displayToast( qsTr( "Failed to autosave feature!" ) );
```
signedav, line 821: Generally I would display toasts if an action that is made by the user fails (like confirm feature edit etc), and just write log message on actions made in the background by auto save or tracking functionalities.

src/core/featuremodel.cpp
suricactus: I ws thinking about it, but I was not sure how detailed I should make the log messages. Thanks for the feedback.

src/core/featuremodel.cpp
suricactus: Done

src/qml/FeatureForm.qml
suricactus: I reply here. First, thanks for the detailed review whether to show a toast or not. I think I've addressed all the comments below too. From now on I will keep in mind that background operations should not pop up toasts. As the log is already written withing the C++ part, I just need to revert the original QML most of the time.

src/qml/FeatureForm.qml
suricactus: I put it back to keep the previous behavior.

src/qml/FeatureForm.qml
suricactus: done!

src/qml/qgismobileapp.qml
suricactus: Clear and noted for the future, thanks!

src/qml/qgismobileapp.qml
```diff
@@ -810,24 +810,25 @@ ApplicationWindow {
...
-                if( overlayFeatureFormDrawer.featureForm.featureCreated ) {
-                  //delete the feature when the geometry gets invalid again
-                  digitizingFeature.deleteFeature()
-                  overlayFeatureFormDrawer.featureForm.featureCreated = false
-                }
+              if( overlayFeatureFormDrawer.featureForm.featureCreated ) {
+                // delete the feature when the geometry gets invalid again
+                // indirect action, no need to check for success and display a toast, the log is enough
+                digitizingFeature.deleteFeature()
+                overlayFeatureFormDrawer.featureForm.featureCreated = false
```
signedav, line 830:
```suggestion
                overlayFeatureFormDrawer.featureForm.featureCreated = digitizingFeature.deleteFeature()
```
Not sure, but seems better to me.

suricactus: I think it should remain always `false`, as it is used a little bit further down to decide whether to call `digitizingFeature.create()` or `digitizingFeature.save()`. A deleteted feature should be saved, rather than created.
Check: 
https://github.com/opengisch/QField/pull/1018/files/c3038e3e592cd4b0005b19a7fd6b6c434515332b#diff-2692409a5f23502d8210c005af0e4787R859-R867

Even if I am wrong and it is used as sort of a success flag, we should change the name, because it is funny to assign on `featureCreated = deleteFeature()` :)

signedav: Ah sorry. Of course it should remain true, if it cannot be deleted. 
So  `featureCreated = !deleteFeature()` or ` if (deleteFeature) featureCreated = false`.

suricactus: done!

PR #1024, opened 2020-05-14
Fix feature count when deleting a feature

suricactus:
When deleting a feature from the feature list model, the number of features in the header remains the old one. For example, if we have selected the first item fron a list of three features, the header shows 1/3. If we delete the first one, the header remains showing 1/3, even though there are only two features left.

What is more, when we delete all features, there is a blank feature list form on the screen. In this PR, if there are no more features, the list is closed.

|before|after|
|--------|------|
| Peek 2020-05-14 18-03 | Peek 2020-05-14 17-59 |

### DIFF
--- a/src/core/multifeaturelistmodel.cpp
+++ b/src/core/multifeaturelistmodel.cpp
@@ -72,6 +72,10 @@ void MultiFeatureListModel::appendFeatures( const QList<IdentifyTool::IdentifyRe
 
 void MultiFeatureListModel::clear()
 {
+  // the model is already empty, no need to trigger "resetModel"
+  if ( mFeatures.count() == 0 )
+    return;
+
   beginResetModel();
   mFeatures.clear();
   endResetModel();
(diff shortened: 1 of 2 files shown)


src/core/multifeaturelistmodel.cpp
```diff
@@ -72,6 +72,10 @@ void MultiFeatureListModel::appendFeatures( const QList<IdentifyTool::IdentifyRe
 
 void MultiFeatureListModel::clear()
 {
+  // the model is already empty, no need to trigger "resetModel"
+  if ( mFeatures.count() == 0 )
```
m-kuhn, line 76:
```suggestion
  if ( mFeatures.isEmpty() )
```

suricactus: thanks!

suricactus:
@signedav thanks, it is better your way. I was so focused on the little change in `deleteFeature`/`countChanged` signal that I forgot to reuse the connection. Cheers!

Issue #1019, opened 2020-05-13
Hyperlinks


VxTedxV:
QGis 3 allow using hyperlinks in attributes form with the attachment widget ("use a hyperlink for document path (read-only)" option) and works well. In QField, the path isn't shown correctly and isn't hyperlink.

I saw pull request #981 but I think it's an issue (and @m-kuhn suggests @oznerol2 to open an issue and it seems not be done yet)

### Reproduction Steps:

In Qgis, configure a field containing an URL as an attachment (attachment widget) and check "display resource path", "use a hyperlink for document path (read-only)" and "show full path"

### Expected behavior:
Hyperlink is displayed and works as well as in QGis

### Observed behavior:

Hyperlink works well in QGis using form view
In QField hyperlink doesn't work and isn't displayed as an hyperlink (displayed as simple black text, not underlined, not blue) and only a part of URL is displayed (in my test, only the part after the last "/" of my URL is displayed)

**QField version:** 1.5.3 (010503991)

VxTedxV:
Labelled as "enhancement" but I think priority is the bug where the full path is not shown (only a part instead). Actually, if the hyperlink doesn't work as expected, user should still be able to see the full path to be able to copy/paste in their web browser.

PR #1029, opened 2020-05-15
Respect hyperlink attachments

suricactus:
Fixes https://github.com/opengisch/QField/issues/1019
image

NOTE: in QField only the filename is shown, unlike QGIS, where the whole path is visualized
https://github.com/opengisch/QField/pull/1029/files#diff-bacb7c98c9810cffdfedd42311433719R71

### DIFF
--- a/src/qml/editorwidgets/ExternalResource.qml
+++ b/src/qml/editorwidgets/ExternalResource.qml
@@ -59,7 +59,23 @@ Item {
     id: expressionEvaluator
     feature: currentFeature
     layer: currentLayer
-    expressionText: currentLayer ? currentLayer.customProperty('QFieldSync/photo_naming')!==undefined ? JSON.parse(currentLayer.customProperty('QFieldSync/photo_naming'))[field.name] : '' : ''
+    expressionText: {
+      if ( currentLayer && currentLayer.customProperty('QFieldSync/photo_naming') !== undefined ) {
+        return JSON.parse(currentLayer.customProperty('QFieldSync/photo_naming'))[field.name] 
+      } else {
+        return ''
+      }
+    }
+  }
+
+  function getPictureFilePath() {
+    var evaluatedFilepath = expressionEvaluator.evaluate()
+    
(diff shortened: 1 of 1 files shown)


src/qml/editorwidgets/ExternalResource.qml
```diff
@@ -20,7 +20,14 @@ Item {
   property ViewStatus __viewStatus
 
   //on all mimetypes image/... and on empty values it should appear as an image widget
-  property bool isImage: value === undefined || FileUtils.mimeTypeName( qgisProject.homePath + '/' + value ).startsWith("image/") || FileUtils.fileName( qgisProject.homePath + '/' + value ) === ''
+  property bool isImage: value === undefined
+    || (
+      ! config.UseLink
```
m-kuhn, line 25: This is the new part, correct?
Could you adjust line 22 to reflect this?

m-kuhn:
Thanks for the fix.
I think it would be even better if it would be changed to a code block with `if` statements etc. the readability of those endless `|| ... && ... () ...` blocks is incredibly hard to read (even if it's better now with the line breaks you added)

suricactus:
I've changed a bit more, but since I was there, I decided to remove the code repetition and the lengthy line.

src/qml/editorwidgets/ExternalResource.qml
suricactus: It is still valid, no need to change the comment.

src/qml/editorwidgets/ExternalResource.qml
```diff
@@ -20,7 +20,20 @@ Item {
   property ViewStatus __viewStatus
 
   //on all mimetypes image/... and on empty values it should appear as an image widget
```
signedav, line 22:
```suggestion
  //on all mimetypes "image/..." and on empty values it should appear as an image widget except when it's configured as a link
```
I think Matthias mentioned already. Just add this part, to have everything here.

PR #1042, opened 2020-06-02
QML warning hunt

suricactus:
Fix some of the QML warnings.

### DIFF
--- a/src/qml/FeatureForm.qml
+++ b/src/qml/FeatureForm.qml
@@ -95,7 +95,7 @@ Page {
         }
 
         Repeater {
-          model: form.model && form.model.hasTabs ? form.model : 0
+          model: form.model.hasTabs ? form.model : 0
 
           TabButton {
             id: tabButton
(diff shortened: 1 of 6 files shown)


src/qml/FeatureForm.qml
```diff
@@ -95,7 +95,7 @@ Page {
         }
 
         Repeater {
-          model: form.model && form.model.hasTabs ? form.model : 0
+          model: form.model.hasTabs ? form.model : 0
```
3nids, line 98: are you sure of this one?

suricactus: When I follow the QML code that leads here, it seems there is (or at least should be) no case where `form.model` is falsy and different from an `AtttributeFormModel` instance.

This line does not particullary cause any warning, but I updated it together with 193, so there are no pieces of code checking for `form.mode` existance and tricking the reader to believe it can be falsy.

m-kuhn:
@suricactus can you also resolve the merge conflict, so it's ready to be merged?

suricactus:
Thanks for the reminder, didn't saw there was a conflict. Done!

PR #1052, opened 2020-06-06
Cppcheck comparison

m-kuhn:


m-kuhn:
Just to confirm that 1.80 (18.04) and 1.70 (16.04) show errors that pass on 1.90 (20.04)

Issue #987, opened 2020-05-05
translations QField 1.6


VxTedxV:
Some strings are not translatable in Transifex :
- [x] value "not available" in global variables
- [x] "show on map canvas"
- [x] "center to current location"
- [x] "features" (attribute form title)

m-kuhn:
> value "not available" in global variables

This one is from QGIS, we'll need to either load qgis translations or to treat it differently

> "show on map canvas"

True, there was an issue open about that one, truly weird (it's tagged as translatable in the source and I think I even verified it's sent to transifex but does not pop up)

> "center to current location"

Shows up for me in the web frontend / transifex

> "features" (attribute form title)

Pull request pending https://github.com/opengisch/QField/pull/989

nirvn:
"Not available" fixed as part of PR #971.

VxTedxV:
"center to current location" -> my bad, I just checked. I made a mistake. It's better like that ;)

m-kuhn:
The last bit has been fixed in https://github.com/opengisch/QField/pull/1071, "Show on map canvas" shows up on transifex.

If there are more missing translation strings, please open a new issue. Thank you.

PR #1059, opened 2020-06-11
Warning hunt

suricactus:
Second iteration of warning hunt, at least those from the start screen are massacred. 

There are a couple more that I have investigated, but I am not advanced enough to find a stable solution. Next time :)

### DIFF
--- a/src/qml/CoordinateLocator.qml
+++ b/src/qml/CoordinateLocator.qml
@@ -29,7 +29,13 @@ Item {
 
   // some trickery here: the first part (!mapSettings.visibleExtent) is only there to get a signal when
   // the map canvas extent changes (user pans/zooms) and the calculation of the display position is retriggered
-  readonly property point displayPosition: !!mapSettings.visibleExtent || true ? mapSettings.coordinateToScreen(currentCoordinate) : 0
+  readonly property point displayPosition: {
+    // this property gets initially evaluated before the `currentCoordinate`
+    if ( ! currentCoordinate )
+      currentCoordinate = !!overrideLocation ? overrideLocation : snappingUtils.snappedCoordinate
+
+    return !!mapSettings.visibleExtent || true ? mapSettings.coordinateToScreen(currentCoordinate) : 0
+  }
 
   readonly property alias snappedCoordinate: snappingUtils.snappedCoordinate // In map coordinates, derived from snappinResult
   readonly property alias snappedPoint: snappingUtils.snappedPoint // In screen coordinates, derived from snappinResult
--- a/src/qml/LayerTreeItemProperties.qml
+++ b/src/qml/LayerTreeItemProperties.qml
@@ -45,6 +45,9 @@ LayerTreeItemProperties {
   }
 
   function isTrackingButtonVisible() {
+    if ( ! index )
+      return false
+
     return layerTree.data(index, FlatLayerTreeModel.Type) === 'layer'
         && layerTree.data(index, FlatLayerTreeModel.Trackable)
         && positionSource.active
--- a/src/qml/OverlayFeatureFormDrawer.qml
+++ b/src/qml/OverlayFeatureFormDrawer.qml
@@ -28,7 +28,6 @@ Drawer {
 
   interactive: overlayFeatureForm.model.constraintsHardValid || qfieldSettings.autoSave ? true : false
   dragMargin: 0
-  Keys.enabled: true
 
   /**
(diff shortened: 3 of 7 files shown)


src/qml/OverlayFeatureFormDrawer.qml
```diff
@@ -98,19 +97,8 @@ Drawer {
...
-    if (event.key === Qt.Key_Back || event.key === Qt.Key_Escape) {
-      if( overlayFeatureForm.model.constraintsHardValid || qfieldSettings.autoSave ) {
-        overlayFeatureFormDrawer.close()
-      } else {
-        //block closing to fix constraints or cancel with button
-        displayToast( qsTr( "Constraints not valid" ) )
-      }
-      event.accepted = true
-    }
-  }
```
3nids, line 115: is this ok to drop all of this?

src/qml/LayerTreeItemProperties.qml
```diff
@@ -45,6 +45,9 @@ LayerTreeItemProperties {
   }
 
   function isTrackingButtonVisible() {
+    if ( ! index )
+      return false
+
```
3nids, line 50: I think we could drop this whole file.
@m-kuhn @nirvn  ?

src/qml/OverlayFeatureFormDrawer.qml
suricactus: I think yes, it never gets triggered. The warning that targets this change says it all:

> Could not attach Keys property to: OverlayFeatureFormDrawer_QMLTYPE_58(0x55ef4b6bbaf0) is not an Item

which I interpret that the Drawer is not an Item, therefore no Keys property can be attached.

3nids: ping @signedav

src/qml/LayerTreeItemProperties.qml
m-kuhn: I don't know, why?

3nids: The layer tree is not used anymore, right?

src/qml/OverlayFeatureFormDrawer.qml
signedav: Since I don't have a computer atm I cannot check it. This implementation has been made to close the drawer with back key (escape on computer) and has to be implemented on every item that could have the focus. Possibly this is not needed here. If it doesn't accept it as property then it's not, i guess...

suricactus: @signedav it is already in `overlayFeatureForm`, I am not sure why it was added to the `Drawer` element, most probably by mistake/copy-paste. I just checked both the `escape` key on desktop and back button on android, and they seem to work after my change.

src/qml/LayerTreeItemProperties.qml
suricactus: This is the layer tree properties, that is shown when a layer tree node is long/double pressed, I don't think we should drop it? Nevertheless, I am not sure why there are two files with that name in qml and qml/ui thought, but both are actively used.

m-kuhn: If we can get rid of the .ui file, that'd be great. I can't explain why, but it seems to trigger the last remaining bit of https://github.com/opengisch/QField/issues/987 .

suricactus: Can we get rid of this file in a seperate PR, just to keep this one small and focused?

src/qml/CoordinateLocator.qml
```diff
@@ -29,7 +29,13 @@ Item {
 
   // some trickery here: the first part (!mapSettings.visibleExtent) is only there to get a signal when
   // the map canvas extent changes (user pans/zooms) and the calculation of the display position is retriggered
-  readonly property point displayPosition: !!mapSettings.visibleExtent || true ? mapSettings.coordinateToScreen(currentCoordinate) : 0
+  readonly property point displayPosition: {
+    // this property gets initially evaluated before the `currentCoordinate`
```
m-kuhn, line 33: That's surprising. I think the `currentCoordinate` should be acquired through a getter (transparently inside Qt). At this point it will be evaluated, so implicitly when it's required here.

I wonder if

1. my understanding of the process is wrong
2. currentCoordinate is not a bound property (and instead a property that is set/pushed / or depends on a property that is pushed)
3. there's a binding loop

Food for thoughts. This looks good anyway.

PR #1063, opened 2020-06-15
FEATURE disable geometry add/edit/delete

### LINKED ISSUE #1065: Lock features and geometry
It would be nice to be able to lock the geometries of a layer.

Use case:

A field worker should visit a number of known assets. For each visited asset, he should be able to adjust some attributes ("validated", "needs replacement", "repaired", "out of order", you name it).

The field worker should not be able to:

 - add new features
 - delete existing features
 - adjust the geometries

This behavior needs to be configurable on a per-layer basis in the project.

suricactus:
Add a new layer option to "lock geometries". This new option should ideally live on a new vector layer properties panel. QField should respect this new configuration option and not allow adding, splitting, modifying etc. geometries.

Peek 2020-06-15 11-32

https://github.com/opengisch/qfieldsync/pull/157 functionality:
Peek 2020-06-17 12-48

Closes #1065


VxTedxV:
I am used to doing that through project properties. Am I wrong ? Is the lock symbol useful ?

suricactus:
This PR adds a new alternative way to modify QFieldSync layer settings. Doing so via "Plugins -> QFieldSync -> Project configuration" is neither wrong nor going anywhere soon.

The lock icon is an indicator that the layer has it's geometries locked, otherwise the user would get confused why the layer is not editable.

VxTedxV:
Thanks for your answer.
 > The lock icon is an indicator that the layer has it's geometries locked, otherwise the user would get confused why the layer geometries are not editable.

In my opinion, if a layer is not editable, it's locked no matter a lock icon is there or not. Beware not to overwhelm QField interface which is used on tiny screens, in the field. I think it must keep visually simple.

> This PR adds a new alternative way to modify QFieldSync layer settings. Doing so via "Plugins -> QFieldSync -> Project configuration" is neither wrong nor going anywhere soon.

I mean in project properties outside of the plugin, in QGis, where we specify read only, searchable or identifiable layers. From that, why not going anywhere soon ?

3nids:
Are you aware of the "read-only" flag for vector layers?
Do we need a lock geometry vs a lock editing?
image

3nids:
So, there is apparently a need to lock geometries only.

m-kuhn:
Reasoning: sometimes it's necessary to send field workers out with geometries of existing infrastructure with the sole job of adjusting some attributes ("validated", "needs replacement", "repaired", "out of order", you name it). This flag was introduced for this. An issue should have been opened with the problem statement first. I have just done that https://github.com/opengisch/QField/issues/1065 .

This functionality is a more fine-grained control compared to "read only" which should continue to work as expected.

UX wise, I think the lock symbol makes sense for completely read only layers. I'm not 100% convinced about it for "geometry locked" layers (--> might trigger a user thinking: "why am I able to edit an attribute on a layer with a lock symbol?).
It might also be considered to hide it if the legend gets to narrow.

src/core/layertreemodel.h
```diff
@@ -39,7 +39,8 @@ class FlatLayerTreeModel : public QAbstractProxyModel
       Name,
       Visible,
       Trackable,
-      InTracking
+      InTracking,
+      IsGeometryLocked
```
m-kuhn, line 43: Can we rename this to "GeometryLocked" or "GeometryIsLocked"? (imperative form)
It currently is not in line with "InTracking" and others which also don't have an "Is" prefix.

suricactus: done!

suricactus:
@m-kuhn I think there should be some kind of indication in both cases of read-only and geometry-locked.  Maybe the lock symbol is not the best in the case of geometry locked, but would be nice to somehow get an explanation why one is not allowed to add/remove features. If you think indicator is too much, it might be shown only in the layer properties `geometry locked layer`?

Hiding the indicator if too narrow is a nice idea.

m-kuhn:
@nirvn you may have an opinion on that also.
@VxTedxV interested to hear your thoughts as well

VxTedxV:
Not trying to convince but this proposition reveals that this icon is not really useful since it becomes hidden...

suricactus:
Closing this as a duplicate of #1070 .

PR #1188, opened 2020-08-03
Rely on hardcoded WGS84 bounds

nirvn:
@m-kuhn , can you check if this fixes the goto issue on smartphone? Thanks


nirvn:
Closing, in favor of "doing the right thing", courtesy of brainstorming over hangout chat.

PR #1268, opened 2020-08-28
Replace deprecated Environment.getExternalStorageDirectory();

signedav:
`Environment.getExternalStorageDirectory()` is deprecated https://developer.android.com/reference/android/os/

I replaced it with `Context.getExternalFilesDirs(null)`. It solves it on opening local projects from e.g. `storage/emulated/0`.

But I'm not aware of the risks and the parts in `QFieldActivity.java`

### DIFF
--- a/android/src/ch/opengis/qfield/QFieldActivity.java
+++ b/android/src/ch/opengis/qfield/QFieldActivity.java
@@ -143,7 +143,7 @@ public class QFieldActivity extends Activity {
         }
 
         if (mExternalStorageAvailable) {
-            String storagePath = Environment.getExternalStorageDirectory().getAbsolutePath();
+            String storagePath = Context.getExternalFilesDirs(null).getAbsolutePath();
             mLocalizedDataPathsDir = storagePath + "/QField/basemaps/";
         }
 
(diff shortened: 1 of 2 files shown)


nirvn:
Did you check that datasets stored in a localized path still works? From my understanding of getExternalFilesDirs, it returns an app specific path, not the root of the external storage.

signedav:
Thanks @nirvn - Yes that's what I understood as well according this https://developer.android.com/training/data-storage#filesExternal 

But the local build succeeded. But I have to double check, maybe I did run the wrong build.

signedav:
Argh. Built from the wrong folder. So I guess you are right.
And this code fails anyway...

Hang on. I'll have a look again.

nirvn:
Right, I wouldn't expect a failure to launch the app here, just the loading of datasets from localized paths to fail. We advertise its location as storage root/QField/basemaps . 

The external files dir has two big issues: 1/ its path is user unfriendly when navigating using a file browser, and more importantly 2/ content is wiped when users remove the app. The latter could seriously piss off ppl copying large datasets onto their storage.

signedav:
@nirvn I completely agree. I did not intended to change the location. I just used the function wrongly.

PR #1311, opened 2020-09-18
Refresh QField logo

nirvn:
Logo harmonization campaign has begun:
Screenshot from 2020-09-18 09-07-13


signedav:
Does it need the white border? Looks strange to me, since there is no space between anymore.

nirvn:
@signedav , do you prefer that:
Screenshot from 2020-09-18 12-11-28

?

mbernasocchi:
definitely without border

nirvn:
Alright, PR updated, border no more.

mbernasocchi (approved):

PR #1312, opened 2020-09-18
Refresh logo on website, harmonize features area to match QFieldCloud website

nirvn:
Logo refresh campaign, part 3:
image

@mbernasocchi , while updating the logo, I've taken the liberty to adjust the features section to more or less match the style on the  qfield.cloud website.


mbernasocchi (approved):
the only micro changes that could be done is tho rename the pointx.svg to point_sync.svg and so on. Also, I think we could remove the _v3 in the css file and just call it qfield.css.

but up to you

nirvn:
@mbernasocchi , the _v3 suffix for the css file is to avoid browsers misrendering the page due to an old cached file. A good trick to remember.

I'll rename the files from point{1,2,3} to point_{deploy,formats,sync} now.

nirvn:
@mbernasocchi , done.

mbernasocchi:
> @mbernasocchi , the _v3 suffix for the css file is to avoid browsers misrendering the page due to an old cached file. A good trick to remember.
fair enough :) I usually add a query param with the date :)

nirvn:
@mbernasocchi , let's go with your query param then :) PR updated.

PR #1386, opened 2020-11-04
Add option to stop/start the gpkg flusher

suricactus:
Useful when manipulating the gpkgs of the currently opened project.

### DIFF
--- a/src/core/qgsgpkgflusher.cpp
+++ b/src/core/qgsgpkgflusher.cpp
@@ -181,6 +181,9 @@ void Flusher::flush( const QString &filename )
 
 void Flusher::stop( const QString &fileName )
 {
+  if ( ! mScheduledFlushes.contains( fileName ) )
+    return;
+
   mScheduledFlushes.value( fileName )->stop();
   mScheduledFlushes.remove( fileName );
 
--- a/src/core/qgsgpkgflusher.cpp
+++ b/src/core/qgsgpkgflusher.cpp
@@ -33,8 +33,13 @@ class Flusher : public QObject
 
     void flush( const QString &filename );
 
+    void stop();
+    void start();
+    bool isStopped() const;
+
   private:
     QMap<QString, QTimer *> mScheduledFlushes;
+    bool mIsStopped = false;
 };
 
 QgsGpkgFlusher::QgsGpkgFlusher( QgsProject *project )
@@ -90,8 +95,26 @@ void QgsGpkgFlusher::onLayersAdded( const QList<QgsMapLayer *> &layers )
   }
 }
 
+void QgsGpkgFlusher::stop()
+{
+  mFlusher->stop();
+}
+
+void QgsGpkgFlusher::start()
+{
+  mFlusher->start();
+}
+
+bool QgsGpkgFlusher::isStopped() const
+{
+  return mFlusher->isStopped();
+}
+
 void Flusher::scheduleFlush( const QString &filename )
 {
+  if ( mIsStopped )
+    return;
+
   if ( mScheduledFlushes.contains( filename ) )
   {
     mScheduledFlushes.value( filename )->start( 500 );
@@ -128,6 +151,30 @@ void Flusher::flush( const QString &filename )
     QgsMessageLog::logMessage( QObject::tr( "Could not flush database %1 (%3) " ).arg( filename, error ) );
     mScheduledFlushes[filename]->start( 500 );
   }
+}
+
+void Flusher::stop()
+{
+  mIsStopped = true;
+
+  const QStringList scheduledFlushesFileNames = mScheduledFlushes.keys();
+  for ( const QString &fileName : scheduledFlushesFileNames )
+  {
+    mScheduledFlushes.value( fileName )->stop();
+    mScheduledFlushes.remove( fileName );
+
+    flush( fileName );
+  }
+}
+
+void Flusher::start()
+{
+  mIsStopped = false;
+}
+
+bool Flusher::isStopped() const
+{
+  return mIsStopped;
 }
 
 #include "qgsgpkgflusher.moc"
--- a/src/core/qgsgpkgflusher.cpp
+++ b/src/core/qgsgpkgflusher.cpp
@@ -28,8 +28,6 @@ class Flusher : public QObject
 {
     Q_OBJECT
 
-  public:
-    static QMutex sMutex;
   public slots:
     void scheduleFlush( const QString &filename );
 
@@ -40,6 +38,7 @@ class Flusher : public QObject
     bool isStopped() const;
 
   private:
+    QMutex mMutex;
     QMap<QString, QTimer *> mScheduledFlushes;
     bool mIsStopped = false;
 };
@@ -112,8 +111,6 @@ bool QgsGpkgFlusher::isStopped() const
   return mFlusher->isStopped();
 }
 
-QMutex Flusher::sMutex = QMutex();
-
 void Flusher::scheduleFlush( const QString &filename )
 {
   if ( mIsStopped )
@@ -135,7 +132,7 @@ void Flusher::scheduleFlush( const QString &filename )
 
 void Flusher::flush( const QString &filename )
 {
-  QMutexLocker locker( &sMutex );
+  QMutexLocker locker( &mMutex );
 
   sqlite3_database_unique_ptr db;
   int status = db.open_v2( filename, SQLITE_OPEN_READWRITE, nullptr );
--- a/src/core/qgsgpkgflusher.cpp
+++ b/src/core/qgsgpkgflusher.cpp
@@ -41,24 +41,24 @@ class Flusher : public QObject
     void flush( const QString &filename );
 
     /**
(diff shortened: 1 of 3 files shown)


qfield-fairy:
Uploaded test apks for
 - [**armv7**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-bca0d5a68f9984255289b7156ab38f16cdaa57a4-armv7.apk)
 - [**x86**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-bca0d5a68f9984255289b7156ab38f16cdaa57a4-x86.apk)
 - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-bca0d5a68f9984255289b7156ab38f16cdaa57a4-arm64_v8a.apk)
 - [**x86_64**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-bca0d5a68f9984255289b7156ab38f16cdaa57a4-x86_64.apk)

src/core/qgsgpkgflusher.h
```diff
@@ -39,7 +40,23 @@ class QgsGpkgFlusher : public QObject
...
 
+    /**
+     * Immediately flushes all currently scheduled tasks and returns. After stopping the flusher, requesting a new flush will be ignored.
+     */
+    void stop();
+
+    /**
+     * Reenables requests for regular flushing.
+     */
+    void start();
```
m-kuhn, line 51: To use it from QML (which I guess is the aim since it's registered) this will either need to be `Q_INVOKABLE` or a `Q_PROPERTY`. I prefer the second one because it's easier to bind to it.

suricactus: I want to pass the instance to a QML component defined in c++ (qfcProjectModel), but since it is constructed in QML, I find this the easiest way to do it. I will not call start/stop directly from QML for now.

m-kuhn: Makes sense :+1:

src/core/qgsgpkgflusher.cpp
```diff
@@ -90,8 +96,26 @@ void QgsGpkgFlusher::onLayersAdded( const QList<QgsMapLayer *> &layers )
   }
 }
 
+void QgsGpkgFlusher::stop()
+{
+  mFlusher->stop();
```
m-kuhn, line 101: This needs to be protected with a mutex and the same mutex needs to be used in flush().
Since `flush()` is run in a thread, stop can be called when a flush is already ongoing. But we want to be sure there is no flushing going on when `stop()` returns.

suricactus: you are very right, will fix this

src/core/qgsgpkgflusher.cpp
```diff
@@ -28,13 +28,20 @@ class Flusher : public QObject
 {
     Q_OBJECT
 
+  public:
+    static QMutex sMutex;
```
m-kuhn, line 32: This could be private non static I think

src/core/qgsgpkgflusher.cpp
```diff
@@ -28,13 +28,20 @@ class Flusher : public QObject
...
   public slots:
     void scheduleFlush( const QString &filename );
 
     void flush( const QString &filename );
 
+    void stop();
```
m-kuhn, line 38: Docs

src/core/qgsgpkgflusher.cpp
suricactus: Yes, it can. I made it public static before I realized there is no need for lock from outside (`QgsGpkgFlusher::stop`).

src/core/qgsgpkgflusher.cpp
suricactus: done

m-kuhn (commented):
We will still have a problem with the current mutex implementation I think.

`mIsStopped` is checked in `scheduleFlush()` which launches `flush()` with a delayed timer.
So if a timer is already running, it will call flush no mather if `stop()` was called in between. What you want to protect is the file access itself which happens inside `flush()`.

Thererfore I think you need to protect the execution of the `flush()` itself with the mutex and also put the `mIsStopped` check in there.

Something like

```
flush();
QMutexLocker locker ( &mMutex );
mStopped=true;
```

src/core/qgsgpkgflusher.cpp
```diff
@@ -90,8 +114,26 @@ void QgsGpkgFlusher::onLayersAdded( const QList<QgsMapLayer *> &layers )
...
+}
+
+bool QgsGpkgFlusher::isStopped() const
+{
+  return mFlusher->isStopped();
+}
+
 void Flusher::scheduleFlush( const QString &filename )
 {
+  if ( mIsStopped )
```
m-kuhn, line 134: We still want to be able to schedule flushes. We just don't want them to be executed.

suricactus: What is the point to schedule a flush, if it is never going to get executed? I think it is safer to keep this `if ( mIsStopped ) return` in `Flusher::scheduleFlush`.  What will happen if you call `Flusher::stop`, then cann `Flusher::scheduleFlush`, the timer starts to count down, and in the mean time you call `Flusher::start`? It will execute the flushes that were scheduled while the `mIsStopped` was true, which I think is a bit uncertain and hard to explain.

Calling `Flusher::stop` already clears all the running timers, because they are immediately executed.

Probably I misunderstood your comment above. Anyway, I made another commit that combines your suggestions, while keeping the `scheduleFlush` early return.

qfield-fairy:
Ta-daaa, freshly created APKs are available:
  - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4e154fee1f557e3ba0c635b656a0ddbcbc02c8fe-arm64_v8a.apk)

Other architectures:

  - [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4e154fee1f557e3ba0c635b656a0ddbcbc02c8fe-armv7.apk)
  - [x86_64](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4e154fee1f557e3ba0c635b656a0ddbcbc02c8fe-x86_64.apk)
  - [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4e154fee1f557e3ba0c635b656a0ddbcbc02c8fe-x86.apk)

src/core/qgsgpkgflusher.cpp
m-kuhn: You are right.

My concern is the situation where you stop for a file A and while it's stopped you want to flush B and loose the event. You'd probably rather want to just defer that to when it's started again.

suricactus: For my usecase this would not be the issue, since the project will be unloaded just after the flush. However, I do agree for the general use it might cause troubles and would be safer to keep the flush for later.
What if we stop the timer by filename `Flusher::stop( const QString &fileName )`? This way we can still have the B flushed and A ignoring flushing requests.

m-kuhn: That would be very clean indeed.
And I also agree that it's more an aesthetic issue at the moment - and coule also be addressed with proper doc strings at the current point in time.

suricactus: I already gave it a try, this should do the trick.

qfield-fairy:
Ta-daaa, freshly created APKs are available:
  - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4d5aef46b1d4deeebadaf7e0411e1c1a634c4f5c-arm64_v8a.apk)

Other architectures:

  - [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4d5aef46b1d4deeebadaf7e0411e1c1a634c4f5c-armv7.apk)
  - [x86_64](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4d5aef46b1d4deeebadaf7e0411e1c1a634c4f5c-x86_64.apk)
  - [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4d5aef46b1d4deeebadaf7e0411e1c1a634c4f5c-x86.apk)

qfield-fairy:
Ta-daaa, freshly created APKs are available:
  - [**arm64_v8a**](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4921b465d922e02622ce58f1125f286940d757a5-arm64_v8a.apk)

Other architectures:

  - [armv7](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4921b465d922e02622ce58f1125f286940d757a5-armv7.apk)
  - [x86_64](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4921b465d922e02622ce58f1125f286940d757a5-x86_64.apk)
  - [x86](https://download.opengis.ch/qfield/ci-builds/qfield-dev-1386-4921b465d922e02622ce58f1125f286940d757a5-x86.apk)

m-kuhn (approved):

PR #1412, opened 2020-11-11
test

m-kuhn:


suricactus:
I think `github.actor` can do the trick too.

m-kuhn:
@suricactus when checking for `if secrets.TX_TOKEN` we had `Unrecognized named-value: 'secrets'.`. Will the same happen when it's run on a commit (and not a PR)?

suricactus:
It was wrong, `secrets.*` would not work in the `if` expression. The secrets should be assigned to environment variables and used as `if: env.TX_TOKEN`.

m-kuhn:
Does this actually work? Internet research didn't give a conclusive answer.

m-kuhn:
@suricactus works now, should we merge as is?

suricactus:
If the env is per job, I think it works fine.

Anyway I prefer your approack, it does not rely on some kind of gh magic and workarounds.

suricactus (approved):

PR #1467, opened 2020-12-08
Eliminate the next portion of warnings

suricactus:
Well, there are zero warnings when using QField.

### DIFF
--- a/src/qml/FeatureForm.qml
+++ b/src/qml/FeatureForm.qml
@@ -1,5 +1,6 @@
 import QtQuick 2.12
 import QtQuick.Controls 2.12
+import QtQuick.Controls.Material 2.12
 import QtQuick.Layouts 1.12
 import QtGraphicalEffects 1.0
 import QtQml.Models 2.11
@@ -590,6 +591,8 @@ Page {
     background: Rectangle {
       color: !model.constraintsHardValid ?  Theme.errorColor : !model.constraintsSoftValid ? Theme.warningColor : Theme.mainColor
     }
+
+    Material.foreground: undefined
 
     RowLayout {
       anchors.fill: parent
--- a/src/qml/Changelog.qml
+++ b/src/qml/Changelog.qml
@@ -89,7 +89,7 @@ Popup {
       }
 
       BusyIndicator {
-        anchors.horizontalCenter: parent.horizontalCenter
+        Layout.alignment: parent.Layout.alignment
 
         visible: changelogContents.status == ChangelogContents.LoadingStatus
         running: visible
--- a/src/qml/FeatureListForm.qml
+++ b/src/qml/FeatureListForm.qml
@@ -36,7 +36,7 @@ Rectangle {
   property bool allowEdit
   property bool allowDelete
   property bool fullScreenView: qfieldSettings.fullScreenIdentifyView
-  property bool isVertical: null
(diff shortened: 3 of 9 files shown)


nirvn:
Nice! Could we add a github PR test that would fail if any warning is found while compiling / launching QField? We don't want to slowly drift back to dozens of warnings we have to ignore.

suricactus:
I don't really see a way, since most of the warnings are happening runtime in QML. I am not sure, but I think @m-kuhn have experimented with static code analyzer a few months ago. Just every few months someone needs to use "Advanced Bee Project" and watch for warnings.

qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 881b6c2b19b8089ddc820e6e050a488cb004af9b:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1467-30f62e8d746f77b4bf73764bd72017ca4fc58303-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1467-30f62e8d746f77b4bf73764bd72017ca4fc58303-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1467-30f62e8d746f77b4bf73764bd72017ca4fc58303-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1467-30f62e8d746f77b4bf73764bd72017ca4fc58303-x86.apk)

m-kuhn:
Nice !
What we could do is to connect to QQmlEngine::warnings when running tests and starting a test app.

suricactus:
Would not really help for now, a lot of the errors are caused runtime, but all we have are unit tests. Some warnings happened in very specific conditions after moving a slider within a relation reference, after closing a popup etc.
 
Have you tried this project? https://github.com/faaxm/spix 
We can write a typical user workflow and then listening for `QQmlEngine::warnings`?

nirvn:
@suricactus , I still think there's value in _at least_ insuring that there are no warning on startup. Until very recently, we had a bunch.

m-kuhn:
- just running the app and checking for warning is a rather simple smoke test
- we should add more unit tests for individual qml components, they can check that they behave correctly and they can check for warnings
- spix looks interesting, let's have a look at that early next year

src/qml/FeatureForm.qml
```diff
@@ -591,6 +592,8 @@ Page {
       color: !model.constraintsHardValid ?  Theme.errorColor : !model.constraintsSoftValid ? Theme.warningColor : Theme.mainColor
     }
 
+    Material.foreground: undefined
```
m-kuhn, line 595: What is this for exactly?

suricactus: The `ToolBox` enters a binding loop implicitly: 

```QML PageHeader (parent or ancestor of Material): Binding loop detected for property "foreground"```

suricactus: So I should set the `Material.foreground` either to `undefined` or color of our choise (while). I think undefined is the safer option.

m-kuhn (approved):

PR #1617, opened 2021-01-24
Update SDK for QField 1.8

nirvn:


nirvn:
@m-kuhn , well, well, well -- I was sure 1.8 was using a 3.16 LTR build, but turns out we were using QGIS (3.17) master already on 1.8 -- makes the whole backporting a lot easier, which can just use the master SDK update.

PR #1735, opened 2021-03-02
Fix clang warning

nirvn:
The missing `#include <qgslabelingresults.h>` line led my compiler to complaint about a risky delete (https://github.com/opengisch/QField/blob/fc7e61407eed3feaa34c40c85bb040cb94ff8dbc/src/core/qgsquick/qgsquickmapcanvasmap.cpp#L173) of undefined class.


m-kuhn (approved):
If you convince mr workflow, that's good for me

PR #1775, opened 2021-03-15
Restore the previous continuous integration configuration

suricactus:
Whoops, there were few specific things that are no longer needed.


suricactus:
The failure is due that gh actions is using the current master to run, and the previous PR was green for the same reason.

nirvn (approved):

PR #1888, opened 2021-05-08
Fix ListView going broken when resetting feature form

### LINKED ISSUE #1884: QField allow to add only a few new features in clayer with constraints
QField does not allow to add more than 3-4 features to working layer with constraints: i have to end the editing mode, change to different layer, change back to  working layer and then start of editing: it allows me again to add 1-3 features, then i have to end editing/layer switching/start editing to be able to add 1-3 features, and so on.

Constraints are not simple "not null", but a bit more elaborated (if this attribute is this, then constraint is active;;; if another attributte is that then you dont have to enter values, etc..) I can upload QField project (this is my first github post, so im not sure if i have to put it here right now).

IMPORTANT: The layer does not act like that if there are no constraints, i cant identify which one is acting weird (there is about 10 constraints and default value calculations)...

Tested with geopackages and shapefiles.

**QField version:** 1.9.5 (5f66b7 v1.9.5)

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: [Dont know (first use of layer with coinstraints)]

nirvn:
This PR fixes #1884 . The long story short here is that when the feature form contained items with conditional visibility and items' visibility would change between the time you closed a feature form and re-opened it, the ListView would go broken. Reason for it going :boom: is our use _against documentation warning_ of ListView.contentY = 0. That led to worlds colliding, Thanos snapping our form out of existence, etc.etc.

### DIFF
--- a/src/qml/FeatureForm.qml
+++ b/src/qml/FeatureForm.qml
@@ -214,7 +214,7 @@ Page {
                 target: master
 
                 function onReset() {
-                  content.contentY = 0
+                    content.positionViewAtBeginning();
                 }
               }
 


m-kuhn (approved):
Awesome

m-kuhn:
On master a massive update on indentation and spacing was done, backporting is considerably harder at the moment

PR #1891, opened 2021-05-08
Focus improvements all around

### LINKED ISSUE #1883: Feature list looses focus
When opening the "feature list" from a layer menu and then selecting a feature by long pressing, the back button no longer closes the panel and instead wants to close the app.

Two things to do:

 - Fix the focus (assumption it is the focus)
 - Add a close button to the panel

nirvn:
This fixes #1883 as well as a bunch of other focus-related issues with tool buttons (e.g. zoom in / out) stealing focus and breaking our back / escape hardware key handling.

The PR also adds a permanent 'back' button to the feature list / form panel, which should please QField users on iOS.

### DIFF
--- a/src/qml/FeatureListForm.qml
+++ b/src/qml/FeatureListForm.qml
@@ -352,6 +352,15 @@ Rectangle {
       }
     }
 
+    onBackClicked: {
+        featureForm.focus = true;
+        if ( featureForm.state != "FeatureList" ) {
+            featureForm.state = "FeatureList";
+        } else {
+            featureForm.state = "Hidden";
+        }
+    }
+
     onStatusIndicatorClicked: {
       featureForm.state = "FeatureList"
     }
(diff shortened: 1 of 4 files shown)


m-kuhn:
... and windows.
Manual testing needed.

signedav:
Maybe it's not scope of this pr, but could a back key be included in the fast editing mode attribute form?

nirvn:
@signedav , yeah, I'm planning to address UI/UX shortcomings when fast editing is on in a separate PR (urgently needed for iOS, windows, etc.)

nirvn:
@m-kuhn , it's good you can give it some manual testing, I'm always a bit weary of touching the delicate balance (or is it? ;) ) of focuses.

m-kuhn (approved):
Feels great.
One inconaistency remains, clicking the "x" in the feature list clears the selection ( repeat long press), back button not. Clearing feels more intuitive to me. Might be a similar place to solve.

PR #1943, opened 2021-05-30
One more legend corner scenario: insertion of rows with pre-existing children

nirvn:
So, turns out an abstract item model can insert row items with pre-existing children. In our legend context, it can happen when a layer has a multi-level rule-based renderer with feature count on. The source layer tree model deletes and re-inserts the root items of the rule-based renderer, with children attached to those parent items.

Practical example: the advanced bees demo's apiary layer.

I didn't know this was possible, I had (wrongly) assumed that inserted items would always be called individually. By not having handled the above scenario, we'd end up losing legend items (e.g., with the advanced bees demo, if you add a point, the individual disease marker legend items disappear).

The solution is to trigger the more costly model re-building when we detect this scenario (we do something similar for row deletion).

For the record, we do _not_ trigger a model re-building on every row being inserted whenever we can avoid it to avoid UI freeze on super large source layer tree models.

### DIFF
--- a/src/core/layertreemodel.cpp
+++ b/src/core/layertreemodel.cpp
@@ -136,6 +136,24 @@ void FlatLayerTreeModelBase::insertInMap( const QModelIndex &parent, int first, 
   if ( mFrozen )
     return;
 
+  bool resetNeeded = false;
+  for ( int i = 0; first + i <= last; i++ )
+  {
+    QModelIndex index = mLayerTreeModel->index( first + i, 0, parent );
+    if ( mLayerTreeModel->hasChildren( index ) )
+    {
+      resetNeeded = true;
+      break;
+    }
+  }
+
+  if ( resetNeeded )
+  {
+    // Added rows with pre-existing children can't be handled, model reset needed
+    buildMap( mLayerTreeModel );
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 8fb0ef6451f23bec000f0d9889bd277e0df54f6c:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1943-e58c411308a5bcc9134af162e344403182d95ec4-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1943-e58c411308a5bcc9134af162e344403182d95ec4-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1943-e58c411308a5bcc9134af162e344403182d95ec4-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1943-e58c411308a5bcc9134af162e344403182d95ec4-x86.apk)

m-kuhn (approved):

PR #1960, opened 2021-06-04
Disable native camera on non android devices

m-kuhn:
It doesn't make sense to show a setting without effect

### DIFF
--- a/src/qml/QFieldSettings.qml
+++ b/src/qml/QFieldSettings.qml
@@ -27,7 +27,7 @@ Page {
     property bool fullScreenIdentifyView: false
     property bool locatorKeepScale: false
     property bool numericalDigitizingInformation: false
-    property bool nativeCamera: true
+    property bool nativeCamera: platformUtilities.supportsNativeCamera
     property bool autoSave: false
     property bool mouseAsTouchScreen: false
   }
@@ -56,6 +56,7 @@ Page {
           title: qsTr( "Use native camera" )
           description: qsTr( "If disabled, QField will use a minimalist internal camera instead of the camera app on the device.<br>Tip: Enable this option and install the open camera app to create geo tagged photos." )
           settingAlias: "nativeCamera"
+          isVisible: true
       }
       ListElement {
           title: qsTr( "Fast editing mode" )
@@ -66,6 +67,15 @@ Page {
           title: qsTr( "Consider mouse as a touchscreen device" )
           description: qsTr( "If disabled, the mouse will act as a stylus pen." )
           settingAlias: "mouseAsTouchScreen"
+      }
+      Component.onCompleted: {
+          var i;
+          for (i = 0; i < count; i++) {
+              if (get(i).settingAlias == 'nativeCamera')
+                  setProperty(i, 'isVisible', platformUtilities.supportsNativeCamera)
+              else
+                  setProperty(i, 'isVisible', true)
+          }
       }
   }
 
@@ -181,6 +191,8 @@ Page {
 
                       delegate: Row {
                           width: parent ? parent.width - 16 : undefined
+                          height: !!isVisible ? undefined : 0
+                          visible: !!isVisible
 
                           Column {
                               width: parent.width - toggle.width
--- a/src/qml/QFieldSettings.qml
+++ b/src/qml/QFieldSettings.qml
@@ -69,12 +69,12 @@ Page {
           settingAlias: "mouseAsTouchScreen"
       }
       Component.onCompleted: {
-          var i;
-          for (i = 0; i < count; i++) {
-              if (get(i).settingAlias == 'nativeCamera')
-                  setProperty(i, 'isVisible', platformUtilities.supportsNativeCamera)
-              else
-                  setProperty(i, 'isVisible', true)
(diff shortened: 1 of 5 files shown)


src/qml/QFieldSettings.qml
```diff
@@ -67,6 +68,15 @@ Page {
           description: qsTr( "If disabled, the mouse will act as a stylus pen." )
           settingAlias: "mouseAsTouchScreen"
       }
+      Component.onCompleted: {
+          var i;
```
m-kuhn, line 72: @suricactus I need your JS super eye

qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 6d8cecda53e335bc479cad47ca3f0b464127a1a5:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-44f61055d933e4b8016d2dd1c4970d0a79008b65-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-44f61055d933e4b8016d2dd1c4970d0a79008b65-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-44f61055d933e4b8016d2dd1c4970d0a79008b65-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-44f61055d933e4b8016d2dd1c4970d0a79008b65-x86.apk)

src/qml/QFieldSettings.qml
```diff
@@ -67,6 +68,15 @@ Page {
...
+          for (i = 0; i < count; i++) {
+              if (get(i).settingAlias == 'nativeCamera')
+                  setProperty(i, 'isVisible', platformUtilities.supportsNativeCamera)
+              else
+                  setProperty(i, 'isVisible', true)
+          }
+      }
   }
```
suricactus, line 80:
```suggestion
      Component.onCompleted: {
          for (var i = 0; i < settingsModel.count; i++) {
              if (settingsModel.get(i).settingAlias === 'nativeCamera') {
                  settingsModel.setProperty(i, 'isVisible', platformUtilities.supportsNativeCamera)
              } else {
                  settingsModel.setProperty(i, 'isVisible', true)   
              }
          }
      }
  }
```

suricactus: @m-kuhn I would write it like this:
1) curly brackets in js are cheap, since they are the same line as the opening statement
2) always use `===` instead of `==`
3) `for (var i = 0; i < l; i++ )` is valid syntax, no need to predefine the `var i`. The only problem with `var` is it is not deleted after the loop, but it is fine.
4) warning, opinion: I think we should not abuse the implicit usage of the QML element's properties and methods. When looking at your diff, I was a bit confused where this `count` variable comes from and I had to check the whole file and context. Now it is clear it is coming from the model. The same is valid for `get` and `setProperty`.

src/qml/QFieldSettings.qml
suricactus: I have answered in a new comment.

src/qml/QFieldSettings.qml
```diff
@@ -181,6 +191,8 @@ Page {
 
                       delegate: Row {
                           width: parent ? parent.width - 16 : undefined
+                          height: !!isVisible ? undefined : 0
```
suricactus, line 194: So by default `isVisible` is assumed to be `false`? I assume `height: undefined` will automatically calculate the height, but note `isVisible` will enforce it for `false`, `undefined`, `null`, `''` and `0`. So if `isVisible` is not set (or set to `undefined`), you will get `height: 0`.

m-kuhn: It's set for each item in the loop, so the default value is never used. Does that work?

suricactus (approved):

src/qml/QFieldSettings.qml
m-kuhn: Agreed on each point. Except for 3, where I think that the "problem" is actually a feature :+1:

qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 0748d1e4c06dc04009df9ec65fefa3bb1faa2760:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-dd0775c30318a1c76cc6b6b830bd9d5f744e783a-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-dd0775c30318a1c76cc6b6b830bd9d5f744e783a-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-dd0775c30318a1c76cc6b6b830bd9d5f744e783a-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1960-dd0775c30318a1c76cc6b6b830bd9d5f744e783a-x86.apk)

PR #1996, opened 2021-06-13
Fix reshape tool handling of geometry operation return values

nirvn:
Simplify and fixing handling of geometry's avoid intersection operation return value (i.e., just rely on empty geometry check to see if something went wrong).

### DIFF
--- a/src/core/utils/geometryutils.cpp
+++ b/src/core/utils/geometryutils.cpp
@@ -70,16 +70,11 @@ QgsGeometry::OperationResult GeometryUtils::reshapeFromRubberband( QgsVectorLaye
       {
         QHash<QgsVectorLayer *, QSet<QgsFeatureId>> ignoredFeature;
         ignoredFeature.insert( layer, QSet<QgsFeatureId>() << fid );
-        if ( geom.avoidIntersections( avoidIntersectionsLayers, ignoredFeature ) != 0 )
-        {
-          layer->destroyEditCommand();
-          return QgsGeometry::NothingHappened;
-        }
+        geom.avoidIntersections( avoidIntersectionsLayers, ignoredFeature );
       }
 
       if ( geom.isEmpty() ) //intersection removal might have removed the whole geometry
       {
-        layer->destroyEditCommand();
         return QgsGeometry::NothingHappened;
       }
     }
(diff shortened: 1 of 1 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 60ed5025b2e330b1e8750116d01f857331265fb1:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1996-c007f86383df3e798fd060993654eb8f0bb6af70-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1996-c007f86383df3e798fd060993654eb8f0bb6af70-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1996-c007f86383df3e798fd060993654eb8f0bb6af70-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-1996-c007f86383df3e798fd060993654eb8f0bb6af70-x86.apk)

m-kuhn (approved):

geckeau:
Still have trouble with the combination "Avoid overlap on active layer" and the QField Reshape Tool.

Reproduce
- Create New QGIS project (tested with both QGIS 3.16.3 and QGIS 3.18.3)
- Create new Polygon Geopackage Layer
- Create new Multipolygon Geopackage Layer
- Activate Snapping option "Avoid overlap on active layer"
- Set up QFieldSync: offline editing for both layers
- Package QField Project
- Open in QField (tested with both official 1.9.6 and unstable version from Playstore as well as latest apk build from this pull)
- Add several Polygons (some of them with and some of them without overlapping during digitizing)
- Try to reshape

On the first try, reshape never works for me. Sometimes it suddenly works, when I reshape a Polygon without overlapping (i.e. snapping to all the vertices of the polygon during reshape, instead of drawing inside). After the first time that a reshape was successful without overlapping, a new level gets magically unlocked. Afterwards it will also work with overlapping. But just for a while. Then the errors return and have to repeat the reshape without overlapping. Generally reshape works better for isolated polygons than for polygons that were added right next to each other with avoid overlap function. No difference noticed between Polygons and Multipolygons.

nirvn:
@geckeau , I've tried replicating the issue with the steps you offered above, and I can't stumble on broken reshaping on current master APK:
image

Maybe try and share your project here? There might be additional settings from your environment that lead to an issue here.

geckeau:
@nirvn Thank you for looking into it! It is really strange, but after today's testing, the isssue seems to have disappeared, when using the current Master apk (Playstore unstable version). It persists only in older versions.

nirvn:
@geckeau , I'll take that as a win :wink: thanks for the follow up.

PR #2091, opened 2021-08-18
Fix feature list model crashes by gracefully handling pre-existing running gatherer

### LINKED ISSUE #2090: Crashes when using the autocompletion
I am trying to use autocompletion to select entries from a larger reference list (range 15.000 - 20.000 entries). This results in both reliably reproducible crashes and seemingly random crashes.

### Reproduction Steps:

1. Create an attribute form in which you can use autocomplete to select entries in a larger reference list for selection (strictly speaking, i have observed the problem so far only with the one reference list you can also find below in the mre. i don't have a second comparably structured list).
2. Tap the desired entry as soon as it appears in the list of entries filtered according to your input string.

### Expected behavior:

The tapped entry is selected and assigned to the corresponding attribute of the current feature.

### Observed behavior:

At the beginning of a QField session, filtering and selection usually works as expected. However, two limitations can be observed that prevent productive use of the function.

1) There are very specific search strings when using QField in combination with the used reference list can be crashed reliably, although the filtering of the entries always works very smoothly. But as soon as you try to select a filtered entry in these cases QField crashes.

2) Often but not reliably reproducible, QField (in combination with the large reference list used) gets into a situation where any interaction with autocomplete causes it to crash.

### Assumptions:

Regarding 1) I assume that the reliable reproducible cases have something to do with the size of the list of filtered entries. Above a certain number of entries, it crashes. Look at my example in the video below. I type "t s r" to filter for "Taraxacum sect. Ruderalia" to filter. Filtering works great as can be seen, but tapping the entry is followed by the inevitable crash. If I type something longer instead of "t s r" to further narrow down the results list, I am eventually allowed to select the desired entry.

Regarding 2) I see a trend here that the longer the QField session lasts, the higher the probability of a crash. Once it crashes, often no restart of the app helps and it crashes with every input attempt. The only thing that helps temporarily is restarting the entire system.

Another factor that I feel increases the likelihood of a crash is the speed at which I type or try to select a filtered entry after typing.

Bottom line: Both together makes me suspect that it might have something to do with memory management or similar (The system's working memory does not seem to be the problem, since there is usually still around 1 GB free)? All the single elements of the function on their own works, but presumably I'm pushing the system to its current limits with the size of the reference list? What is the largest reference list you guys have used so far without any problems? Is there such a thing as "official" specifications of QField's performance limits?

### mre

[mre_autocompletion_crash.zip](https://github.com/opengisch/QField/files/7001295/mre_autocompletion_crash.zip)

### Screenrecording

https://user-images.githubusercontent.com/80956612/129756769-bd069b05-d92a-49dc-a43a-ec6d5fbc38b5.mp4

### adb-log 

Here is also an adb log. Please let me know if and how I can create it more helpful, it's my first one ;) Is it better for you to have the verbose log? 

[adb_log_autocomplete_crash.txt](https://github.com/opengisch/QField/files/7001299/adb_log_autocomplete_crash.txt)

**QField version:** 1.9.6 (d5dd01)

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: Don't know.
* Problem can be reliably reproduced, doesn't happen randomly: Yes, see above.
* Problem happens with all files and projects, not only some files or projects: Presumably with all reference lists above a certain size?

nirvn:
… when preparing a new one.

This fixes #2090 on my machine, and I suspect a bunch more. @m-kuhn , you mentioned a postgis crash too, was there an open issue?

### DIFF
--- a/src/core/featurelistmodel.cpp
+++ b/src/core/featurelistmodel.cpp
@@ -281,6 +281,14 @@ void FeatureListModel::gatherFeatureList()
       request.setFilterExpression( QStringLiteral( " (%1) AND (%2) " ).arg( mFilterExpression, searchTermExpression ) );
   }
 
+  if ( mGatherer )
+  {
+    disconnect( mGatherer, &QThread::finished, this, &FeatureListModel::processFeatureList );
+    connect( mGatherer, &QThread::finished, mGatherer, &QObject::deleteLater );
+    mGatherer->stop();
+    mGatherer = nullptr;
+  }
+
   mGatherer = new FeatureExpressionValuesGatherer( mCurrentLayer, fieldDisplayString, request, QStringList() << keyField() );
   connect( mGatherer, &QThread::finished, this, &FeatureListModel::processFeatureList );
   mGatherer->start();
@@ -298,7 +306,11 @@ void FeatureListModel::processFeatureList()
   if ( mAddNull )
     entries.append( Entry( QStringLiteral( "<i>NULL</i>" ), QVariant(), QgsFeatureId() ) );
 
-  for ( const FeatureExpressionValuesGatherer::Entry &gatheredEntry : mGatherer->entries() )
+  const QVector<FeatureExpressionValuesGatherer::Entry> gatheredEntries = mGatherer->entries();
+  mGatherer->deleteLater();
+  mGatherer = nullptr;
+
+  for ( const FeatureExpressionValuesGatherer::Entry &gatheredEntry : gatheredEntries )
   {
     Entry entry;
 
@@ -336,9 +348,6 @@ void FeatureListModel::processFeatureList()
              : entry1.fuzzyScore > entry2.fuzzyScore;
     } );
   }
-
-  mGatherer->deleteLater();
-  mGatherer = nullptr;
 
   beginResetModel();
   mEntries = entries;


m-kuhn (approved):
Nice !

qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 2998104e9f45671c88e071e01ce3698f014816b0:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2091-b4eacf7494e805eac4ec1918e30b86d2f9f4bace-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2091-b4eacf7494e805eac4ec1918e30b86d2f9f4bace-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2091-b4eacf7494e805eac4ec1918e30b86d2f9f4bace-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2091-b4eacf7494e805eac4ec1918e30b86d2f9f4bace-x86.apk)

PR #2100, opened 2021-08-21
Fix data corruption when multi editing

### LINKED ISSUE #2082: Multi editing can lead to severe data corruption
### Disclaimer:

The following description is still a bit vague. I'm still troubleshooting and haven't had the time to test it extensively and to create an minimal, reproducible example. However, due to the potential severe impact of this bug, I wanted to make this post in advance. In parts also as a warning for other users and maybe someone else with more background knowledge already has an idea how to fix or at least avoid the problem.

### Szenario:

Layer A: Parent, Polygon Layer

Layer B: Child of A, Point Layer

IDs: Both Layers have a sequential number as ID, ID of Layer A is foreign key in Layer B

### Problem description:

Multi edits in Layer B sometimes leads to very strange behaviour. I assume it is only the case, if the value for a foreign key of an edited feature to Layer A, for example "123" also exists as primary key of a feature in Layer B. In this cases not only the selected features are manipulated by the multi edit but also is the feature in Layer B with the primary key "123".

In the process, two things happen that should not:

1) Unwanted changes in unselected feature

As a result the attributes of the feature in Layer B that has the primary key "123" get some unwanted attribute changes, the systematics of which I haven't quite understood yet. Fact is: The foreign key is set to 0 and as a result some values are moved one attribute column to the right (including the old foreign key).

2) Creation of polygons in a point layer

Another strange thing about this is that the point(!) feature with ID "123" now gets as geometry the polygon of the feature with ID "123" from Layer A. I didn't know that such a thing is even possible.

But that's how I noticed it in the first place, because on the one hand more and more points disappeared in my rather large project over time and on the other hand I had the case more and more often that I had selection results from the point layer B as well, although I had only clicked on a polygon of layer A. (Note: the polygons in layer B are of course invisible because the layer only has a point symbology. But the "Point-Polygons" can be reliably selected once you know where they are).

I guess a simple workaround would be to use only UUIDs as IDs, which i'll try (or of course to stop using the multi edits for the time being). However, in this case, I am only editing a client's dataset and did not create the data model myself.

In any case, the described data corruption by using the multi-edit function should be prevented.

Please let me know if you need more information on this. I would then try to create a minimal, reproducible example.

**QField version:** 1.9.6 (d5dd01)

nirvn:
This PR fixes a serious data corruption when multi editing features. The problem was that identifying additional features in the canvas while in multi editing mode would identify features from _all_ layers, and silently check/select them as part of the group of features being edited. It slipped through our (well, my :wink:) watchful eyes as the initial implementation did not auto select identified features.

The solution is to refine the identification of features to _only append features from the layer within which features are selected_. 

@lp-dj , this fixes #2082 . Feel free to test and confirm things are working as expected for you with these APKs.

### DIFF
--- a/src/core/identifytool.cpp
+++ b/src/core/identifytool.cpp
@@ -58,7 +58,7 @@ void IdentifyTool::identify( const QPointF &point ) const
 
   QgsPointXY mapPoint = mMapSettings->screenToCoordinate( point );
 
-  const QList<QgsMapLayer *> layers { mMapSettings->mapSettings().layers() };
+  const QList<QgsMapLayer *> layers = mModel->selectedLayer() ? QList<QgsMapLayer *>() << mModel->selectedLayer() : mMapSettings->mapSettings().layers();
   for ( QgsMapLayer *layer : layers )
   {
     if ( !layer->flags().testFlag( QgsMapLayer::Identifiable ) )


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for e9be8c28b8f0322580d6f3f954a2efbe2c77f67b:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2100-87bdbd6b162b8e29109dbf8ab100f3a6635a3664-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2100-87bdbd6b162b8e29109dbf8ab100f3a6635a3664-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2100-87bdbd6b162b8e29109dbf8ab100f3a6635a3664-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2100-87bdbd6b162b8e29109dbf8ab100f3a6635a3664-x86.apk)

m-kuhn (approved):

lp-dj:
Ok, now you were faster than me ;) But I was happy to see that the bugfix has arrived in the current QField Dev version via the PlayStore. Thanks a lot!

PR #2127, opened 2021-09-05
Add Android build instructions for Qt Creator

hieu-van:
Just trying to make life easier for those who favors Qt Creator over Docker :-)


CMakeLists.txt
```diff
@@ -178,10 +190,14 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "iOS")
   set(QFIELD_RESOURCE_FILES ${SHARE_PATH})
 endif()
 
-set (ENABLE_TESTS TRUE CACHE BOOL "Build unit tests?")
+set(ENABLE_TESTS CACHE BOOL "Build unit tests")
+IF(ANDROID)
+  set(ENABLE_TESTS FALSE)
+else()
+  set(ENABLE_TESTS TRUE)
```
m-kuhn, line 197: This will override any user defined (`CACHE`) value from line 193.
Either we leave it to the user to disable tests (via command line `-DENABLE_TESTS=FALSE`) or this is set before 193 and passed in as default value.

CMakeLists.txt
```diff
@@ -1,14 +1,17 @@
 #############################################################
 # CMake settings
 cmake_minimum_required(VERSION 3.5)
+cmake_policy(SET CMP0071 NEW)
+cmake_policy(SET CMP0074 NEW)
 
+set(QML_IMPORT_PATH ${CMAKE_SOURCE_DIR}/src/qml/imports CACHE PATH "")
```
m-kuhn, line 7: What's the reason for this? I can't see it used anywhere

CMakeLists.txt
```diff
@@ -18,6 +21,12 @@ project(${APP_NAME}
 )
 
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
+if(ANDROID)
+  # You probably have installed Qt for Android somewhere else
+  # other than /usr. In that case, this is necessary
+
+  LIST(APPEND CMAKE_FIND_ROOT_PATH "/")
```
m-kuhn, line 28: This looks suspiciously like a fix for the local setup and not a general config that belongs into `if(ANDROID)`.
I think that should be moved to the documentation to be passed as command line option instead.

CMakeLists.txt
```diff
@@ -157,6 +165,10 @@ find_package(Qt5 COMPONENTS PrintSupport QUIET)
 find_package(QGIS REQUIRED)
 if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "iOS")
   find_package(Proj)
+  if(ANDROID)
+    # If the build target is Android, we use OSGeo4A libraries instead
+    set(ENV{GDAL_ROOT} ${OSGEO4A_STAGE_DIR}/${ANDROID_ABI})
+  endif(ANDROID)
```
m-kuhn, line 171: Should this go into FindGDAL instead?

m-kuhn (commented):
Thanks a lot for improving the scripts and the documentation.
I have left a couple of comments for your consideration.

CMakeLists.txt
hieu-van: Qt Creator reports that it can't find `Theme` QML module, thus QML designer doesn't work. If this doesn't happen to you then I guess it's a local problem.

CMakeLists.txt
```diff
@@ -39,7 +46,7 @@ set (QFIELD_LIB_SUBDIR     ${DEFAULT_LIB_SUBDIR}     CACHE STRING "Subdirectory
 set (QFIELD_INCLUDE_SUBDIR ${DEFAULT_INCLUDE_SUBDIR} CACHE STRING "Subdirectory where header files will be installed")
 
 set (WITH_VCPKG OFF CACHE BOOL "Build with vcpkg packaging layout")
-set (USE_MAC_BUNDLING NOT ${WITH_VCPKG} CACHE BOOL "Use mac bundling")
+set (USE_MAC_BUNDLING ${WITH_VCPKG} CACHE BOOL "Use mac bundling")
```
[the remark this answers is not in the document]
hieu-van: When done configuring the value of `USE_MAC_BUNDLING` (in my case) is `NOT OFF`, displayed as a false value. This doesn't look right too I guess?

CMakeLists.txt
```diff
@@ -68,10 +75,11 @@ endif ()
 
 if (ANDROID)
   add_definitions(-DQGIS_INSTALL_DIR="") # TODO: do we need osgeo4a/[lib]/files here? see qgis.pri
-  set (ANDROID_TARGET_PLATFORM CACHE INT "The target android platform sdk version")
-  set (ANDROID_PACKAGE_SOURCE_DIR ${CMAKE_SOURCE_DIR}/android)
-  set (OSGEO4A_STAGE_DIR NONE CACHE DIRECTORY "OSGEO4A stage path (base path without architecture)")
-  set (OSGEO4A_LIB_DIR ${OSGEO4A_STAGE_DIR}/${ANDROID_ABI}/lib)
+  set(ANDROID_TARGET_PLATFORM CACHE INT "")
```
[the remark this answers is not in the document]
hieu-van: bruh I will add it back

CMakeLists.txt
hieu-van: Do we have to re-implement FindGDAL from the default module or what else?

CMakeLists.txt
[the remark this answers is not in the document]
m-kuhn: `USE_MAC_BUNDLING` on affects osx/ios and has no influence on android builds.
It should normally be enabled. Only in the case of vcpkg we want it to be disabled by default.

CMakeLists.txt
m-kuhn: I.e. this improves Qt Creator properly detecting the `Theme` module?
That sounds really nice!
Can you add a short comment that this is for Qt Creator IDE support?

CMakeLists.txt
hieu-van: Let's pass the variable in as default

CMakeLists.txt
m-kuhn: No, I realize now, that this is actually an input for FindGDAL, sorry I missed that before.

I am wondering if all this osgeo4a specific handling is the way to go or if we should rather patch `CMAKE_FIND_ROOT_PATH` as it's being done in scripts/docker-build.sh.

A `if(${OSGEO4A_STAGE_DIR}` -> `list(APPEND CMAKE_FIND_ROOT_PATH ${OSGEO4A_STAGE_DIR}/${ANDROID_ABI}` would be a lot less intrusive over all?

hieu-van: Without tinkering with `CMAKE_FIND_ROOT_PATH` CMake won't be able to locate libs in the stage path. Also, I find this snippet working instead:
```
if(OSGEO4A_STAGE_DIR)
  list(APPEND CMAKE_FIND_ROOT_PATH /)
endif()
```
If you agree I will push it.

m-kuhn: That looks cleaner.
Can we possibly have one single if for osgeo4a?

hieu-van: I've just collected all the necessary things into `if(ANDROID)` block. Please check the latest commit.

m-kuhn (approved):
Thanks, that looks good

Can you resolve the merge conflicts and then we are good to go?

m-kuhn (approved):

PR #2165, opened 2021-09-30
drop i18n folder (outdated)

3nids:


i18n/translations.qrc
```diff
@@ -1,47 +0,0 @@
-<RCC>
```
m-kuhn, line 1: This file might still be needed

m-kuhn: Or generated

m-kuhn:
It looks like someone copied .qm files from qt into the source code here
@m-kuhn why did you do that?

m-kuhn:
No idea ... we should probably take those files from the installation or remove them from the resource file

nirvn:
FYI, we should dynamically generate the translations.qrc to avoid missing translations (see the Russian discovery in PR #2186 )

3nids:
superseded by https://github.com/opengisch/QField/pull/2190

PR #2229, opened 2021-10-16
Add a brand new show labels checkbox in the layer tree item properties pop up

nirvn:
This PR adds a new [x] show labels checkbox option in the layer tree items, a useful feature to have in the field when users might want to temporarily disable labels from specific vector layer(s).

The checkbox is only visible when two conditions are met: 1/ the layer tree item is a vector layer, and 2/ the vector layer has a labeling engine already set up in the loaded project.

Looks:
image

### DIFF
--- a/src/core/layertreemodel.cpp
+++ b/src/core/layertreemodel.cpp
@@ -759,6 +759,40 @@ QVariant FlatLayerTreeModelBase::data( const QModelIndex &index, int role ) cons
       return mTreeLevelMap.contains( index.row() + 1 ) && mTreeLevelMap[index.row() + 1] > mTreeLevelMap[index.row()];
     }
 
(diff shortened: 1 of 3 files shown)


nirvn:
Note: just like the [x] show on map option, it is ephemeral and won't be remembered across session. For users in need of remembering labels (and layers) visibility across session, configuring map themes is still the way to go.

nirvn:
The overall goal here is to follow up with a PR to slightly improve our support for opening of individual datasets by enabling (under a limited set of circumstances*) labels by default. 

(*) basically, if QgsVectorLayerUtils::guessFriendlyIdentifierField( fields ) return a friendly identifier, the dataset is considered as being label friendly and labels are shown. If not, then no labels.

PR #2409, opened 2022-01-16
Make sure the field visibility is enforced when opening the form of the feature created just before

### LINKED ISSUE #2406: Enhancement : Keep the expression for visibility (form)
Hello, 

When using expressions to control the visibility of containers (in attributes form) , the expression is working when we add a new entity, but all containers are shown when modifying the entity. 

For example, when creating the new entity, if "field 1" = 'X' , container 1 is shown. However, when modifying this entity, "field 1" stays as = 'X', and all the other containers are still visible.

I think, just like QGIS, QFIELD should check if the expression is valid when we chose to modify the entity and then show the right container. 

QField version: [1.10.0 (1da006 v1.10/0)]

## Additional information:

* Problem started happening recently, didn't happen in an older version of QField: [No]

suricactus:
Fixes #2406 

### Reproduction

1) create a feature with dependent fields. Make sure the newly created feature hides some of the fields.
2) open the newly created feature feature form.
3) the hidden fields are visible.

### Code comments

The problem was that `AttributeFormModelBase::applyFeatureModel` was not called once the form is opened.

Probably "the best solution" would be to use the `FeatureListModelSelection::focusedItemChanged()` event to call `AttributeFormModelBase::applyFeatureModel`. However, this would make more spaghetti, than explicitly calling the method.

### DIFF
--- a/src/core/attributeformmodel.cpp
+++ b/src/core/attributeformmodel.cpp
@@ -80,6 +80,11 @@ QVariant AttributeFormModel::attribute( const QString &name )
   return mSourceModel->attribute( name );
 }
 
+void AttributeFormModel::applyFeatureModel()
+{
+  return mSourceModel->applyFeatureModel();
+}
+
 bool AttributeFormModel::filterAcceptsRow( int source_row, const QModelIndex &source_parent ) const
 {
   return mSourceModel->data( mSourceModel->index( source_row, 0, source_parent ), CurrentlyVisible ).toBool();
--- a/src/core/attributeformmodel.h
+++ b/src/core/attributeformmodel.h
@@ -94,6 +94,9 @@ class AttributeFormModel : public QSortFilterProxyModel
      */
     Q_INVOKABLE QVariant attribute( const QString &name );
 
+    //! Forces the form to update the fields visibility and constraints
+    Q_INVOKABLE void applyFeatureModel();
+
   signals:
     void featureModelChanged();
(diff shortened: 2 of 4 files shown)


duke-nyuki:
Task linked: [QF-1600 Make sure the field visibility is enforced when opening the form of the feature created just before](https://app.clickup.com/t/2192114/QF-1600)

nirvn (approved):

qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 0a855f58c7f8b74cca30eaed326eb312afad26fe:
  - [**arm64_v8a**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2409-da14d3677248dc836e689704a4aaad9f0fc2a415-arm64_v8a.apk)

Other architectures: [armv7](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2409-da14d3677248dc836e689704a4aaad9f0fc2a415-armv7.apk), [x86_64](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2409-da14d3677248dc836e689704a4aaad9f0fc2a415-x86_64.apk), [x86](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2409-da14d3677248dc836e689704a4aaad9f0fc2a415-x86.apk)

PR #2723, opened 2022-04-08
Added a testimonial section on the index page

olivierdalang:
Took some random comments from the google play store, not sure if allowed ?

<kbd>

</kbd>

<kbd>

</kbf>

### DIFF
--- a/css/qfield.css
+++ b/css/qfield.css
@@ -129,7 +129,7 @@ footer.footer {
 #about { 
   background-color: #212121;
   height: auto;
-  margin: 40px auto 0px auto; 
+  margin: 0 auto 0px auto;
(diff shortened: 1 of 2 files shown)


mbernasocchi:
@olivierdalang Thanks a lot for this, sorry on the late answer, we're looking for high profile users to add the text to your design.

mbernasocchi:
@m-kuhn could you add the Deutsche bahn text?

@olivierdalang could we add something like https://www.freshworks.com/freshdesk/lp/home/#customers higher than the testimonials here

boardend:
Continued as https://github.com/opengisch/QField/pull/5314

PR #2895, opened 2022-05-13
Do *not* use qgis data directory as profile directory

nirvn:
This PR fixes a pretty serious issue on Android (and iOS):
- on Android, we were setting the QGIS user profile directory to match that of QGIS' data resources (i.e. what is supposed to be a read-only directory that provides templates). It did work in part because qgis.db was in that directory to begin with, _however_ it meant that at every QField update, we'd _overwrite_ the user qgis.db with the template one. Practically speaking, this meant that every time someone updates QField, he/she would lose _all saved bookmarks_ :fearful: 
- On iOS, we were setting the QGIS user profile to the same problematic location, on top of which it was a read-only directory, meaning bookmarks wouldn't even save across sessions.

This is one of these "how was it ever working to begin with" moments :wink:

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -172,7 +172,11 @@ int main( int argc, char **argv )
 #if defined( Q_OS_ANDROID )
   QString projPath = PlatformUtilities::instance()->systemGenericDataLocation() + QStringLiteral( "/proj" );
   qputenv( "PROJ_LIB", projPath.toUtf8() );
-  QgsApplication app( argc, argv, true, PlatformUtilities::instance()->systemGenericDataLocation() + QStringLiteral( "/qgis/resources" ), QStringLiteral( "mobile" ) );
+
+  const QDir rootPath = QStandardPaths::writableLocation( QStandardPaths::AppDataLocation );
+  rootPath.mkdir( QStringLiteral( "qgis_profile" ) );
+  const QString profilePath = QStandardPaths::writableLocation( QStandardPaths::AppDataLocation ) + QStringLiteral( "/qgis_profile" );
+  QgsApplication app( argc, argv, true, profilePath, QStringLiteral( "mobile" ) );
 
   QSettings settings;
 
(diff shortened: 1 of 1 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for db5b827cd8aca2ec5e156f499489f6861ff1dff1:
  - [**arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2895-757baaa63ad75606e8a9b084c08c7fef7f8f82d6-arm64-android.apk)

Other architectures: [arm-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2895-757baaa63ad75606e8a9b084c08c7fef7f8f82d6-arm-android.apk), [arm-neon-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2895-757baaa63ad75606e8a9b084c08c7fef7f8f82d6-arm-neon-android.apk), [x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2895-757baaa63ad75606e8a9b084c08c7fef7f8f82d6-x64-android.apk), [x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-2895-757baaa63ad75606e8a9b084c08c7fef7f8f82d6-x86-android.apk)

src/app/main.cpp
```diff
@@ -172,19 +172,30 @@ int main( int argc, char **argv )
 #if defined( Q_OS_ANDROID )
   QString projPath = PlatformUtilities::instance()->systemGenericDataLocation() + QStringLiteral( "/proj" );
   qputenv( "PROJ_LIB", projPath.toUtf8() );
-  QgsApplication app( argc, argv, true, PlatformUtilities::instance()->systemGenericDataLocation() + QStringLiteral( "/qgis/resources" ), QStringLiteral( "mobile" ) );
```
suricactus, line 175: Why don't you move this logic to platformutils?

nirvn: @suricactus , I felt it wasn't needed since neither Linux nor Windows are in need of this.

suricactus: I was more because (almost) the same code is a few lines below. But if you say it is not needed, you are more familiar with the change. Feel free to merge.

suricactus (approved):

src/app/main.cpp
nirvn: @suricactus , right, for the iOS lines, it might further change based on @3nids ' own testing and findings.

m-kuhn:
> On iOS, we were setting the QGIS user profile to the same problematic location, on top of which it was a read-only directory, meaning bookmarks wouldn't even save across sessions

Would that also explain wmts troubles on ios (r/o cache dir)? Cc @3nids

3nids:
sadly not yet.

PR #3076, opened 2022-07-15
Remove szip overlay

m-kuhn:
Should build with upstream port too.


nirvn (approved):
APK running fine, confirmed as working. Nice one.

Issue #1085, opened 2020-06-26
QRCode scan feature


atabaraud:
The idea is to implement the possibility of scanning a QRCODE to fill a field in a form with the raw value of the QRCODE

Kabouik:
That would be super useful, I agree.

miili:
The library should also support scanning of barcodes.

marteinnm:
Has there been any development on this issue?

miili:
Is there an update, can one help with the implementation of this feature?

miili:
Some QML QRCode  // barcode resources:
https://scythe-studio.com/en/blog/how-to-scan-barcodes-in-qt-qml-application%22%20
https://www.dynamsoft.com/codepool/qt-qml-qr-code-scanner.html

This seems to be rather low hanging and can be integrated into `src/qml/editorwidgets/TextEdit.qml`.
How could this attribute _allowing QR scanning_ be communicated from the Attribute Form to QField's TextEdit?

nirvn:
@miili , @marteinnm , @Kabouik , @atabaraud , this has been fully implemented in this PR (https://github.com/opengisch/QField/pull/3142). Feel free to test the APKs in there and report on your experience.

Kabouik:
That's great news! For the record, I still can't see that option (no three-dot menu in the Beekeper field in the example projectcddc) despite having subscribed to the unstable branch on the store, and reinstalled Qfield. The about page says I'm using master-dev (dc1f4a) which seems to be the latest commit, but I see that PR #1085 is now open again?

Kabouik:
Actually I assumed it was merged when you posted, but I was probably wrong. I'll wait and try later when the PR is merged.

PR #3104, opened 2022-07-24
QR (and other bar) code reading search bar functionality

nirvn:
This PR implements QR (and other bar-type) code scanning / decoding functionality to QField through the search bar. Its UI/UX is quite simple: when clicking on the search bar button, a new QR code icon is now present embedded in the search text field:

When clicking on it, a new code reader panel opens. Users can then simply focus their camera on a QR/bar code to decode. A visual "flash" feedback  occurs as soon as QField successful decoded to string. The decoded string appears below the camera frame for users to review:

When satisfied, users simply click on the ( :heavy_check_mark: ) button to trigger a search against that string. For e.g. here, the QR code was a user identifier linked to several features in a project dataset:

Other cool things can be achieved here (e.g., encoding a lat,lon string into a QR code), sky's the limit. Well, the search bar functionality is really :wink: 

There is a flashlight to help scanning codes in low light conditions.

Fixes https://github.com/opengisch/QField/issues/1085

_A follow up PR will be opened to add code scanning / decoding in the feature form text edit widget._


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for ccc2bd4d9ed9f88e8ecf593b3182dab35c42ca63: [**arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3104-95e21f5defd3b6067a82b6f36d3fa2aaf0517c30-arm64-android.apk)
<details>

- [arm-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3104-95e21f5defd3b6067a82b6f36d3fa2aaf0517c30-arm-android.apk)
- [x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3104-95e21f5defd3b6067a82b6f36d3fa2aaf0517c30-x64-android.apk)
- [x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3104-95e21f5defd3b6067a82b6f36d3fa2aaf0517c30-x86-android.apk)
</details>

CMakeLists.txt
```diff
@@ -188,6 +188,9 @@ if(MSVC)
   add_link_options("$<$<NOT:$<CONFIG:Debug>>:/OPT:ICF>")
 endif()
 
+
+find_package(ZXing REQUIRED)
```
m-kuhn, line 192: Can you move that into src/core/CMakeLists.txt?

nirvn: Done.

nirvn:
Also, for the viewers at home, iOS failure is expected. It'll all magically fix itself when we move the iOS builds to vcpkg.

src/core/qgismobileapp.cpp
```diff
@@ -149,6 +149,8 @@
 #include <qgsproject.h>
 #include <qgsprojectstorage.h>
 #include <qgsprojectstorageregistry.h>
+
+#include <ZXing/ReadBarcode.h>
```
m-kuhn, line 153: Is this include used for anything?

src/app/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 find_package(${QT_PKG}QuickCompiler QUIET)
 
+if(NOT Qt5QuickCompiler_FOUND
+   OR CMAKE_BUILD_TYPE MATCHES Debug
+   OR CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
+  set(RESOURCES ${CMAKE_SOURCE_DIR}/src/qml/qml.qrc)
+endif()
```
m-kuhn, line 7: This looks unrelated

src/core/qgismobileapp.cpp
nirvn: Oups, it was used to test things here, will remove.

src/app/CMakeLists.txt
nirvn: Right, I need to remove this commit. We still need to find a way for QField to compile properly on linux without vcpkg.

src/core/barcodedecoder.h
```diff
@@ -0,0 +1,60 @@
...
+
+  signals:
+    void decodedStringChanged();
+
+  private:
+    QImage mImage;
+    QString mDecodedString;
+};
+
+class BarcodeDecoderThread : public QThread
```
m-kuhn, line 46: Can we move that to a private header?

nirvn: Or we move the class into the BarcodeVideoFilter cpp file? It's only used there ATM.

src/qml/CMakeLists.txt
```diff
@@ -1,16 +1,17 @@
 find_package(${QT_PKG}QuickCompiler QUIET)
 find_package(${QT_PKG} COMPONENTS Quick)
 
+file(GLOB_RECURSE QML_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/*.qml")
```
m-kuhn, line 4: Can we merge https://github.com/opengisch/QField/pull/3106 and rebase?

nirvn: @m-kuhn , sounds good, approved that PR just now.

m-kuhn:
> Edit: just noticed ubuntu actually has the library (and the -dev counterpart) in its repository, it's probably not worth dragging the whole 200,000 lines of code in our source tree then.

Good, I think vendoring dependencies is a last resort only, we should refrain from doing so and stick to upstream

nirvn:
And just like that, thanks to @m-kuhn Herculean efforts, iOS turns green :)

m-kuhn:
Did you just add a 🔦 ???

nirvn:
@m-kuhn , :)

This is ready for review and merging. I'll implement the feature form functionality in a different PR.

src/core/barcodedecoder.cpp
```diff
@@ -0,0 +1,80 @@
...
+  if ( imageFormat != ZXing::ImageFormat::None )
+  {
+    ZXing::ImageView imageView( image.bits(), image.width(), image.height(), imageFormat, static_cast<int>( image.bytesPerLine() ) );
+
+    ZXing::DecodeHints hints;
+    hints.setFormats( ZXing::BarcodeFormat::Any );
+    hints.setTryRotate( false );
+
+    ZXing::Result result = ZXing::ReadBarcode( imageView, hints );
+    QString resultText = QString::fromWCharArray( result.text().c_str() );
```
m-kuhn, line 72:
```suggestion
    ZXing::Result result = ZXing::ReadBarcode( imageView, hints );
    const std::wstring text = result.text();
    QString resultText = QString::fromWCharArray( text.c_str() );
```

Just to be sure we really get the 16 bit return value

src/core/barcodedecoder.h
```diff
@@ -0,0 +1,46 @@
...
+
+    Q_PROPERTY( QString decodedString READ decodedString WRITE setDecodedString NOTIFY decodedStringChanged )
+
+  public:
+    explicit BarcodeDecoder( QObject *parent = nullptr );
+
+    QString decodedString() const { return mDecodedString; }
+    void setDecodedString( const QString &decodedString );
+
+    void decodeImage( const QImage &image );
```
m-kuhn, line 36: Worth a test?

m-kuhn (commented):
C++ part looks good, some minor things/questions

@suricactus could you have an eye on the QML parts?

nirvn:
@m-kuhn , thanks for the review, comments addressed. As for the :flashlight: , it's implemented, but my phone doesn't seem to register its flash as a torch. Either that or we're facing another 5.14 bug.

mbernasocchi:
@nirvn @m-kuhn is ther a way to trigger another apks build? :)

src/core/barcodedecoder.cpp
```diff
@@ -0,0 +1,82 @@
...
+    hints.setTryRotate( true );
+
+    ZXing::Result result = ZXing::ReadBarcode( imageView, hints );
+    const std::wstring text = result.text();
+    QString resultText = QString::fromWCharArray( text.c_str() );
+    if ( !resultText.isEmpty() && mDecodedString != resultText )
+    {
+      mDecodedString = resultText;
+      emit decodedStringChanged();
+    }
```
suricactus, line 78: If I get it right, it means if I have `"scanned barcode"` already set from the previous scan and I scan again and I scan empty string `""`, there will be `decodedStringChanged()`, therefore the UI would not update?

Is it on purpose?

nirvn: Correct.

src/core/barcodevideofilter.cpp
```diff
@@ -0,0 +1,143 @@
...
+    {
+    }
+
+  private:
+    void run() override
+    {
+      if ( mDecoder )
+      {
+        mDecoder->decodeImage( mImage );
+      }
```
suricactus, line 38: Just wondering how many calls it has per second? If they are too much, we can drain the battery like crazy, probably a small sleep makes sense, so we do 1-2 scans per second max?

src/core/barcodevideofilter.h
```diff
@@ -0,0 +1,51 @@
...
+  public:
+    explicit BarcodeVideoFilter();
+
+    BarcodeDecoder *decoder() const { return mDecoder; }
+    void setDecoder( BarcodeDecoder *decoder );
+
+    void decodeVideoFrame( QVideoFrame *input );
+    bool isDecoding() const { return mDecodingThread; }
+
+    QVideoFilterRunnable *createFilterRunnable() override;
```
suricactus, line 41: A bit of docs would be great.

src/qml/qgismobileapp.qml
```diff
@@ -2994,6 +2994,22 @@ ApplicationWindow {
       }
   }
 
+  MouseArea {
+    id: barcodeReaderCatcher
+    anchors.fill: parent
+    enabled: barcodeReader.visible
+
+    onClicked: {
+      // Needed to avoid people modifying context within which the code reader was called
```
suricactus, line 3003: Sorry, it is not clear what context you are talking about and how the context could be modified?

src/qml/BarcodeReader.qml
```diff
@@ -0,0 +1,267 @@
...
+import org.qfield 1.0
+
+import Theme 1.0
+
+Popup {
+  id : barcodeReader
+
+  signal decoded(var string)
+
+  property int popupWidth: mainWindow.width <= mainWindow.height ? mainWindow.width - 40 : mainWindow.height - 40
```
suricactus, line 16: While not sure if it is worth it for this PR, I believe we should make all this magic numbers Theme variables.   In this case is `popupScreenEdgeMargin` I guess.

src/qml/BarcodeReader.qml
```diff
@@ -0,0 +1,267 @@
...
+            strokeColor: "white"
+            strokeStyle: ShapePath.SolidLine
+            joinStyle: ShapePath.MiterJoin
+            fillColor: "transparent"
+
+            startX: 20
+            startY: 60
+            PathLine { x: 20; y: 25 }
+            PathArc { x: 25; y: 20; radiusX: 5; radiusY: 5 }
+            PathLine { x: 60; y: 20 }
```
suricactus, line 179: I hope we never have to change these values, as we have to recalculate them manually. Other option would be to make them dynamically calculated like:
```
ShapePath {
    startX: 20
    startY: 60
    property int borderRadius:  5
    PathLine { x: parent.startX; y: parent.startX + parent.borderRadius }
    // ...
 }
```
No strong opinion here.

suricactus (approved):
Fantastic work! :iphone: 📓

src/core/barcodevideofilter.cpp
nirvn: I've discussed this with @suricactus yesterday. To leave a record of our discussion: capping decoding to say 2 scans per second would probably save some battery (not that much as the camera is what sucks battery out), but it risks having situations where the environment isn't great (shacky hands like mine, shacky barcode, or both) and we'd hit "a bad frame" twice per second and end up having to wait a long longer for the barcode to be decoded.  When decoding as much as possible, we maximize chance of success.

Hence thinking we should go for the decode-as-many-frames-as-possible to begin with.

src/qml/BarcodeReader.qml
nirvn: @suricactus , very good idea sir.

nirvn:
@suricactus , comments addressed, all very useful.

PR #3158, opened 2022-08-02
do not fetch if not required

3nids:


m-kuhn (approved):

m-kuhn (commented):
Looks like it's required -- at least as long as the script lives in a separate file

PR #3210, opened 2022-08-18
Add meta link canonical

m-kuhn:
this is the legacy doc, the new one is on docs.qfield.org.
to make search engines index the new page instead

```bash
find . -name '*.html' | xargs sed -i '/link.*qfield.css/a \    <link rel="canonical" href="https:\/\/docs.qfield.org" \/>'
```


suricactus:
What if we just delete all the unneeded pages, and add `404.html` with that redirect `<meta >`?

faebebin (approved):
LGTM :+1: 
According to https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#define-canonical seems like a good solution.
The href of the file I chose randomly out the 2664 led to the correct address.

PR #3213, opened 2022-08-19
Show visible features

domi4484:
image

PR #3249, opened 2022-08-25
Layer properties popup reorganizing

nirvn:
With the recent addition of the show label checkbox, the layer opacity, and the show visible features list, the layer properties popup has reached a state of overcrowdedness :)

This PR proposes some modest improvements:
image

The changes include:
- Moving the rarely-used 'reload data' to sit at the top-right of the popup as a label-less reload icon tool button
- Removing the expend legend items checkbox (which was cutting through styling-related items) and moving the functionality to a double-tap on legend item action (feels a lot more natural and de-crowds the popup)
- Putting the show >>visible<< features list action into a new dropdown menu attached to the show feature list button

I feel pretty good about all of these changes, and it's nice to have a new dropdown menu functionality coded into QfButton, can't wait to see how many more places end up using it in the coming months :)

On the expanding / collapsing of legend items, we had two possible actions to rely on: the double tap (felt natural), or the tap and hold. Both of these actions until this PR triggered the layer properties to pop up. If someone feels strongly about using the long tap to collapse and keep the double tap to open the pop up, let me know.

### DIFF
--- a/src/qml/LayerTreeItemProperties.qml
+++ b/src/qml/LayerTreeItemProperties.qml
@@ -34,6 +34,9 @@ Popup {
     itemVisibleCheckBox.checked = layerTree.data(index, FlatLayerTreeModel.Visible);
     itemLabelsVisibleCheckBox.checked = layerTree.data(index, FlatLayerTreeModel.LabelsVisible);
 
+    expandCheckBox.text = layerTree.data( index, FlatLayerTreeModel.Type ) === 'group' ? qsTr('Expand group') : qsTr('Expand legend item')
+    expandCheckBox.checked = !layerTree.data(index, FlatLayerTreeModel.IsCollapsed)
+
     reloadDataButtonVisible = layerTree.data(index, FlatLayerTreeModel.CanReloadData)
     zoomToButtonVisible = layerTree.data(index, FlatLayerTreeModel.HasSpatialExtent)
     showFeaturesListButtonVisible = isShowFeaturesListButtonVisible();
@@ -105,6 +108,21 @@ Popup {
         text: '<img src="' + invalidIcon + '" width="' + invalidSize + '" height="' + invalidSize + '"> '
               + qsTr('This layer is invalid. This might be due to a network issue, a missing file or a misconfiguration of the project.')
         font: Theme.tipFont
+      }
+
+      CheckBox {
+        id: expandCheckBox
+        Layout.fillWidth: true
+        topPadding: 5
+        bottomPadding: 5
+        text: qsTr('Expand legend item')
+        font: Theme.defaultFont
+        visible: index && layerTree.data(index, FlatLayerTreeModel.HasChildren) ? true : false
+
+        onClicked: {
+          layerTree.setData(index, checkState === Qt.Unchecked, FlatLayerTreeModel.IsCollapsed);
+          close()
+        }
       }
 
       CheckBox {
--- a/src/qml/LayerTreeItemProperties.qml
+++ b/src/qml/LayerTreeItemProperties.qml
@@ -47,21 +47,43 @@ Popup {
         ? qsTr('Stop tracking')
         : qsTr('Setup tracking')
 
-    opacitySliderVisible = layerTree.data(index, FlatLayerTreeModel.HasSpatialExtent)
+    opacitySliderVisible = layerTree.data( index, FlatLayerTreeModel.Type ) !== 'group' &&
+                           layerTree.data(index, FlatLayerTreeModel.HasSpatialExtent)
   }
 
   Page {
     width: parent.width
     padding: 10
-    header: Label {
-      padding: 10
-      topPadding: 15
-      bottomPadding: 0
-      width: parent.width - 20
-      text: title
-      font: Theme.strongFont
-      horizontalAlignment: Text.AlignHCenter
-      wrapMode: Text.WordWrap
+    header: RowLayout {
+      spacing: 2
+      Label {
+        id: titleLabel
+        Layout.fillWidth: true
+        Layout.leftMargin: 10
+        topPadding: 4
+        bottomPadding: 4
+        text: ''
+        font: Theme.strongFont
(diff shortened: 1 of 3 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 53cf5cefebff15f2b84f93ed76c38548bc652632: [**arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3249-7b795f9f9bcf943c4e42a07a225ca0cd61e318d3-arm64-android.apk)
<details>

- [arm-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3249-7b795f9f9bcf943c4e42a07a225ca0cd61e318d3-arm-android.apk)
- [x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3249-7b795f9f9bcf943c4e42a07a225ca0cd61e318d3-x64-android.apk)
- [x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3249-7b795f9f9bcf943c4e42a07a225ca0cd61e318d3-x86-android.apk)
</details>

suricactus:
I would propose two things here:
1) move show labels and refresh to a hamburger menu on top right, instead of just the "refresh" button there.
2) keep "Expand legen item" button for some time, but we show a message popup suggesting "hey, double tap on the legend item, it works".
3) the small arrow of the legend items on the left should also be single clickable, it's quite intuitive from QGIS and other software that it is collapsible. I know it's hard to click it, but would be nice for people with a pen or tiny fingers.
3) I am not really convinced about the arrow down button. Not sure if a separate button is worse or better. Having the filter on the feature list makes most sense to me. I like the use case that @marioba proposed here, it makes things even more useful and discoverable https://github.com/opengisch/QField/pull/3213#issuecomment-1227039878.

marioba:
> * move show labels and refresh to a hamburger menu on top right, instead of just the "refresh" button there.

Ok, but I don't have a strong opinion on this

> * keep "Expand legen item" button for some time, but we show a message popup suggesting "hey, double tap on the legend item, it works".

I agree on the second part, but I would still keep "Expand legend item", not only temporary

> * the small arrow of the legend items on the left should also be single clickable, it's quite intuitive from QGIS and other software that it is collapsible. I know it's hard to click it, but would be nice for people with a pen or tiny fingers.

I agree

> * I am not really convinced about the arrow down button. Not sure if a separate button is worse or better. Having the filter on the feature list makes most sense to me. I like the use case that @marioba proposed here, it makes things even more useful and discoverable [Show visible features #3213 (comment)](https://github.com/opengisch/QField/pull/3213#issuecomment-1227039878).

:+1:

nirvn:
>> move show labels and refresh to a hamburger menu on top right, instead of just the "refresh" button there.

I do have a strong -1 on this one :) I consume a lot of standalone datasets, and I use this all the time. I think that needs to be staying in plain sight the same way show/hide layer and opacity are.

On the possibility of a hamburger menu, I'm +1 on the idea, that's something I had in mind too, however I'd wait to introduce it here until we have two actions in the header. ATM, we have refresh only (for label, see paragraph above :wink:)

>> keep "Expand legen item" button for some time, but we show a message popup suggesting "hey, double tap on the legend item, it works".

>> I agree on the second part, but I would still keep "Expand legend item", not only temporary

Seems we have strong feelings about wanting the checkbox to stay here. @m-kuhn , care to chip in? I can re-introduce the checkbox for now. As for the double click action, do we want to revert it to opening the layer properties or do we think it's a good idea to have it change collapsed state as is proposed here?

>> the small arrow of the legend items on the left should also be single clickable, it's quite intuitive from QGIS and other software that it is collapsible. I know it's hard to click it, but would be nice for people with a pen or tiny fingers.

OK, we can implement that in a separate PR. 

>> I am not really convinced about the arrow down button. Not sure if a separate button is worse or better. Having the filter on the feature list makes most sense to me. I like the use case that @marioba proposed here, it makes things even more useful and discoverable https://github.com/opengisch/QField/pull/3213#issuecomment-1227039878.

OK, my thoughts on this:
- I don't think it can replace the layer properties action with this filter proposal for one important reason: for remote layers (say postgis, wfs), or even insanely large on-disk datasets (I'm thinking of this efi project with like gb-sized geopackages), triggering a _full features list_ takes a long, long time. Much like QGIS itself, it's useful to open and populate the list with a subset of the whole dataset to avoid long lags. So I think we need to keep the action in there. And so I think we still need to figure out how to place it in the popup. Any objections to the drop down action here?
- Considering that the feature list is mostly opened via finger tapping on features that are visible within the map canvas extent, I'm not sure how the filter button here would ultimately be of use 99% of the time :) 
- We could have the filter button not only filter features _out_ but also _add_ features that are within a given map extent, but since our feature list is not bound to a specific layer at any given time, how would we handle this? We add all features from all vector layers, or only the ones from identifiable layers, or only features from layer(s) listed in the feature list upon first finger tap?

suricactus:
> As for the double click action, do we want to revert it to opening the layer properties or do we think it's a good idea to have it change collapsed state as is proposed here?
I like the collapse/uncollapse idea, it's intuitive.

nirvn:
OK, I've re-added the expand legend item/group checkbox in the UI, which by the above comments felt like the only thing that wasn't liked by neither Mario nor Ivan. I think the rest was acceptable to all:
image

As for a possible feature list UI/UX to filter a preexisting list, let's keep the discussion alive, it'd be implemented separately anyways. And it wouldn't remove the need for a view visible feature list action in the layer properties to be able to populate an initial feature list that isn't the whole dataset.

nirvn:
Nice example of why this PR is needed:
image

In landscape with the reorganizing, we juuust fit in.

marioba:
> do we think it's a good idea to have it change collapsed state as is proposed here?

+1

> * Any objections to the drop down action here?

No, let's do that. Your arguments make sense.

PR #3270, opened 2022-08-30
Make the slider opacity use the new QfSlider widget

suricactus:
The main UI difference is that the values is shown on the right.

image

Also, fix double feature count shown in layer properties panel.

### DIFF
--- a/src/qml/LayerTreeItemProperties.qml
+++ b/src/qml/LayerTreeItemProperties.qml
@@ -208,7 +208,7 @@ Popup {
             from: 0
             to: 100
             stepSize: 1
-            suffixText: "%"
+            suffixText: " %"
             height: 40
 
             onMoved: function () {
--- a/src/qml/LayerTreeItemProperties.qml
+++ b/src/qml/LayerTreeItemProperties.qml
@@ -374,7 +374,7 @@ Popup {
           if (count !== undefined && count >= 0) {
               var countSuffix = ' [' + count + ']'
 
-              if (!titleLabel.text.endsWith(countSuffix))
+              if (!title.endsWith(countSuffix))
                   title += countSuffix
           }
         }
--- a/src/qml/LayerTreeItemProperties.qml
+++ b/src/qml/LayerTreeItemProperties.qml
@@ -199,12 +199,20 @@ Popup {
             text: qsTr("Opacity")
             font: Theme.defaultFont
           }
-          Slider {
+
+          QfSlider {
             Layout.fillWidth: true
+
             id: slider
-            value: index !== undefined ? layerTree.data(index, FlatLayerTreeModel.Opacity) : 0
-            onMoved: {
-              layerTree.setData(index, value, FlatLayerTreeModel.Opacity)
+            value: index !== undefined ? layerTree.data(index, FlatLayerTreeModel.Opacity) * 100 : 0
+            from: 0
+            to: 100
+            stepSize: 1
+            suffixText: "%"
+            height: 40
+
+            onMoved: function () {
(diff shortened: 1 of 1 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for da5f4d0dc37883199bef1cd2a9a564c6aae3bc44: [**arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3270-986c90ef1a03e17de9f2e470368d94d22e1c3fd2-arm64-android.apk)
<details>

- [arm-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3270-986c90ef1a03e17de9f2e470368d94d22e1c3fd2-arm-android.apk)
- [x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3270-986c90ef1a03e17de9f2e470368d94d22e1c3fd2-x64-android.apk)
- [x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3270-986c90ef1a03e17de9f2e470368d94d22e1c3fd2-x86-android.apk)
</details>

nirvn:
I'm wondering now, could we get rid of the label opacity sitting on top of the slider and just have the icon? Feels like it's explicit enough now that we have the value % next to the slider

nirvn (approved):
Tested, looks good.

suricactus:
Either this, or we have the icon on the same row as "opacity" label and have the slider full width. Would be much better IMO.

In a perfect universe, we can have "show on map" checkbox and the opacity slider integrated somehow. Something like:

- [x] Visible on map
--------------------------------------------o------- 80%

And if you click the checkbox, the slider gets hidden.

src/qml/LayerTreeItemProperties.qml
```diff
@@ -199,12 +199,20 @@ Popup {
...
+
             id: slider
-            value: index !== undefined ? layerTree.data(index, FlatLayerTreeModel.Opacity) : 0
-            onMoved: {
-              layerTree.setData(index, value, FlatLayerTreeModel.Opacity)
+            value: index !== undefined ? layerTree.data(index, FlatLayerTreeModel.Opacity) * 100 : 0
+            from: 0
+            to: 100
+            stepSize: 1
+            suffixText: "%"
```
m-kuhn, line 211:
```suggestion
            suffixText: " %"
```

Proposing a nonbreaking space

m-kuhn (approved):

src/qml/LayerTreeItemProperties.qml
suricactus: > There are various spacing rules for the percentage sign depending on the language being used (according to the writers who do [research paper services for Ultius](https://www.ultius.com/writing-services/research-paper-writing-services.html)). In English the percent sign and the number is corresponds to are not separated by space, but the [International System of Units](https://www.nist.gov/programs-projects/international-system-units-si) and ISO 31-0 standard require a non-breaking space between them.

> The Chinese percent sign never uses space because Chinese characters and words are never separated. In Swedish, German, French, Finnish, and Czech there is a non-breaking space between the number and percent sign. In Italian, the percent sign is never spaced. In Persian and Turkish, the percent sign precedes the number instead of following it (for example %3). It is possible that the 3% we know today evolved from using the slash sign in Latin-based languages prior to the use of the decimal point (an example is 1/100th or 1% written as 0/01).  Perhaps 0/0 evolved into the percent sign we know today: %.

I like it more without the space, as in English. But let's conform the ISO...

PR #3300, opened 2022-09-04
Spix improvements, message logs & WMS layer testing

nirvn:


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 71a0ffe66006d05e066230f7e83324667452a45d: [**arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3300-ff0d807c46ea5fed3195606c4ca9ef246c2475a9-arm64-android.apk)
<details>

- [arm-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3300-ff0d807c46ea5fed3195606c4ca9ef246c2475a9-arm-android.apk)
- [x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3300-ff0d807c46ea5fed3195606c4ca9ef246c2475a9-x64-android.apk)
- [x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3300-ff0d807c46ea5fed3195606c4ca9ef246c2475a9-x86-android.apk)
</details>

test/spix/smoke_test.py
```diff
@@ -27,13 +27,13 @@ def test_start_app(app, screenshot_path, extra, process_alive):
...
     assert app.existsAndVisible("mainWindow/welcomeScreen")
 
-    first_title = app.getStringProperty(
-        "mainWindow/welcomeScreen/loadProjectItem_1", "title"
-    )
-    app.mouseClick("mainWindow/welcomeScreen/loadProjectItem_1")
+    app.setStringProperty("mainWindow", "width", "500")
+    app.setStringProperty("mainWindow", "height", "500")
+
+    time.sleep(2)
```
nirvn, line 35: Sleep for an unreasonable amount of time.

```suggestion
    time.sleep(20)
```

nirvn:
@m-kuhn , this is ready to review. 

For the record, this PR will prevent us from regressing on the WMS/WMTS front without us knowing about it.

test/spix/smoke_test.py
```diff
@@ -26,18 +26,36 @@ def test_start_app(app, screenshot_path, extra, process_alive):
     """
     Starts a test app to the welcome screen and creates a screenshot.
     """
+    if not os.environ.get("SPIX_PROJECT_FILE") is None:
```
m-kuhn, line 29: - This should be added as a new test, next to `test_start_app`
- The filename should be coded in this file (not in an env variable)
- Ideally, we would even [parameterize the test](https://github.com/opengisch/qgis_server_render_geojson/blob/master/tests/test_request.py#L54-L67), so we can run the same test on a matrix of files

m-kuhn:
Good stuff!
I'd like to restructure this so we can start to have a growing set of tests in here

test/spix/smoke_test.py
nirvn: @m-kuhn , +1 to make a new test_message_logs test, that makes a lot of sense.

As for the filename not being an environment variable, I'm open to suggestions but I'm not sure how we'd get the source tree path to direct us to the test/testdata/...qgz to begin with. Also, we'd need to do it in the other file (conftest.py) as we need for it to be loaded via exec argument. Thoughts on ways to proceed are most welcome :)

m-kuhn: The path to this python file could help?
Otherwise we could also just add a `TESTDATA` path as env var.

nirvn: @m-kuhn , path to python file sounds good.

nirvn: @m-kuhn , I've pushed something, let me know what you think of it.

test/spix/smoke_test.py
```diff
@@ -28,23 +28,35 @@ def test_start_app(app, screenshot_path, extra, process_alive):
     """
     assert app.existsAndVisible("mainWindow")
 
-    assert app.existsAndVisible("mainWindow/welcomeScreen")
+    app.setStringProperty("mainWindow", "width", "500")
+    app.setStringProperty("mainWindow", "height", "500")
+
+    time.sleep(2)
```
m-kuhn, line 34: What's that sleep for?

nirvn: I've seen intermittent failures otherwise.

nirvn, line 34:
```suggestion
    time.sleep(10)
```

m-kuhn: Don't remember the reason?

nirvn: It actually had to do with giving QField some time to load the project. I guess I can get rid of this one for the startup test.

test/spix/conftest.py
```diff
@@ -72,10 +75,13 @@ def process():
         "./output/bin/qfield.app/qfield_spix.exe",
         "./output/bin/qfield.app/Contents/MacOS/qfield_spix",
     ]
+
+    projectpath = str(Path(__file__).parent.parent / "testdata" / "test_wms.qgz")
```
m-kuhn, line 79: Could we parametrize this? Something like https://stackoverflow.com/a/33879151/2319028 so we could have an array of projects to test?

test/spix/smoke_test.py
nirvn: (actually no, we'd want to get a screen capture from the loaded project too, so let's keep the sleep time)

m-kuhn: We'll want to connect to a signal when rendering has finished?

nirvn: Eventually, whenever spix supports that :)

m-kuhn: It does not :fearful: ?
Then we can just do a param `finishedRendering` and check if this is set in a loop for now?

m-kuhn (approved):

test/spix/smoke_test.py
nirvn: (Discussed on chat)

PR #3378, opened 2022-09-26
Try enabling tests on Qt6 builds

nirvn:


test/test_deltafilewrapper.cpp
```diff
@@ -970,7 +976,7 @@ TEST_CASE( "Delta File Wrapper" )
             },
             "old": {
               "attributes": {
-                "attachment": null,
+                "attachment": %3,
```
suricactus, line 979: Why is that? I think it is misleading if the value is `""` instead of pure JSON `null`? Probably needs fix upstream or there is some hidden arguments I don't understand.

nirvn: I suspect the delta code will need updating to take this new truth into account:

In Qt5, QVariant(QString()).isNull() is true
In Qt6, that now returns false, because while the qstring is null the qvariant itself isn't

Does that make you think of any part of the delta code? :)

suricactus: hm, it seems is caused from `oldVal.isNull()` not returning true, therefore `QJsonValue::fromVariant( oldVal ) )` is returning `""`.
It is here and https://github.com/opengisch/QField/blob/master/src/core/deltafilewrapper.cpp#L585-L618

nirvn: @suricactus , I suspect we'll have to use QgsVariantUtils::isNull (https://github.com/qgis/QGIS/commit/8171b76ddd64049715bdd3229ceeaf8ac8dd73b9) when we update to QGIS 3.28, that'll be the proper fix.

qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 5abe157eb0b1feb1712f2b567dad38172144a48e: [**arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3378-8d585a9a789d1209097ff01e3dede51d46ba32a5-arm64-android.apk)
<details>

- [arm-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3378-8d585a9a789d1209097ff01e3dede51d46ba32a5-arm-android.apk)
- [x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3378-8d585a9a789d1209097ff01e3dede51d46ba32a5-x64-android.apk)
- [x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-3378-8d585a9a789d1209097ff01e3dede51d46ba32a5-x86-android.apk)
</details>

test/test_deltafilewrapper.cpp
m-kuhn: Qt6 builds already build against qgis 3.28

nirvn: @m-kuhn , true that!

nirvn: @suricactus , @m-kuhn , fix incoming.

suricactus: if you don't mind, I will push a change that covers all the places I used this logic.

nirvn:
Interesting set of observations here. While my Qt6[.3.1] build runs all tests successfully here, the Qt6 OSX build fails _a lot_:
```
	  1 - TestLayerObserver (SEGFAULT)
	  4 - Delta File Wrapper (SEGFAULT)
	  6 - GeometryUtils (SEGFAULT)
	  7 - SgGeometry (SEGFAULT)
	 10 - Digitizing logger (SEGFAULT)
	 11 - Attribute form model (SEGFAULT)
	 12 - qmltest (Failed)
	 13 - smoke_test (Failed)
```

While I'm not exactly sure why half of them SEGFAULTs, it seems the qmltest and smoke_test both fail due to QField likely crashing on launch. We get this bit from the qmltest failure:
```
12/13 Test #12: qmltest ..........................***Failed    1.94 sec
********* Start testing of test_qml_editorwidgets *********
Config: Using QtTest library 6.3.1, Qt 6.3.1 (x86_64-little_endian-lp64 static release build; by Clang 13.0.0 (clang-1300.0.29.30) (Apple)), macos 11.7
QWARN  : test_qml_editorwidgets::tst_editorwidgets::compile() 
  /Users/runner/work/QField/QField/test/qml/tst_editorwidgets.qml produced 2 error(s):
    /Users/runner/work/QField/QField/test/qml/tst_editorwidgets.qml:9,5: Type TextEdit unavailable
    /Users/runner/work/QField/QField/src/qml/editorwidgets/TextEdit.qml:2,1: module "QtQuick.Controls" is not installed
  Working directory: /Users/runner/builddir/test
  Import paths:
    '/Users/runner/work/QField/QField/src/qml/imports'
    '/Users/runner/builddir/output/bin'
    'qrc:/qt-project.org/imports'
  Plugin paths:
    '.'
```

@m-kuhn , that to me looks a lot like some failing qml import module while building in static mode wouldn't you say?

test/test_deltafilewrapper.cpp
nirvn: @suricactus , that'd be fantastic, you're level of intimacy makes you a better committer for this :) you can close the PR I opened.

src/qml/imports/QFieldControls/+Qt6/CalendarPanel.qml
```diff
@@ -27,8 +27,6 @@ Popup {
 
     onAboutToShow: {
```
m-kuhn, line 28: @nirvn what would you suggest doing here, should the `calendar.` lines be commented (so it's easier to revive) or the file minimized to a simple `Item` or similar in order to clearly mark them as "missing implementation". The current state of "80% something" could be misleading?

nirvn: @m-kuhn , TBH, I'd leave it as is, the end result will pretty much match what's in there.

m-kuhn (approved):

src/qml/imports/QFieldControls/+Qt6/CalendarPanel.qml
m-kuhn: I'll trust you on that one

PR #3499, opened 2022-10-24
By the power of javascript, I grant you the latest windows release

nirvn:
@marioba , @m-kuhn , that's a possibility.

### DIFF
--- a/get_latest/index.html
+++ b/get_latest/index.html
@@ -1,0 +1,46 @@
+<!DOCTYPE html>
+<head>
+<title>Get QField</title>
+</head>
+<body>
+<script type="text/javascript">
+const API_LATEST_RELEASE_URL = 'https://api.github.com/repos/opengisch/QField/releases/latest';
+const BROWSER_LATEST_RELEASE_URL = 'https://github.com/opengisch/QField/releases/latest';
+try {
+  const queryParams = new Proxy(new URLSearchParams(window.location.search), {
+    get: (searchParams, prop) => searchParams.get(prop),
+  });
+  if (!queryParams.platform) {
+    throw new Error('No "platform" parameter provided!');
+  }
+  fetch(API_LATEST_RELEASE_URL)
+    .then((response) => {
+      if (!response.ok) throw new Error();
+      return response.json();
+    })
+    .then(json => {
+      var downloadUrl = "https://github.com/opengisch/QField/releases/download/" + json['tag_name'] + "/qfield-" + json['tag_name'] + "-";
+      if (queryParams['platform'] == "windows") {
+        downloadUrl += "windows-x64.exe";
+      } else if (queryParams['platform'] == "linux") {
+        downloadUrl += "linux-x64.AppImage";
+      }
+      if (json['assets'].find(asset => asset["browser_download_url"] === downloadUrl)) {
+        return downloadUrl;
+      } else {
+        return BROWSER_LATEST_RELEASE_URL;
+      }
+    })
+    .then(browserUrl => {
+      window.location.replace(browserUrl);
+    })
+    .catch(error => {
+      window.location.replace(BROWSER_LATEST_RELEASE_URL);
+    });
+} catch (error) {
+  window.location.replace(BROWSER_LATEST_RELEASE_URL);
+}
+</script>
+<p>Redirecting to the latest QField package. If this does not work please go to the <a href='https://github.com/opengisch/QField/releases/latest'>latest release page</a>.</p>
+</body>
+</html>


nirvn:
(we could improve the script to detect whether the browser reports linux/windows/etc. to direct to AppImage et al)

suricactus:
Please note that `gh-pages` always has single commit, so it will be replaced every time after build. There is javascript file that can be configured in (material-?)mkdocs for that purpose.
Also, I am not sure if javascript solution is the best way to go, or the same logic during the build process.
Finally, if we go for javascript solution, I would prefer using the `fetch()` API instead.

nirvn:
@suricactus , I always worry about browser compatibility with fetch. But we're in 2022, that's likely not a problem these days. +1.

I'm not sure I fully understand the remaining points. This would just be a static html page doing a JavaScript redirect.

suricactus:
my bad, I got confused which repository I am on. nvm

nirvn:
@suricactus , and in your opinion, we're getting over 90% browser compatibility when using fetch() in 2022? It's definitively a cleaner approach.

nirvn:
@suricactus , fetch variant implemented.

@marioba , I added the same trick for linux, we'd now have https://qfield.org/get_windows and https://qfield.org/get_linux as hyperlinks pointing to the latest {.exe, .AppImage} release.

get_linux/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<head>
+<title>Get QField for Windows</title>
```
suricactus, line 3: :penguin: :gun:

get_linux/index.html
```diff
@@ -0,0 +1,17 @@
...
+</head>
+<body>
+<script type="text/javascript">
+fetch('https://api.github.com/repos/opengisch/QField/releases/latest')
+.then(response => {
+    if (response.status !== 200) throw new Error();
+    return response.json();
+})
+.then(json => { window.location.replace('https://github.com/opengisch/QField/releases/download/'+json['tag_name']+'/qfield-'+json['tag_name']+'-linux-x64.AppImage') })
+.catch(error => { window.location.replace('https://github.com/opengisch/QField/releases/latest') });
```
suricactus, line 13:
```suggestion
const API_LATEST_RELEASE_URL = 'https://api.github.com/repos/opengisch/QField/releases/latest';
const BROWSER_LATEST_RELEASE_URL = 'https://github.com/opengisch/QField/releases/latest';
const queryParams = new Proxy(new URLSearchParams(window.location.search), {
  get: (searchParams, prop) => searchParams.get(prop),
});

if (queryParams.platform) {
  fetch(API_LATEST_RELEASE_URL)
    .then((response) => {
      if (!response.ok) throw new Error();
      return response.json();
    })
    .then(json => {
      const downloadUrl = `https://github.com/opengisch/QField/releases/download/${json['tag_name']}/qfield-${json['tag_name']}-${queryParams['platform']}`;
      if (json['assets'].find(asset => asset["browser_download_url"] === downloadUrl)) {
        return downloadUrl;
      } else {
        return BROWSER_LATEST_RELEASE_URL;
      }
    })
    .then(browserUrl => {
      window.location.replace(browserUrl);
    });
} else {
  window.location.replace(BROWSER_LATEST_RELEASE_URL);
}
```

suricactus: Making it like this means we can add whatever links we want at any point of time and reduce code duplication.

get_linux/index.html
suricactus: I will keep it "Get QField" only.

get_linux/index.html
```diff
@@ -0,0 +1,17 @@
...
+</script>
+<p>Redirecting to the windows download page. If this does not work please go to the <a href='https://github.com/opengisch/QField/releases/latest'>latest release page</a>.</p>
```
suricactus, line 15:
```suggestion
<p>Redirecting to the ${queryParams.nickname} download page. If this does not work please go to the <a href='https://github.com/opengisch/QField/releases/latest'>latest release page</a>.</p>
```
We can pass the platform name also as a queryParameter. Needs a bit more js to work.

get_linux/index.html
suricactus: Or make it "Get QField for ${queryParams.nickname}".

get_linux/index.html
nirvn: Actually there I'd just no refer to the platform name.

get_linux/index.html
nirvn: +1 to not name the platform.

get_linux/index.html
nirvn: I'm fine with that alternative approach. What would be the URL? qfield.org/get_package/?platform={windows,linux} ?

suricactus: Yup, sounds good.

nirvn:
@suricactus , good comments. The PR now implements the following URLs:

https://qfield.org/get_latest/?platform=windows
https://qfield.org/get_latest/?platform=linux

That'll be super useful for the documentation, as well as in communication with people out there through social media / messenger / etc.

suricactus (approved):
Shall we try it in the wild/

get_latest/index.html
```diff
@@ -0,0 +1,46 @@
...
+      if (!response.ok) throw new Error();
+      return response.json();
+    })
+    .then(json => {
+      var downloadUrl = "https://github.com/opengisch/QField/releases/download/" + json['tag_name'] + "/qfield-" + json['tag_name'] + "-";
+      if (queryParams['platform'] == "windows") {
+        downloadUrl += "windows-x64.exe";
+      } else if (queryParams['platform'] == "linux") {
+        downloadUrl += "linux-x64.AppImage";
+      }
```
suricactus, line 27: Note with this implementation difference it means you have to edit this file any time you want to download a new platform binary. Why don't you pass the whole suffix of the asset name, as suggested in my original comment?

nirvn: @suricactus , I thought qfield.org/get_latest/?platform={windows,linux} were nicer links than qfield.org/get_latest/?platform={windows-x64.exe,linux-x64.AppImage}.

I'd rather have to do one edit whenever we add a new platform (which would be once per few years? :) ) than having longer and slightly more "intimidating" URLs.

suricactus: we experience different laziness levels  :)

suricactus: A bit too late, but just to satisfy my laziness and your less intimidating URLs, we should have added `else { downloadUrl = queryParams['platform'] ;  }`

PR #3510, opened 2022-10-26
Update default values when editing a new feature to match QGIS behavior

### LINKED ISSUE #3509: Default values using other fields in expressions inconsistent behavior
### Main issue
Using expressions referencing other fields in the project, default values are inconsistently filled. All do work within QGIS.

- a string field concatenating geometry X and Y only when a second boolean field is set to true, else is set to NULL. (does not work)
- a datetime field set to now() with the same boolean field condition as the previous. (does not work)
- another string field concatenating gnss_coordinate when a second boolean field is set to true. (does not work)
- another datetime field with the same conditions but referencing the second boolean field (this one sometimes works, it is the last field in the form, may be relevant).

Live update is inconsistent. Depending on the order in which the fields are filled, sometimes the expression succeeds during entity edition for the second datetime field (the last field in the form).

Even if live update doesn't work, when saving the entity, the last datetime field will always have been filled. Not the other three.

This happens on projects packaged through QFieldSync, QGIS projects opened directly in QField, and QFieldCloud projects. Tested on two different devices.

### Reproduction steps
1. Set a layer with conditional expressions using a boolean field true/false state ( if() function ) for default value
2. Send project to QField via any means (direct, QFieldSync, QFieldCloud)
3. Add entity and tick boolean checkboxes. String and datetime fields do not live update, some are filled on saving entity.

### Expected behavior
Either for all default values to work if supported or for all to fail if not.

### Observed behavior
Only certain fields work, and live update is inconsistent. One of two same type fields with the exact same expression writes default value on saving. Not the other.

### Desktop
 - OS: Windows 10
 - QGIS 3.26.3
 - QFieldSync 4.2.0

### Mobile
 - Device 1: Motorola G7 Play
   - OS: Android 10
   - QField 2.4.4 (db90b0)
 - Device 2: Samsung Galaxy Tab A
   - OS: Android 7
   - QField 2.4.4 (db90b0)

### Additional information
* Problem started happening recently, didn't happen in an older version of QField: Unknown
* Problem can be reliably reproduced, doesn't happen randomly: Yes
* Problem happens with all files and projects, not only some files or projects: Yes
* QFieldCloud project _Echantillonnage_ can be checked under user: **gabrielc** 

[Test_data.zip](https://github.com/opengisch/QField/files/9869960/Test_data.zip)

nirvn:
It does make sense to do "live" update of default values when the [ ] apply default value on update checkbox isn't on for new features that haven't been created (i.e. saved) yet while editing the feature form post digitizing.

This fixes #3509 and in doing so harmonizes QField's behavior with that of QGIS.

### DIFF
--- a/src/core/attributeformmodelbase.cpp
+++ b/src/core/attributeformmodelbase.cpp
@@ -494,7 +494,8 @@ void AttributeFormModelBase::updateDefaultValues( int fieldIndex, QVector<int> u
   {
     QStandardItem *item = constraintIterator.key();
     int fidx = item->data( AttributeFormModel::FieldIndex ).toInt();
-    if ( fidx == fieldIndex || !fields.at( fidx ).defaultValueDefinition().isValid() || !fields.at( fidx ).defaultValueDefinition().applyOnUpdate() )
+    const bool isNewFeature = mFeatureModel->feature().id() == FID_NULL;
+    if ( fidx == fieldIndex || !fields.at( fidx ).defaultValueDefinition().isValid() || ( !isNewFeature && !fields.at( fidx ).defaultValueDefinition().applyOnUpdate() ) )
       continue;
 
     QgsExpression exp( fields.at( fidx ).defaultValueDefinition().expression() );


suricactus (approved):
Looking at the your info seems it's quite an edge case, code looks good.

PR #3869, opened 2023-02-01
add windows button

mbernasocchi:


nirvn (approved):
@mbernasocchi , would be great to make use of this image in the documentation site too.

PR #3913, opened 2023-02-16
convert sh to unix files, if needed

tkauppin:
When I cloned the project to Windows, sh files got windows line breaks.
This conversion would revert them back to unix line breaks. Perhaps not the optimal solution but the best one I could find.

### DIFF
--- a/.docker/android_dev/Dockerfile
+++ b/.docker/android_dev/Dockerfile
@@ -58,6 +58,9 @@ RUN groupadd android && useradd -d /opt/android-sdk-linux -g android -u 1000 and
 
 COPY tools /opt/tools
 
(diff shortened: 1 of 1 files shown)


suricactus:
@tkauppin isn't it there this native git for windows funtionality that auto-converts the newlines to OS specific?

m-kuhn:
I think we need this on .sh files: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

tkauppin:
> I think we need this on .sh files: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
> 
> And welcome @tkauppin !

That looks promising.

PR #3966, opened 2023-02-24
Update OverlayFeatureFormDrawer.qml

GeoCode-Dev:
Hi,

I’m often using QField for data collection. I’ve recently stated reviewing the code to suggest some changes. At first, I would suggest to make the drawer taller because it really increase user experience when collecting data. Mostly, it increase readability when forms contains many fields.

Sincerely,

### DIFF
--- a/src/qml/OverlayFeatureFormDrawer.qml
+++ b/src/qml/OverlayFeatureFormDrawer.qml
@@ -28,7 +28,7 @@ Drawer {
      if (qfieldSettings.fullScreenIdentifyView || parent.width > parent.height) {
          parent.height
      } else {
-         Math.min(Math.max( 200, parent.height / 2 ), parent.height)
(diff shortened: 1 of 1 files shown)


nirvn:
@GeoCode-Dev , hey there, glad to welcome a new code provider to this community! :)

On the specific change proposed here, the decision to have the drawer only take half of the screen height when the device is in portrait mode was quite deliberate. It is in part to insure that a selected point will be visible when [x] auto zoom to feature is activated. Changing that would reduce the visibility of features, which isn't great. See e.g. a point:
image

That being said, are you aware that you can drag the feature form green title bar upwards to "fullscreen" the feature form? You can toggle between half and full screen by dragging up and down.

So while I'm going to close this PR, I really hope this will not turn you down, I'm expecting more PRs from you in the near future! :wink:

PR #3968, opened 2023-02-24
Updated 'translations.yaml' to use latest version of Transifex CLI (G…

why-not-try-calmer:
…o binary).

__NB__: Because secrets are obviously not share with forks, and because this PR is coming from my fork, tests are expected to fail. To remedy the situation, a maintainer could hijack the PR.


nirvn (approved):
Let's give this a try.

PR #3970, opened 2023-02-25
Updated resource identifiers in tx config

why-not-try-calmer:
Tying up loose ends of #3968.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -2,13 +2,13 @@
 host = https://www.transifex.com
 lang_map = zh-Hant: zh_TW
 
-[qfield-for-qgis.qfield]
+[o:opengisch:p:qfield-for-qgis:r:qfield]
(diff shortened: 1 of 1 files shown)


nirvn:
@why-not-try-calmer , let me know when you want me to approve and merge, I trust what you're doing :)

why-not-try-calmer:
> @why-not-try-calmer , let me know when you want me to approve and merge, I trust what you're doing :)

It's merge-ready! Sorry it had to be done in two steps... If any third step is required, that means the CLI chokes on pattern-matching the filenames of the translation resources, to the effect that we'll need to "expand" the pattern-matching to concrete file names (it happens sometimes but not always, hence the minimal effort approach here).

nirvn:
Let's do it!

PR #4193, opened 2023-04-19
Fix log directory creation

domi4484:


nirvn (approved):
Good catch, I had checked the presence of folder in my external mem. card.

PR #4226, opened 2023-04-24
Delay creation of the bluetooth discovery agent until we need it

nirvn:
@m-kuhn , -30 seconds


m-kuhn (approved):
Confirmed on OSGeoLive, loading time to 0 seconds

PR #4239, opened 2023-04-27
Update dev.md

GeoCode-Dev:
Adding a full iOS built process to help those who want to custom their QField app on iOS

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -214,20 +214,23 @@ xcode-select --install
 sudo xcode-select --switch /Library/Developer/CommandLineTools
 sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
 
+QT_ROOT=$HOME/Documents/qt5
+
+
 # Then, of course, Qt is needed (Could be installed with 'aqtinstall' or with the official tools)
-pip3 install aqtinstall && aqt install-qt mac ios 5.15.2 -O qt5 -m qtcharts # Qt
+pip3 install aqtinstall && aqt install-qt mac ios 5.15.2 -O $QT_ROOT -m qtcharts # Qt
 
 # Setup the environment for the build tools
 
 export PATH="$(brew --prefix flex)/bin:$PATH" && export PATH="$(brew --prefix bison)/bin:$PATH" # For QGis
 
-export Qt5_DIR=$HOME/Documents/qt5/5.15.2/ios/ # Seems that a full path might be needed for dependencies based on Qt like Qca
+export Qt5_DIR=$QT_ROOT/5.15.2/ios/ # Seems that a full path might be needed for dependencies based on Qt like Qca
 
 # Configure using CMake
 
 cmake -S . -B build-arm64-ios \
-	-DCMAKE_PREFIX_PATH=$HOME/Documents/qt5/5.15.2/ios/lib/cmake/Qt5 \
-	-DCMAKE_FIND_ROOT_PATH=$HOME/Documents/qt5/5.15.2/ios/ \
+	-DCMAKE_PREFIX_PATH=$QT_ROOT/5.15.2/ios/lib/cmake/Qt5 \
+	-DCMAKE_FIND_ROOT_PATH=$QT_ROOT/5.15.2/ios/ \
 	-DSYSTEM_QT=ON \
 	-DVCPKG_TARGET_TRIPLET=arm64-ios \
 	-DWITH_VCPKG=ON \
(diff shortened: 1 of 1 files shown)


doc/dev.md
```diff
@@ -202,6 +202,52 @@ cmake -S . -B build-x64-ios -DVCPKG_TARGET_TRIPLET=x64-ios -GXcode -DWITH_VCPKG=
...
+xcode-select --install
+sudo xcode-select --switch /Library/Developer/CommandLineTools
+sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
+
+# Then, of course, Qt is needed (Could be installed with 'aqtinstall' or with the official tools)
+pip3 install aqtinstall && aqt install-qt mac ios 5.15.2 -O qt5 -m qtcharts # Qt
+
+# Then, really important, some variables need to be exported in the shell
+
+export PATH="$(brew --prefix flex)/bin:$PATH" && export PATH="$(brew --prefix bison)/bin:$PATH" # For QGis
```
m-kuhn, line 222:
```suggestion
export PATH="$(brew --prefix flex)/bin:$PATH" && export PATH="$(brew --prefix bison)/bin:$PATH"
```

doc/dev.md
```diff
@@ -202,6 +202,52 @@ cmake -S . -B build-x64-ios -DVCPKG_TARGET_TRIPLET=x64-ios -GXcode -DWITH_VCPKG=
...
+
+export Qt5_DIR=$HOME/Documents/qt5/5.15.2/ios/ # Seems that a full path might be needed for dependencies based on Qt like Qca
```
m-kuhn, line 224: Can we

- remove the `$HOME` part in here? Especially since the install command on line 218 is used with a relative path. If it requires a full path, we can just use `$(pwd)/qt5/5.15.2` instead
- remove the comment -- or at least the "Seems" and "might be" parts of it?

doc/dev.md
```diff
@@ -202,6 +202,52 @@ cmake -S . -B build-x64-ios -DVCPKG_TARGET_TRIPLET=x64-ios -GXcode -DWITH_VCPKG=
...
+
+# The Cmake configuration command should look like something like this one (The architecture parameter seems to be really important)
```
m-kuhn, line 226:
```suggestion
# Configure using CMake
```

doc/dev.md
```diff
@@ -202,6 +202,52 @@ cmake -S . -B build-x64-ios -DVCPKG_TARGET_TRIPLET=x64-ios -GXcode -DWITH_VCPKG=
...
+
+cmake -S . -B build-arm64-ios \
+	-DCMAKE_PREFIX_PATH=$HOME/Documents/qt5/5.15.2/ios/lib/cmake/Qt5 \
+	-DCMAKE_FIND_ROOT_PATH=$HOME/Documents/qt5/5.15.2/ios/ \
+	-DSYSTEM_QT=ON \
+	-DWITH_QT5=1 \
```
m-kuhn, line 232:
```suggestion
```

Only `BUILD_WITH_QT6` exists

doc/dev.md
```diff
@@ -202,6 +202,52 @@ cmake -S . -B build-x64-ios -DVCPKG_TARGET_TRIPLET=x64-ios -GXcode -DWITH_VCPKG=
...
+	-DVCPKG_TARGET_TRIPLET=arm64-ios \
+	-DWITH_VCPKG=ON \
+	-DVCPKG_BUILD_TYPE=release \
+	-DCMAKE_SYSTEM_NAME=iOS \
+	-DCMAKE_OSX_SYSROOT=iphoneos \
+	-DCMAKE_OSX_ARCHITECTURES=arm64 \
+	-GXcode
+
+# The "build/CMakeFiles/[version]/CMakeSystem.cmake" file may need to be modified.
+# The "CMAKE_SYSTEM_PROCESSOR" variable must be set to "aarch64" manually (only if the cmake configuration fails).
```
m-kuhn, line 242: I wish we could find a better way and avoid this manual patching requirement, but for now let's document it as done here, thanks!

doc/dev.md
```diff
@@ -202,6 +202,52 @@ cmake -S . -B build-x64-ios -DVCPKG_TARGET_TRIPLET=x64-ios -GXcode -DWITH_VCPKG=
...
+
+# In "Xcode" build settings, it is possible that the additional flag "CoreFoundation" had to be removed manually.
+
+# Then, compile. Hopefully, all done. To install an app on iOS, it must be signed using Xcode tools.
```
m-kuhn, line 246:
```suggestion
# Then, compile. To install an app on iOS, it must be signed using Xcode tools.
```

m-kuhn (commented):
Thank you very much, this is very helpful !
I have added a couple of suggestions to make it a bit more "consumer friendly" but it's really just some details.

PR #4308, opened 2023-05-24
Attribute values need to be reset to have their default value calculated by QGIS

nirvn:


m-kuhn (approved):

nirvn:
Thanks. We will want a 2.8.1 ASAP as this 2.8.0 regression broke point tracking with default value expressions

Issue #4625, opened 2023-10-11
Labels font and symbol size have become smaller since latest update


BoswachterMarc:
### Describe the issue
Label fonts and symbol size are reduced since the latest update to QField 3.0.0 Amazonia fb70fb v3.0.0
The labels become unreadable for collaborators.

### Reproduction steps
1. Have QField updated to the latest version
2. Open project
3. Compare with latest version

### Expected behavior
All fonts and symbols have the former size

### Observed behavior
When you zoom in, the former label and symbol size is there again, but after miliseconds the size is reduced again.
The font size in the QGis project is set to 7.5 and 8.5. I enlarged the font size to 9 but the size remains unreadable. I think I should enlarge it to 12 or 13 but then it becomes to big to use in QGis...

### Screenshots and GIFs

![Screenshot_20231011-092710](https://github.com/opengisch/QField/assets/57868975/da2d4816-68af-458f-8b16-e2e40a822194)
![Screenshot_20231011-093832](https://github.com/opengisch/QField/assets/57868975/189d0585-8578-4dc3-90bb-2d094c9f9b07)

### Desktop (please complete the following information)
 - OS: Ubuntu 23.04
 - QGIS Version 3.32.3
 - QFieldSync Version 4.6.0

### Mobile (please complete the following information)
 - Device: Samsung A52, A50 and samsung tablet(s)
 - OS: Android 13
 - QField version: 3.0.0 Amazonia fb70fb v3.0.0
 - 
### Additional information
* Problem started happening recently, didn't happen in an older version of QField: [Yes]
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* Problem happens with all files and projects, not only some files or projects: [Yes]

Apart from this, the new canvas rendering is a very good improvement! It is much more fluent which was a problem for some devices.

nirvn:
@BoswachterMarc , good catch, these APKs should fix this issue: https://github.com/opengisch/QField/pull/4628

BoswachterMarc:
Yes @nirvn, the issue is fixed with the QField Beta 4628. Thank you to everyone involved. Will you implement this quit quickly or do I need to have the Beta version installed on colleagues' devices?

BoswachterMarc:
@nirvn Great, thank you. A colleague informed me that something is also wrong with the measurement tool. I tried it and indeed, there is a small problem. I will make a new issue for that.

PR #4628, opened 2023-10-11
Use screen's physical DPI value again, android regression fixed in Qt 6.5.3

nirvn:
Fixes a 3.0 visual regression reported here (https://github.com/opengisch/QField/issues/4625). We're lucky, Qt 6.5.0 through to 6.5.2 had a regression which was fixed in 6.5.3, woupidou. 

Thanks to the user for reminding me to test this again.


suricactus (approved):
Good catch!

PR #4715, opened 2023-10-31
Enable pdal

m-kuhn:


m-kuhn:
- pdal is only supported on '!(windows & staticcrt)', which does not match x64-windows-static
- hdf5 problems on android

nirvn:
@m-kuhn , rebased and tried again on our current setup, still failing. Shall we close for now?

PR #4736, opened 2023-11-05
Fix AppStream ID

darkdragon-001:
See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html\#tag-id-generic


m-kuhn (approved):

m-kuhn:
Thanks

PR #4807, opened 2023-11-25
Improve scalebar rendering by moving to shapes

nirvn:
Let's put an end of rectangle abuse and move to proper shape-based rendering.

One of the things it fixes is a small visual imperfection our scale bar had for as long as I can remember:

image

The inconsistent top of the tick's outline width is gone:

image

### DIFF
--- a/src/qml/ScaleBar.qml
+++ b/src/qml/ScaleBar.qml
@@ -1,12 +1,17 @@
 import QtQuick 2.14
+import QtQuick.Shapes 1.14
 
 import org.qfield 1.0
 import org.qgis 1.0
 import Theme 1.0
 
 Item {
+  id: scaleBar
+
   property alias mapSettings: measurement.mapSettings
   property double lineWidth: 2
+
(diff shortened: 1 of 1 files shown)


src/qml/ScaleBar.qml
```diff
@@ -1,90 +1,93 @@
...
+      strokeWidth: barLine.strokeWidth + 1.5
+      strokeColor: "#CCFFFFFF"
+      fillColor: "transparent"
+      startX: 0
+      startY: 0
+
+      PathLine {
+        x: 0
+        y: bar.height
+      }
```
suricactus, line 60: I just got used to `startX`, `startY`, `x`, `y` to be on one line. What do we prefer in the end?

suricactus (approved):
Great!

src/qml/ScaleBar.qml
nirvn: @suricactus , I much prefer that yes; let's embrace it!

PR #4816, opened 2023-12-01
Implement expressed driven feature-level locked geometry

nirvn:
This PR implements a new expression driven locked geometry state for individual features within vector layers. This allows for a same layer to hold both features that have their geometry locked while others can still be edited. 

The code also covers moving of a group of selected features. If the group of selected features contains one or more locked features, it will not be possible to move the group. However, if all features are defined by the expression as not locked, group move will be allowed.

It's a really nice addition, makes it possible for administrators of project to define within a single layer features deemed as having a finalized geometry state vs. not.

Within a QFieldCloud environment, it gets even sweeter as cloud variables can drive locked state.

_(Requires this QFieldSync improvement when configuring the project in QGIS: https://github.com/opengisch/qfieldsync/pull/548)_

### DIFF
--- a/src/core/featuremodel.cpp
+++ b/src/core/featuremodel.cpp
@@ -69,6 +69,8 @@ void FeatureModel::setFeature( const QgsFeature &feature )
   mAttributesAllowEdit.clear();
   emit featureChanged();
   endResetModel();
+
+  updateGeometryLocked();
 }
 
 void FeatureModel::setFeatures( const QList<QgsFeature> &features )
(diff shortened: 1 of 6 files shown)


src/core/featuremodel.h
```diff
@@ -289,6 +293,11 @@ class FeatureModel : public QAbstractListModel
     bool startEditing();
     void setLinkedFeatureValues();
     void updateDefaultValues();
+    void updateGeometryLocked();
+
+    bool mGeometryLocked = false;
+    bool mGeometryLockedByDefault = false;
+    QString mGeometryLockedExpression;
```
suricactus, line 300:
```suggestion
    bool mGeometryLocked = false;
    bool mGeometryLockedByDefault = false;
```

Small docs about these two would help future us to understand what is used when without following the logic.

src/core/featuremodel.cpp
```diff
@@ -138,7 +140,24 @@ void FeatureModel::setCurrentLayer( QgsVectorLayer *layer )
       ( *sRememberings )[mLayer].rememberedFeature = mFeature;
       ( *sRememberings )[mLayer].rememberedAttributes.fill( false, layer->fields().size() );
     }
+
+    mGeometryLockedByDefault = mLayer->customProperty( QStringLiteral( "QFieldSync/is_geometry_locked_expression" ), false ).toBool();
+    const bool geometryLockedExpressionActive = mLayer->customProperty( QStringLiteral( "QFieldSync/is_geometry_locked_expression_active" ), false ).toBool();
+    if ( geometryLockedExpressionActive )
```
suricactus, line 146:
```suggestion
    const QString geometryLockedExpressionActive = mLayer->customProperty( QStringLiteral( "QFieldSync/is_geometry_locked_expression_active" ) ).toString();
    if ( !geometryLockedExpressionActive.trimmed().isEmpty() )
```

I think this is a bit more precise what we want?

suricactus (approved):
Looks good! See the comments I put on linqfieldsync and QFieldSync.

src/core/featuremodel.cpp
nirvn: @suricactus , no, in this case the active is a boolean, vs. the expression itself which is a string.

PR #4874, opened 2023-12-20
Use lossless format to get sharper project preview images

domi4484:


nirvn:
@domi4484 , can you add PNG to that condition over here too: https://github.com/opengisch/QField/blob/master/src/core/localfilesmodel.cpp#L234

+1 to sharper welcome screen :)

domi4484:
> can you add PNG to that condition over here too

Done thanks!

src/core/qgismobileapp.cpp
```diff
@@ -667,7 +667,7 @@ void QgisMobileapp::onMapCanvasRefreshed()
   disconnect( mMapCanvas, &QgsQuickMapCanvasMap::mapCanvasRefreshed, this, &QgisMobileapp::onMapCanvasRefreshed );
   if ( !mProjectFilePath.isEmpty() )
   {
-    if ( !QFileInfo::exists( QStringLiteral( "%1.jpg" ).arg( mProjectFilePath ) ) )
+    if ( !QFileInfo::exists( QStringLiteral( "%1.png" ).arg( mProjectFilePath ) ) )
```
nirvn, line 670: @domi4484 , I think we need to keep the .jpg in there, otherwise people will start to see their old preview JPG appear which isn't nice UX.

nirvn: @domi4484 , sorry, my comment was wrong here. Merging.

nirvn (approved):

PR #4922, opened 2024-01-09
Fix postgresql crash when running against openssl 3.2

nirvn:
@m-kuhn , only noticeable regression I've seen from 3.1.5, ~~and thanks Buddha there's a release fixing it already available :)~~ the solution is to downgrade to 3.1.4 for now (see e.g. https://github.com/Homebrew/homebrew-core/issues/155651)

For the record, things like this: https://opengisch.sentry.io/issues/4819317904/?project=6119013&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=6


nirvn:
Alright postgresql 16.1 did *not* fix the issue. As per other projects out there, the lesson learned is to downgrade to 3.1.4 for now.

nirvn:
Note to self: in the near future, add an additional test on the CI to check against a project containing a vector layer provided by a SSL connection to a postgresql server.

nirvn:
Confirmed the crash is gone with APKs here

m-kuhn (approved):

PR #4988, opened 2024-01-27
Remove use of secondary activities for native camera/gallery/file picker on Android

nirvn:
This PR applies a major rework of how we handle camera, gallery, file picking, and file viewing on Android. The aim is to remove our reliance on secondary activities to execute these tasks.

While the primary goal here is to try and improve the stability of the native camera on devices with low memory, it gives us a nice set of bonuses.

For e.g., since we don't launch a dummy activity to call the gallery, QField looks a lot more integrated. Left (before) vs. right (PR):

image

I've also taken the time to clean up parenting a bit across the board.

Edit: in addition to the significant java activity rework, I've taken the time to update some java dependencies, including updating gradle to 8.2.2 and moving from JDK 11 to JDK 17. All done in the hope that this will increase our stability. Nice bonus, it'll insure that we'll stay compatible with quite a few more releases now :)

PR #4995, opened 2024-01-30
Write settings to permanent storage when OS or user force kills app

nirvn:
I was testing https://github.com/opengisch/QField/pull/4988 on iOS yesterday when I realized modified settings were for the most part never saved. That's because on iOS, there's no double back button quitting per say, only force quitting by swiping the app up.

This commit remedies to that situation. It also benefits Android when people force kill app. Simple step to reproduce issue:
- Open QField
- Load bees project
- Add a navigation destination (long press on canvas -> set as destination)
- Suspend QField, then force kill it by showing running apps and swipping QField up into nonexistence)
- Re-open QField, destination gone.

### DIFF
--- a/src/core/navigation.cpp
+++ b/src/core/navigation.cpp
@@ -49,7 +49,6 @@ Navigation::Navigation()
 
 Navigation::~Navigation()
 {
-  mModel->save();
 }
 
 bool Navigation::isActive() const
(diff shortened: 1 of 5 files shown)


suricactus (approved):
Code looks good, good move. On the `firstRunFlag`, +1 on renaming. Just make sure it is properly working, it is a workaround for Linda's bug (tm). Cannot test myself.

m-kuhn:
> it is a workaround for Linda's bug (tm)

For what I know, this bug is no more and was properly fixed (by using a newer SDK for the complete set of dependencies)

suricactus:
I am not aware if we use this flag for something else? If not, it is safe to be removed.

PR #5054, opened 2024-02-25
Enable background pushing of changes to the cloud as well as automated auto-push

nirvn:
First, a glorious image:

image

Now the description :wink: 

This PR improves the delta file wrapper (keeping track of changes done by users) by adding an awareness of a "pushing" state. When in that state, the delta file wrapper will continue collecting changes, but will keep them pending until we either:
- successfully pushed the previous deltas, upon which the deltas collected during the "pushing" state are added to a new deltas "ID"; or
- fail at pushing the previous deltas, upon which the deltas collected during the "pushing" state are added *on top* of the previous deltas.

This means that we have a fully functional QField _while pushing_, insuring that no data addition/editing/deletion is loss while communicating to the cloud server. 

At this stage, this already fixes a known data loss scenario: ongoing tracking would have the changes made while pushing disappear after the deltas were pushed (because until now the code assumed that no activity would happen during the pushing).

Where it gets cool is that the improvements allow us to do background pushing of changes _in an automated and regular interval_, which QField now does if the user toggles that option on (see above screenshot).

The auto-push every 30 minutes setting is not on by default, and is a cloud project-level setting, allowing for users to turn this on for cloud project A but not make use of it for cloud project B.

### DIFF
--- a/src/core/deltafilewrapper.cpp
+++ b/src/core/deltafilewrapper.cpp
@@ -647,103 +647,7 @@ void DeltaFileWrapper::addPatch( const QString &localLayerId, const QString &sou
 
   delta.insert( QStringLiteral( "old" ), oldData );
   delta.insert( QStringLiteral( "new" ), newData );
-
-  mIsDirty = true;
-
-  QMap<QString, int> layerPkDeltaIdx = mLocalPkDeltaIdx.value( localLayerId );
-  QString localPk = delta.value( QStringLiteral( "localPk" ) ).toString();
-
-  qInfo() << "DeltaFileWrapper::addPatch: localPk=" << localPk << " layerPkDeltaIdx=" << layerPkDeltaIdx;
-
-  if ( layerPkDeltaIdx.contains( localPk ) )
-  {
-    int deltaIdx = layerPkDeltaIdx.take( localPk );
-    QJsonObject deltaCreate = mDeltas.at( deltaIdx ).toObject();
-
-    Q_ASSERT( deltaCreate.value( QStringLiteral( "method" ) ).toString() == QStringLiteral( "create" ) );
-
-    QJsonObject newCreate = deltaCreate.value( QStringLiteral( "new" ) ).toObject();
-    QJsonObject attributesCreate = newCreate.value( QStringLiteral( "attributes" ) ).toObject();
-
-    qInfo() << "DeltaFileWrapper::addPatch: replacing an existing create delta: " << deltaCreate << "at" << deltaIdx;
-
-    if ( !newData.value( QStringLiteral( "geometry" ) ).isUndefined() )
-    {
-      newCreate.insert( QStringLiteral( "geometry" ), newData.value( QStringLiteral( "geometry" ) ) );
-    }
-
-    const QStringList attributeNames = tmpNewAttrs.keys();
-    for ( const QString &attributeName : attributeNames )
-    {
-      attributesCreate.insert( attributeName, tmpNewAttrs.value( attributeName ) );
-    }
-
-    newCreate.insert( QStringLiteral( "attributes" ), geometryToJsonValue( newGeom ) );
-    newCreate.insert( QStringLiteral( "attributes" ), attributesCreate );
-    deltaCreate.insert( QStringLiteral( "new" ), newCreate );
-    deltaCreate.insert( QStringLiteral( "sourcePk" ), delta.value( QStringLiteral( "sourcePk" ) ) );
-
-    mDeltas.replace( deltaIdx, deltaCreate );
-
-    qInfo() << "DeltaFileWrapper::addPatch: replaced an existing create delta: " << deltaCreate;
-
-    return;
-  }
-  else
-  {
-    for ( qsizetype i = mDeltas.size() - 1; i >= 0; i-- )
-    {
-      QJsonObject existingDelta = mDeltas[i].toObject();
-      const QString existingLayerId = existingDelta.value( QStringLiteral( "localLayerId" ) ).toString();
-      const QString existingLocalPk = existingDelta.value( QStringLiteral( "localPk" ) ).toString();
-      const QString existingMethod = existingDelta.value( QStringLiteral( "method" ) ).toString();
-      if ( existingLayerId == localLayerId && existingLocalPk == newFeature.attribute( localPkAttrName ) && existingMethod == "patch" )
-      {
-        QJsonObject existingOldData = existingDelta.value( QStringLiteral( "old" ) ).toObject();
-        QJsonObject existingNewData = existingDelta.value( QStringLiteral( "new" ) ).toObject();
-        if ( newData.contains( "geometry" ) )
-        {
-          existingNewData.insert( "geometry", newData.value( QStringLiteral( "geometry" ) ).toString() );
-          if ( !existingOldData.contains( "geometry" ) )
-          {
-            // Previous patch did not contain a geometry change, add old geometry data
-            existingOldData.insert( "geometry", oldData.value( QStringLiteral( "geometry" ) ) );
-          }
-        }
-        const QStringList attributeNames = tmpNewAttrs.keys();
-        if ( !attributeNames.isEmpty() )
-        {
-          QJsonObject existingOldAttributes = existingOldData.value( QStringLiteral( "attributes" ) ).toObject();
-          QJsonObject existingNewAttributes = existingNewData.value( QStringLiteral( "attributes" ) ).toObject();
-          for ( const QString &attributeName : attributeNames )
-          {
-            existingNewAttributes.insert( attributeName, tmpNewAttrs.value( attributeName ) );
-            if ( !existingOldAttributes.contains( attributeName ) )
-            {
-              // Previous patch did not contain this attribute change, add old attribute value
-              existingOldAttributes.insert( attributeName, tmpOldAttrs.value( attributeName ) );
-            }
-          }
-          existingOldData.insert( "attributes", existingOldAttributes );
-          existingNewData.insert( "attributes", existingNewAttributes );
-        }
-        existingDelta.insert( "old", existingOldData );
-        existingDelta.insert( "new", existingNewData );
-
-        mDeltas.replace( i, existingDelta );
-
-        qInfo() << "DeltaFileWrapper::addPatch: replaced an existing patch delta: " << existingDelta;
-
-        return;
-      }
-    }
-
-    mDeltas.append( delta );
-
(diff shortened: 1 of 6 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for acd2c6b4d799912980a4e43f24e0a1363d4d2d5d: [**arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-arm64.apk)
<details>

- [arm-neon-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-arm-neon.apk)
- [x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-x64.apk)
- [x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-x86.apk)
- [**all access arm64-android**](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield_all_access-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-arm64.apk)
- [all access arm-neon-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield_all_access-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-arm-neon.apk)
- [all access x64-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield_all_access-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-x64.apk)
- [all access x86-android](https://sos-ch-dk-2.exo.io/qfieldapks/ci-builds/qfield_all_access-dev-5054-806692f9a8c0ada086fc1db8a2de90efa2d8ff9c-android-x86.apk)
</details>

src/qml/QFieldCloudPopup.qml
```diff
@@ -652,8 +711,6 @@ Popup {
         displayToast(qsTr('Local changes reverted'))
       else
         displayToast(qsTr('Failed to revert changes'), 'error')
-
-      return
```
suricactus, line 656: Sure about this one? There is another `displayToast` 3 lines below after the `if`?

src/core/qfieldcloudprojectsmodel.h
```diff
@@ -429,6 +433,10 @@ class QFieldCloudProjectsModel : public QAbstractListModel
 
         QDateTime lastLocalDataLastUpdatedAt;
         QDateTime lastRefreshedAt;
+
+        bool autoPushEnabled = false;
+        int autoPushDelay = 30; // in minutes
```
suricactus, line 438:
```suggestion
        int autoPushDelayMins = 30;
```
No need for the comment ;)

Btw, I would name it `autoPushIntervalMins`, it's not a delay?

src/core/qfieldcloudprojectsmodel.cpp
```diff
@@ -1940,10 +1948,27 @@ QHash<int, QByteArray> QFieldCloudProjectsModel::roleNames() const
...
   return roles;
 }
 
+void QFieldCloudProjectsModel::projectSetAutoPushEnabled( const QString &projectId, bool enabled )
+{
+  const QModelIndex projectIndex = findProjectIndex( projectId );
+
+  if ( projectIndex.isValid() )
+  {
+    qDebug() << "IN!";
```
suricactus, line 1963: debug :)

src/core/qfieldcloudprojectsmodel.cpp
```diff
@@ -1940,10 +1948,27 @@ QHash<int, QByteArray> QFieldCloudProjectsModel::roleNames() const
...
+    CloudProject *project = mProjects[projectIndex.row()];
+    project->autoPushEnabled = !project->autoPushEnabled;
+    qDebug() << ( project->autoPushEnabled ? "enabled" : "disabled" );
```
suricactus, line 1966: debug :)

src/qml/QFieldCloudPopup.qml
```diff
@@ -413,6 +412,67 @@ Popup {
...
+              onRunningChanged: {
+                if (running && pushButton.enabled) {
+                  var dt = cloudProjectsModel.currentProjectData.LastLocalPushDeltas
+                  if (dt) {
+                    dt = new Date(dt)
+                    var now = new Date()
+                    if ((now - dt) >= interval) {
+                      projectUpload(false)
+                    }
+                  }
```
suricactus, line 464:
```suggestion
                  const dtStr = cloudProjectsModel.currentProjectData.LastLocalPushDeltas

                  if (dtStr) {
                    const dt = new Date(dtStr)
                    const now = new Date()

                    if ((now - dt) >= interval) {
                      projectUpload(false)
                    }
                  }
```
I would avoid type conversions whenever possible, especially if I can use the `const` instead of `let`.

src/core/deltafilewrapper.cpp
```diff
@@ -892,17 +778,187 @@ void DeltaFileWrapper::addCreate( const QString &localLayerId, const QString &so
...
-  mLocalPkDeltaIdx[localLayerId][localPk] = static_cast<int>( mDeltas.count() );
+  appendDelta( delta );
+}
 
-  mDeltas.append( delta );
-  mIsDirty = true;
+void DeltaFileWrapper::appendDelta( const QJsonObject &delta )
+{
+  mergeDelta( delta );
+}
```
suricactus, line 787: Not clear what is the purpose of this method, if it is just a renaming for `mergeDelta`? 

Isn't the goal "always merge deltas unless deltafile wrapper is busy". If so, then `appendDelta` is enough and whether we merge or append to the deltafile is decided within `appendDelta`, not by the caller.

src/core/deltafilewrapper.cpp
```diff
@@ -892,17 +778,187 @@ void DeltaFileWrapper::addCreate( const QString &localLayerId, const QString &so
...
 
-  qInfo() << "DeltaFileWrapper::addCreate: Added a new create delta: " << delta;
+void DeltaFileWrapper::mergeDelta( const QJsonObject &delta )
```
suricactus, line 789: Can we break up this method into three, instead of three gigantic `if` bodies:

```
if (deltamethod == "create" ) {
   appendCreateDelta( delta );
} else if (deltamethod == "patch" ) {
   appendPatchDelta( delta );
} else if (deltamethod == "delete" ) {
   appendDeleteDelta( delta );
} else {
   Q_ASSERT( 0 );
}
```

We can save a bit of nesting and improve readability.

suricactus (commented):
Nice and long requested functionality! Thanks!

Wondering about the UI, did you consider adding this in the QField settings screen? I don't think this is something you want to change every now and then and adding it where it is on the screenshot feel a bit overwhelming to me. Another option is to add a hamurger menu on the screenshoted screen, where one can choose on/off of this option.

Ideally it should be a slider similar to the way the "dim screen timeout" works.

src/core/deltafilewrapper.cpp
nirvn: @suricactus , doh, big part missing here.

```
if ( mIsPushing )
{
  mPendingDeltas << delta;
}
else
{
  mergeDelta( delta );
}
```

I thought it'd be better to regroup that logic into one function instead of having it three times in add{Create,Delete,Patch}().

src/core/deltafilewrapper.cpp
nirvn: @suricactus , we can, but maybe just adding some // Comment with more spacing to identify the sections would be OK? I'm indifferent really :)

nirvn:
@suricactus , thanks for the fast review, I suspected you'd be interested by this one :)

Your comments have all been addressed. As for the UI, I've moved the toggle to sit below the 3 action buttons for now:

image

We can't move it in the QField settings as it is not a app-wide setting, it is bound to the current cloud project. And as for an hamburger menu, I feel at this stage it'd just be burying the feature. IMHO, I think it's worth keeping it exposed this way for now.

src/core/deltafilewrapper.h
```diff
@@ -364,6 +384,23 @@ class DeltaFileWrapper : public QObject
...
+
+
+    /**
+     * Merge the generated \a delta into stored deltas.
+     */
+    void mergeDelta( const QJsonObject &delta );
+
+    void mergeCreateDelta( const QJsonObject &delta );
+    void mergeDeleteDelta( const QJsonObject &delta );
+    void mergePatchDelta( const QJsonObject &delta );
```
suricactus, line 401:
```suggestion

    /**
     * Merge the generated create \a delta into stored deltas. Should only be called from `mergeDelta` method.
     */
    void mergeCreateDelta( const QJsonObject &delta );


    /**
     * Merge the generated delete \a delta into stored deltas. Should only be called from `mergeDelta` method.
     */
    void mergeDeleteDelta( const QJsonObject &delta );


    /**
     * Merge the generated patch \a delta into stored deltas. Should only be called from `mergeDelta` method.
     */
    void mergePatchDelta( const QJsonObject &delta );
```

Just to make everything documented in this class :)

src/core/deltafilewrapper.cpp
```diff
@@ -892,17 +778,213 @@ void DeltaFileWrapper::addCreate( const QString &localLayerId, const QString &so
...
+  }
+  else
+  {
+    mergeDelta( delta );
+  }
+}
+
+void DeltaFileWrapper::mergeCreateDelta( const QJsonObject &delta )
+{
+  const QString localPk = delta.value( QStringLiteral( "localPk" ) ).toString();
```
suricactus, line 798:
```suggestion
  Q_ASSERT( delta.value( QStringLiteral( "method" ) ) == "create" );

  const QString localPk = delta.value( QStringLiteral( "localPk" ) ).toString();
```

Just being paranoid.

suricactus (approved):
Thanks for the fast addressing of the review. Yup, agreed on your UX comment. Now it is much nicer and less error prone for fat finger. As these settings increase, we can introduce the hamburger menu.

Two small comments, all the rest looks very good, thanks!

nirvn:
@suricactus , done.

src/core/deltafilewrapper.cpp
m-kuhn: I don't think we ever build "Debug" builds where this would trigger.
We could consider to define our own `QF_ASSERT` that reports to sentry

PR #5251, opened 2024-05-18
Information drawer functionality

nirvn:
This PR aims at improving both the style as well as the UX of the way we display positioning/navigation/sensor information on top of the main canvas. 

@mohsenD98 , I've given you write access to QField's main repository, I'd recommend you update you branch to use this.

### DIFF
--- a/src/qml/NavigatingDrawer.qml
+++ b/src/qml/NavigatingDrawer.qml
@@ -1,0 +1,123 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+
+import org.qfield 1.0
+import Theme 1.0
+
+Drawer {
+  id: controller
+  height: defaultHeight
+  Overlay.modal: null
+  modal: false
+  width: parent.width
+  edge: Qt.BottomEdge
+  dragMargin: 0
+  closePolicy: Popup.NoAutoClose
+
+  property Navigation navigation
+  property PositioningSettings positioningSettings
+  property real defaultHeight: positioningPreciseView.height + details.height + 32
+  property real positioningPreciseViewHeight
+  property alias positioningPreciseView: positioningPv
+  property bool closeRequested: false
+  property bool shouldBeOpen: false
+  property bool positioningPreciseEnabled: false
+
+  onShouldBeOpenChanged: {
+    if(shouldBeOpen){
+      closeRequested = false;
+      open()
+    }else{
+      closeRequested = true;
+      close()
+    }
+  }
+
+  onClosed: {
+    if (!closeRequested) {
+      // view dragged down but we should bring it back
+      controller.height = details.height + 24
+      controller.open()
+    }
+  }
+
+  background: Rectangle {
+    anchors.fill: parent
+    color: "transparent"
+  }
+
+  Column {
+    spacing: 8
+    width: parent.width
+
+    Rectangle {
+      id: details
+      anchors.horizontalCenter: parent.horizontalCenter
+      width: parent.width - 16
+      height: navigationInformationView.height
+      radius: 8
+      color: "black"
+
+      NavigationInformationView {
+        id: navigationInformationView
+        navigation: controller.navigation
+        width: parent.width
+        anchors.centerIn: parent
+        radius: parent.radius
+        clip: true
+      }
+
+      Rectangle {
+        width: 50
+        height: 2
+        radius: 4
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.top: parent.top
+        anchors.topMargin: 2
+      }
+
+      MouseArea {
+        anchors.fill: details
+
+        property int dragStartY: 0
+
+        onPressed: mouse => {
+          dragStartY = mouse.y
+        }
+        onPositionChanged: mouse => {
+         var deltaY = mouse.y - dragStartY
+         if (deltaY < -details.height / 4){ // && positioningPv.canShow) {
+            controller.height = controller.defaultHeight
+          }
+        }
+      }
+    }
+
+    Rectangle {
+      anchors.horizontalCenter: parent.horizontalCenter
+      width: parent.width - 16
+      height: positioningPv.height
+      radius: 8
+      color: "black"
+
+      PositioningPreciseView {
+        id: positioningPv
+        precision: positioningSettings.preciseViewPrecision
+        width: parent.width
+        height: positioningPreciseViewHeight
+      }
+
+      Rectangle{
+        visible: !positioningPreciseEnabled
+        anchors.fill: positioningPv
+        color: "#cc2e2e2e"
+
+        Text{
+          anchors.centerIn: parent
+          text: "Enable Gps to show!"
+          color: "white"
+        }
+      }
+    }
+  }
+}
--- a/src/qml/NavigatingDrawer.qml
+++ b/src/qml/NavigatingDrawer.qml
@@ -7,68 +7,59 @@ import Theme 1.0
 Drawer {
   id: controller
   width: parent.width
-  height: defaultHeight
+  height: mainContent.height
   Overlay.modal: null
   modal: false
   edge: Qt.BottomEdge
   dragMargin: 0
   closePolicy: Popup.NoAutoClose
 
-  property alias positioningPreciseView: positioningPv
+  property real realtimeHeight: controller.height * controller.position
+  property bool closeRequested: true
+  property bool openRequested: false
+
+  // NavigationInformationView
+  property bool navigationInformationViewEnabled: navigation.isActive
+
+  // PositioningInformationView
   property Navigation navigation
+  property double antennaHeight
+  property bool positioningInformationViewEnabled: false
+
+  // PositioningPreciseView
+  property alias positioningPreciseView: positioningPv
   property PositioningSettings positioningSettings
   property Positioning positionSource
-  property double antennaHeight
-  property real realtimeHeight: controller.height * controller.position
-  property real positioningPreciseViewHeight
-  property real defaultHeight: details.height + positioningIfo.height + positioningPreciseViewRect.height + 24
-  property bool closeRequested: true
-  property bool navigationInformationViewEnabled: navigation.isActive
-  property bool positioningInformationViewEnabled: false
-  property bool shouldOpen: false
   property bool positioningPreciseEnabled: false
+  property real positioningPreciseViewHeight
 
   function resetHeight() {
     let newHeight = 0
-    newHeight += (details.height + 8)
-    newHeight += (positioningIfo.height + 8)
-    newHeight += (positioningPreciseViewRect.height + 8)
+    newHeight += (navigationInformationView.height + 8)
+    newHeight += (positioningInformationView.height + 8)
+    newHeight += (positioningPv.height + 8)
     newHeight += 16
     controller.height = newHeight
   }
 
-  onNavigationInformationViewEnabledChanged: {
-    if (!navigationInformationViewEnabled) {
-      controller.height -= (navigationInformationView.height + 8)
+  function updateDrawerHeight(visibleView, viewHeight){
+    if (!visibleView) {
+      controller.height -= (viewHeight + 8)
       controller.open()
-    } else if (shouldOpen) {
-      controller.height += (navigationInformationView.height + 8)
-      controller.open()
-    }
-  }
-
-  onPositioningInformationViewEnabledChanged: {
-    if (!positioningInformationViewEnabled) {
-      controller.height -= (positioningInformationView.height + 8)
-      controller.open()
-    } else if (shouldOpen) {
-      controller.height += (positioningInformationView.height + 8)
+    } else if (openRequested) {
+      controller.height += (viewHeight + 8)
       controller.open()
     }
   }
 
-  onPositioningPreciseEnabledChanged: {
-    if (!positioningPreciseEnabled) {
-      controller.height -= (positioningPreciseViewHeight + 8)
-      controller.open()
-    } else if (shouldOpen) {
-      controller.height += (positioningPreciseViewHeight + 8)
-      controller.open()
-    }
-  }
+  onNavigationInformationViewEnabledChanged: updateDrawerHeight (navigationInformationViewEnabled, navigationInformationView.contentHeight)
 
-  onShouldOpenChanged: {
-    if (shouldOpen) {
+  onPositioningInformationViewEnabledChanged: updateDrawerHeight (positioningInformationViewEnabled, positioningInformationView.contentHeight)
+
+  onPositioningPreciseEnabledChanged: updateDrawerHeight (positioningPreciseEnabled, positioningPreciseViewHeight)
+
+  onOpenRequestedChanged: {
+    if (openRequested) {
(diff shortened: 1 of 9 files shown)


src/qml/PositioningInformationView.qml
```diff
@@ -19,15 +19,15 @@ Rectangle {
...
   property color alternateBackgroundColor: Theme.positionBackgroundColor
   property color textColor: positionSource.currentness ? Theme.mainTextColor : Theme.secondaryTextColor
 
-  height: grid.rows * positioningInformationView.rowHeight
+  property real contentHeight: grid.rows * positioningInformationView.rowHeight
   width: parent.width
   anchors.margins: 20
 
-  color: Theme.mainBackgroundColor
+  color: "#dd" + Theme.mainBackgroundColor.toString().slice(1)
```
nirvn, line 30: I think we'd be better of adding a Theme.mainBackgroundColorSemiOpaque instead of this more obscure (and prone to be inconsistent) method.

src/qml/PositioningPreciseView.qml
```diff
@@ -31,7 +31,8 @@ Item {
 
   Rectangle {
     anchors.fill: parent
-    color: Theme.mainBackgroundColor
+    color: "#ee" + Theme.mainBackgroundColor.toString().slice(1)
```
nirvn, line 34: Same as above.

src/qml/qgismobileapp.qml
```diff
@@ -931,16 +931,35 @@ ApplicationWindow {
     }
   }
 
+  NavigatingDrawer{
```
nirvn, line 934: I think we need a more generic name here. I'd name this item `InformationDrawer` instead because not only do we have positioning details (which aren't navigation-centric when you don't see a navigation destination), but we also have sensor data (more on that below ;-) )

src/qml/qgismobileapp.qml
```diff
@@ -931,16 +931,35 @@ ApplicationWindow {
...
+    positioningPreciseEnabled: !elevationProfile.visible
+                               && !isNaN(navigation.distance)
+                               && navigation.isActive
+                               && (positioningSettings.alwaysShowPreciseView
+                                  || ( positioningPreciseView.hasAcceptableAccuracy
+                                  &&  positioningPreciseView.projectDistance < positioningPreciseView.precision ))
+
+    positioningInformationViewEnabled: positioningSettings.showPositionInformation && !elevationProfile.visible
+    positionSource: positionSource
+    antennaHeight: positioningSettings.antennaHeightActivated ? positioningSettings.antennaHeight : NaN
```
nirvn, line 953: Could we move that logic into the ~~Navigating~~InformationDrawer item itself? Less code in qgismobileapp.qml is better.

On that note, very happy to see you've moved individual information items into the drawer QML file, shrinks our mammoth main qml file a bit :)

src/qml/qgismobileapp.qml
nirvn, line 934: (also, don't forget the space before the {)

```suggestion
  InformationDrawer {
```

src/qml/qgismobileapp.qml
```diff
@@ -1124,33 +1095,26 @@ ApplicationWindow {
 
   QfToolButton {
     id: compassArrow
+    y: navigatingDrawer.openRequested ? ( parent.height - height ) - (navigatingDrawer.realtimeHeight) - 44 :
```
nirvn, line 1098: Hm, I'm not a big fan of setting the ys on all of these items. 

I think we should go for what we discussed over on google chat here, which is an item that you would use as parent to anchor these items.

E.g.

```
Item {
  id: mapOverlays
  width: mapCanvas.width
  height: informationDrawer.openRequested ? (mapCanvas.height - informationDrawer.realtimeHeight) : mapCanvas.height

  QfToolButton {
    id: compassArrow
    ...
    anchors.bottom: parent.bottom
     ...
  }
}
```

That way you avoid complicated logic for a bunch of items by handling height in a single parent item.

That comment is valid for every item you've added a y property to here.

qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for f016e81a7675ef9794bba6b9f6060dc3ca64e649: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5251-67412a55fe24ae84214295ee79cf20ec31876c3b-android-x86.apk)
</details>

src/qml/NavigatingDrawer.qml
```diff
@@ -0,0 +1,125 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+
+import org.qfield 1.0
+import Theme 1.0
+
+Drawer {
```
nirvn, line 7: How does this Drawer interact with the dashboard (which is also a drawer)? Are they stacking properly?

src/qml/NavigatingDrawer.qml
```diff
@@ -0,0 +1,125 @@
...
+      anchors.horizontalCenter: parent.horizontalCenter
+      clip: true
+      visible: positioningInformationViewEnabled
+      positionSource: controller.positionSource
+      antennaHeight: positioningSettings.antennaHeightActivated ? positioningSettings.antennaHeight : NaN
+      radius: 8
+    }
+
+    PositioningPreciseView {
+      id: positioningPv
```
nirvn, line 117: Let's use the full name here for the id, no obscure PV acronym :)

src/qml/NavigatingDrawer.qml
```diff
@@ -0,0 +1,125 @@
...
+    if (!visibleView) {
+      controller.height -= (viewHeight + 8)
+      controller.open()
+    } else if (openRequested) {
+      controller.height += (viewHeight + 8)
+      controller.open()
+    }
+  }
+
+  onNavigationInformationViewEnabledChanged: updateDrawerHeight (navigationInformationViewEnabled, navigationInformationView.contentHeight)
```
nirvn, line 55: To be consistent, let's avoid spaces between the function name and the (

```suggestion
  onNavigationInformationViewEnabledChanged: updateDrawerHeight(navigationInformationViewEnabled, navigationInformationView.contentHeight)
```

src/qml/NavigatingDrawer.qml
```diff
@@ -0,0 +1,125 @@
...
+  onOpenRequestedChanged: {
+    if (openRequested) {
+      closeRequested = false
+      resetHeight()
+      open()
+    } else {
+      closeRequested = true
+      close()
+    }
+  }
```
nirvn, line 70: This confuses me a bit, why do we need the complicated pair of bool property logic here? What QMl shortcoming are we fighting against?

src/qml/NavigatingDrawer.qml
```diff
@@ -0,0 +1,125 @@
...
+    width: parent.width
+
+    NavigationInformationView {
+      id: navigationInformationView
+      width: parent.width - 8
+      height: navigationInformationViewEnabled ? contentHeight : 0
+      anchors.horizontalCenter: parent.horizontalCenter
+      clip: true
+      navigation: controller.navigation
+      radius: 8
```
nirvn, line 101: Let's make the radius value a `property real itemRadius: 8` of InformationDrawer, that way we can easily tweak the radius value in one place to impact all items.

src/qml/qgismobileapp.qml
```diff
@@ -957,54 +976,6 @@ ApplicationWindow {
...
-    }
-
-    Rectangle {
-      visible: positioningInformationView.visible && sensorInformationView.visible
-      width: parent.width
-      height: 1
-      color: Theme.sensorBackgroundColor
-    }
-
     SensorInformationView {
```
nirvn, line 979: FYI, this also needs to move into the InformationDrawer.

nirvn (commented):
@mohsenD98 , great work here, super excited to see this coming along! Some comments added.

mohsenD98:
> @mohsenD98 , great work here, super excited to see this coming along! Some comments added.

I'm thrilled to have been able to help and can't wait to dive into the next steps!

Thank you for your kind words and for taking the time to add some comments. I'm always eager to learn from your feedback and make adjustments to ensure that the information is accurate, clear, and helpful.

src/qml/PositioningInformationView.qml
mohsenD98: I was initially hesitant to modify the `Theme.qml` file, but since I've been granted permission, I'll to make the necessary changes.

mohsenD98: any suggestion for name? `mainBackgroundColorTransparent` ? `transparentMainBackground` ? What do you think?

src/qml/qgismobileapp.qml
mohsenD98: it makes sense :+1:

src/qml/qgismobileapp.qml
mohsenD98: Thank you for your kind words ♡

actually 'elevationProfile.visible'  is a bit annoying. i can pass it as a property to "InformationDrawer" but I'm struggling to come up with a suitable property name

src/qml/qgismobileapp.qml
mohsenD98: sure, additionally, it would be great if we could enable auto-formatting. As it stands, I've had to disable it to avoid interfering with other parts of the project.

src/qml/qgismobileapp.qml
mohsenD98: I completely understand your point about not using the y property. In fact, I never use it myself. However, in this particular case, I had to attach elements to the Drawer and unfortunately, I couldn't find a suitable way to attach the edge of the drawer without relying on the y property. ( i'll search more)

src/qml/NavigatingDrawer.qml
mohsenD98: `InformationDrawer` is depend on `mapCanvasMap.isEnabled` there for when `dashboard` is open `InformationDrawer` is hidden and after closing `dashboard`, `InformationDrawer` reappears.

src/qml/NavigatingDrawer.qml
mohsenD98: sure :D

src/qml/NavigatingDrawer.qml
mohsenD98: sure

src/qml/NavigatingDrawer.qml
mohsenD98: When the drawer is in the "Up" position, it becomes interactive and allows the user to push it down. To determine whether the user is pushing the drawer down or we are closing the drawer.

At first i was thinking we should make drawer in multiple states. `closed`, `semi open`, `expanded`. maybe user wants to collapse and see map for a short time. 

But then i removed `semi open` state to make sure every thing works as it must. 

I kept this to bring back pushed down drawer. i can make its `interactive = false` then user cannot touch it and then we can remove this lines.

src/qml/NavigatingDrawer.qml
mohsenD98: sure.

src/qml/NavigatingDrawer.qml
nirvn: I'd rather have a cleaner set of states than a pair of booleans TBH, it'd make it easier to read/understand.

src/qml/PositioningInformationView.qml
nirvn: mainBackgroundColorSemiOpaque to match other color names please

src/qml/qgismobileapp.qml
nirvn: @mohsenD98 , enabling auto formating is on the wishlist :) we tried in the past but it had deficiencies. That was before Qt 6, it's probably much better now. A nice mission for another day :)

src/qml/qgismobileapp.qml
mohsenD98: what about `uiConflictFree: !elevationProfile.visible` ?

src/qml/NavigatingDrawer.qml
mohsenD98: @nirvn should i remove interaction ability? 

because right now we don't support multiple state in it. also it needs new UX definitions. for example what should we show in collapsed mode(when different elements are showing) or what should we show in expanded mode? 

if i remove the interaction this complexity will be removed.

src/qml/qgismobileapp.qml
nirvn: @mohsenD98 , let's just _"cheat"_ for now and just use the globally referenced elevationProfile item from within InformationDrawer. I would like to merge this and work on smaller follow up PRs.

src/qml/qgismobileapp.qml
nirvn: (executed in commit pushed today)

src/qml/NavigatingDrawer.qml
nirvn: @mohsenD98 , yes, let's remove for now.

src/qml/NavigatingDrawer.qml
nirvn: @mohsenD98 , that's not a good variable to use here, we still want info drawer when e.g. we lock the screen. 

For now, let's just have both drawers show, and insure that the dashboard sits on top.

nirvn (commented):
This looks good to me now. Let's merge this when the CI is green and work on follow up tweaks in separate PRs to avoid too many line changes :)

PR #5280, opened 2024-05-30
Value map toggle buttons functionality follow ups

nirvn:
@mohsenD98 , while testing feature attribute transfer functionality, I stumbled on an issue with the value map toggle buttons whereas once you manually click on a button, changes coming from the model are ignored. That's fixed.

I've tweaked the looks / spacing / colors a bit too to harmonize it with other pre-existing editor widgets.


mohsenD98 (approved):
Thanks for fix.

PR #5316, opened 2024-06-07
Update CMakeLists.txt

GeoCode-Dev:
I was trying to compile QField for Android using the Docker image. The whole process of dependencies built went well. Then, when configuring the project, CMake failed. It was related to the QT_HOST_PATH_CMAKE_DIR variable not being found by QtPublicDependencyHelpers.cmake.

By commenting the line where we manually set the variable, the process successfully finished.

Could it be that some change in Qt 6.7 fixed the previous bug of not finding the path? Now that it's fixed, CMake finds it automatically, so we shouldn't need to set the parameter manually…


m-kuhn:
The docker build script is missing `VCPKG_HOST_TRIPLET` as specified here https://github.com/opengisch/QField/blob/master/.github/workflows/android.yml#L139 we forgot to update that

m-kuhn:
Thanks let's give a complete removal a try, comprehensive approach in #5317

PR #5343, opened 2024-06-17
Instead of reloading a whole project, reset data sources

nirvn:
ATM, we force reload a whole project to apply credentials for as few as a single layer. Let's be more surgical about things :) 

@m-kuhn , keen to hear what you think of this alternative approach.

### DIFF
--- a/src/core/qfieldappauthrequesthandler.cpp
+++ b/src/core/qfieldappauthrequesthandler.cpp
@@ -21,7 +21,9 @@
 #include <QThread>
 #include <qgscredentials.h>
 #include <qgsdatasourceuri.h>
+#include <qgsmaplayer.h>
 #include <qgsmessagelog.h>
+#include <qgsproject.h>
 
 QFieldAppAuthRequestHandler::QFieldAppAuthRequestHandler()
 {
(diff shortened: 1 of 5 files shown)


src/core/qfieldappauthrequesthandler.cpp
```diff
@@ -83,7 +85,15 @@ bool QFieldAppAuthRequestHandler::handleLayerLogins()
       }
       else
       {
-        emit reloadEverything();
+        const QList<QgsMapLayer *> mapLayers = QgsProject::instance()->mapLayers().values();
```
m-kuhn, line 88: Can we inject the `QgsProject` instead (as a property)?
It's already here: https://github.com/opengisch/QField/blob/master/src/core/qgismobileapp.cpp#L264 and the appauthrequesthandler is created in the same source file, so could easily be injected.

src/core/qfieldappauthrequesthandler.cpp
```diff
@@ -83,7 +85,15 @@ bool QFieldAppAuthRequestHandler::handleLayerLogins()
...
+        for ( QgsMapLayer *mapLayer : mapLayers )
+        {
+          if ( !mapLayer->isValid() && mapLayer->dataProvider() )
+          {
+            mapLayer->setDataSource( mapLayer->source(), mapLayer->name(), mapLayer->dataProvider()->name() );
```
m-kuhn, line 93: If I'm not mistaken, that forces a reload of the datasource?
If that's the goal of this line, a comment would be justified, that's not immediately clear for a reader.

m-kuhn:
A good thing 👍
Hard to tell the risks of it, needs proper testing

Issue #4512, opened 2023-08-13
Always open child form in the first container


Ipickedausername:
Hello, 

When a child feature is added, the displayed container is always the last one that was open. 
If there is 4 containers and you save the feature while you were on the third one, then when you add a second  child it will open on the third container. 

I suppose it should always show the first container (I didn't check what happen if the third container visibility depends on a value from the first container, which one is missing when we add a child). 

Here is a video that shows this behavior:

https://github.com/opengisch/QField/assets/72577399/16d67513-8ab7-42d2-a0ea-61de438346af

### Desktop (please complete the following information)
 - OS: [e.g. Windows 10]
 - QGIS Version [ 3.26.0]
 - QFieldSync Version [.4.a.0]

### Mobile (please complete the following information)
 - Device: [samsung Galaxy 9]
 - OS: [e.g. Android 8]
 - QField version: qgield master-dev af01c8
 
* Problem started happening recently, didn't happen in an older version of QField: [Yes]
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* Problem happens with all files and projects, not only some files or projects: [Yes]

Thank you !

she-weeds:
Just FYI your post shows only a thumbnail jpg, not a video.

Ipickedausername:
Post edited. Thank you :)

nirvn:
@Ipickedausername , could you attach a small test project for this? Thanks

suricactus:
@Ipickedausername thanks for backreferencing! Should be fixed by #5436.

suricactus:
Ah, if I prematurely closed the issue, please reopen, sorry for the noise.

PR #5436, opened 2024-07-10
Rest feature form tabs on confirmed or cancelled.

mohsenD98:
This PR tries to resolve this issue: 

**Steps to remake issue:**
- Select the apiary layer in the legend, then insure that editing is ON
- Add an apiary point onto the map (fill in the required info), and make sure you click on the OK check mark button when - you are in the 2nd or 3rd tab.
- Then when the form is gone, add another apiary point.
- You'll see the feature form opens at the tab you last  had focused.
- What we should do here is that when we create a new feature, it should reset to the 1st tab.

**Fix:**
we add a reset tab function in `FeatureForm.qml` and after `confirm` or `cancel` we call that.  

Closes #5417

### DIFF
--- a/src/qml/FeatureForm.qml
+++ b/src/qml/FeatureForm.qml
@@ -46,6 +46,10 @@ Page {
     }
   }
 
+  function resetTabs() {
+    tabRow.currentIndex = 0;
+  }
+
   clip: true
   states: [
     State {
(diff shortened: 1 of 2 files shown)


nirvn (approved):
@mohsenD98 , this fixes an issue right?

mohsenD98:
> @mohsenD98 , this fixes an issue right?

Yes [this one](https://github.com/opengisch/QField/discussions/5417)

suricactus:
@mohsenD98 if you write "fixes #5417" or "fix #5417" in the opening message of the PR, github will automatically close the ticket if it is an issue. Not so sure about discussions, but let's see.

suricactus (approved):
Looks good to me.

Ipickedausername:
This PR must also fixes the issue I raised a while ago : #4512

PR #5479, opened 2024-07-26
Add positioning test for featureModel.

mohsenD98:
Using `nmea`, we need to verify that the `positionInformation` in `featureModel` is assigned with values and is **not** `null`.


nirvn:
@mohsenD98 , we're 20% to where we want to be :wink:

nirvn (approved):

PR #5582, opened 2024-08-21
Use the new QfSearchBar item in more places

nirvn:
Less code to maintain, and higher chance of the search bar getting more love across the board. Win-win :)


mohsenD98 (approved):
Grate work :+1:

suricactus (approved):
Lovely!

Issue #5576, opened 2024-08-19
USB connection...


bettellam:
Today I was "playing" with Qfield windows.

I usually use bluetooth connection to connect the GNSS receiver and I don't encounter any particular problems, everything OK....
... at a certain point I wanted to test USB connection...
... try, and try again no result.... Qfield does not receive position.

After two hours of attempts and checks with other programs I thought of checking the default configuration of the COM port in windows:
baudrate was set to 9600, but my GNSS receiver is configured to 115200.
![image](https://github.com/user-attachments/assets/65b95ec2-db09-4318-8a25-666230fb6dc8)

**With this simple change magically QField started to work, everything perfect
I hope this solution can be useful to other users.**

I think it is useful to **implement in QField a combobox** where you can choose the connection baudrate:
https://github.com/opengisch/QField/blob/1dde789af9ed962364b99e7b770a3bd8e554a82a/src/core/positioning/serialportreceiver.cpp#L57
SWMaps:
![image](https://github.com/user-attachments/assets/92bbe929-deda-4a2d-9f6b-c75b317b80d1)

Here .... I wonder:
"is this one of the reasons why QField, with an android device, does not connect to an external GNSS receiver connected with USB? "
... this regardless of the access permissions to the smartphone's USB port (`android.hardware.usb`)... https://github.com/opengisch/QField/blob/80fe9af7b97a2a32c869172d9f1fef4952b2097a/platform/android/AndroidManifest.xml.in#L31

m-kuhn:
@bettellam could you provide a pull request with a patch to test this?
Even if it's without a UI yet, it would be good to validate if the assumption on Android side holds true.

bettellam:
@m-kuhn Thank you for your trust, but my programming skills are very poor,
it's not my job .. at my age I should just be a grandfather... ;-)
(I read code already written and I can interpret it).
Anyway I'll try.

nirvn:
@bettellam , is this still a need?

bettellam:
@nirvn These were tests I ran with Windows, and with the correct configuration, they resolved the issue. I think it can be closed permanently. No other users have reported any problems with this topic.

Issue #4329, opened 2023-06-01
Android USB GPS not connecting - permissions


erik3dg:
### Describe the issue
When attempting to use an external GPS via USB-Serial connection, the serial device is detected but QField won't connect to it because it has "insufficient permissions".

### Reproduction steps
1. Go to 'Settings - Positioning'
2. Add Device - Serial port (NMEA)
3. Tap 'Connect to USB gps'
4. Connection error is now displayed & permission to use device is not asked for

### Expected behavior
QFIeld should ask for permission to use the USB device and then read in the external GPS NMEA position information

### Mobile (please complete the following information)
 - Device: Samsung S21
 - OS: Android 13
 - QField version: 2.8.1

### Additional information
* Problem started happening recently, didn't happen in an older version of QField: [Yes/No]
* Problem can be reliably reproduced, doesn't happen randomly: [Yes/No]
* Problem happens with all files and projects, not only some files or projects: [Yes/No]

[If the problem happened with QFieldCloud, please add your username and project name.]

m-kuhn:
Serial Port was mainly added for Windows, it was never tested with USB on Android. It seems a specific dialog prompting for permissions needs to be implemented. Work in this area is very welcome.

bettellam:
@erik3dg , @m-kuhn 
Hi...
It can be solved using the TCPUART app on android, which takes care of sending the USB_UART NMEA messages to a TCP port (in my case default: 10.1.10.1:8080)
In QField connection type choose: TCP (NMEA) with same port indicated by TCPUART
Buon lavoro...
Ciao :-)

![Screenshot_20230702-154411_TCPUART](https://github.com/opengisch/QField/assets/23143342/ccc5d9c0-982b-441d-a10e-b1549809d4ea)

![Screenshot_20230702-154621](https://github.com/opengisch/QField/assets/23143342/311f16c4-c788-4d52-8c50-6cd0cb623fc8)

![Screenshot_20230702-154545](https://github.com/opengisch/QField/assets/23143342/7a7ba3a7-9e62-4fbe-97dd-b1add19419b8)
![Screenshot_20230702-154527_TCPUART](https://github.com/opengisch/QField/assets/23143342/b377291c-2990-4cd2-bb80-d75dc4b4ad9a)

ruocalgis:
[...]

PR #5618, opened 2024-09-06
Android USB update

bettellam:


nirvn:
@bettellam , hey there, nice to see you opening a pull request over here, hoping this will be one of more to come :) 

If you are interested, I can guide you through adding the UI to select bault rate if this test turns out to fix your issue.

bettellam:
@nirvn Thanks
All this comes from  #5576 
Here I am like "a fish out of water". As I already wrote to @m-kuhn ,
I would never be able to complete something like this, I am not a programmer, I barely read code...
but it seemed like an interesting thing to investigate....

m-kuhn:
@bettellam as soon as the workflows finished, you should get apks in here https://github.com/opengisch/QField/actions/runs/10750024510?pr=5618
There will still be a failure indicated, but let's ignore that for now.

Let's see where we go with this after the first test results.

bettellam:
@m-kuhn Ok. I'm home this afternoon. I'll do a test right away. Thanks

bettellam:
@m-kuhn Test with the window version:
I reset the default setting of the COM1 port to 9600 baud rate, leaving the receiver set to 115200.
image


Started QField and it connects. !!!!!
image



This means that having changed
`mSocket->setBaudRate( QSerialPort::Baud115200 );`
is a correct choice.
**It will be useful to create a combobox where the user chooses the connection baud rate to the USB port.**

As for android I can't find the apr file.

I think that the `android.hardware.usb` alone is not enough because there are all the **authorization problems for the use of the USB!**
Here's what happens with SWMaps:
image

m-kuhn:
If you scroll down here https://github.com/opengisch/QField/actions/runs/10750305033 can you see the apk files?

bettellam:
I don't understand what the file is?
It seems to me that it is not there......

m-kuhn:
image

You should be able to download the [qfield_dev-android-arm64-android](https://github.com/opengisch/QField/actions/runs/10750305033/artifacts/1904277548) file and install it on your phone/tablet for testing. You might have to uninstall a potentially already installed beta version of qfield before (no conflict with the stable installed version)

bettellam:
Nothing... it doesn't install... I tried with 2 devices (android 8 and android10)

m-kuhn:
Do you have any error message?

bettellam:
Android 8

https://github.com/user-attachments/assets/c640e74d-c22f-4b3d-b1e9-97f699f3c8c4

nirvn:
@bettellam , do you have a previous QField 'beta' (i.e. an APK from a PR) installed on your phone?

bettellam:
> @bettellam , do you have a previous QField 'beta' (i.e. an APK from a PR) installed on your phone?

@m-kuhn Qfield dev beta uninstalled!
Standard Qfiled installed.

nirvn:
@bettellam , alright, new APKs have been generated in the latest Android workflow runs and you'll be able to install those. Let us know :)

bettellam:
@nirvn  Installation OK
There is nothing in the listbox that refers to my connected USB device.
The USB port access permissions are missing!
(a system form should open with which the user enables USB access)

https://github.com/user-attachments/assets/26f9c6dd-50c6-46fc-8427-d3b17f8e6a2e

m-kuhn:
I think you need to programmatically ask for these permissions (similar to here https://github.com/opengisch/QField/blob/5c5273e29b456bfe68df5a4418e499c3e240c8c4/src/core/platforms/android/androidplatformutilities.cpp#L678 )

nirvn:
@bettellam , have your been able to investigate this further?

bettellam:
@nirvn 
No, I haven't done anything anymore.
Also because I changed phone...
android14 ....
a disaster ...
it doesn't recognize my bluetooth devices anymore...
it's no longer possible to write in the folder
android/data/ch.opengis.qfield

Pravicy and Security are blocking everything!

nirvn:
@bettellam , that sounds dramatic.

> it's no longer possible to write in the folder android/data/ch.opengis.qfield

You mean you can't copy files into that folder via USB cable?

bettellam:
@nirvn Exactly!

with this application you can copy files to the qfield folder:
https://play.google.com/store/search?q=fv+file+manager&c=apps

nirvn:
@bettellam , I don't think Android 14 is the cause of USB cable transfer not working anymore. I have a pixel 6a with Android 14 and it copies files into that folder without a problem.

bettellam:
@nirvn my realme gt6 I did a test with USB cable from PC and it works.
Thanks
I was used to transfer files with android app in wifi
I will investigate...

bettellam:
@nirvn Back to the USB issue
you could try with
android.hardware.usb.host
You need approval to continue....

nirvn:
@bettellam , are you still interested in the baud rate control here? If so, I think we could guide you into coding it :)

bettellam:
@nirvn Thanks for the trust... but I don't have specific technical knowledge in programming to deal with this problem. If I find some solution I will not fail to report it!

bettellam:
@nirvn , The alternative solution to connect a GNSS receiver with USB port in Android is to use the TCPUART app
https://play.google.com/store/apps/details?id=com.hardcodedjoy.tcpuart
, I reported it in this:
 https://github.com/opengisch/QField/issues/4329

PR #5627, opened 2024-09-08
Skip comment when CI not running from opengisch repository

nirvn:
Less red.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -13,6 +13,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+env:
+  SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+
 jobs:
   build:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


m-kuhn:
We might want to do [what QGIS does for this](https://github.com/qgis/QGIS/blob/master/.github/workflows/build_artifact_comment.yml), it's not yet as flexible and portable as I'd like it to be, with a bit of luck I get around to implementing something better.

nirvn:
@m-kuhn , probably still worth turning red into green for the immediate future, then eventually replace. 

The PR also includes two additional fixes to shut down daily action failures (cron jobs) on 3rd party repositories. IMHO pretty important as it otherwise means anybody forking QField ATM gets notification spammed :)

m-kuhn:
> @m-kuhn , probably still worth turning red into green for the immediate future, then eventually replace.

:+1: 

> The PR also includes additional fixes to shut down daily action failures (cron jobs) on 3rd party repositories. IMHO pretty important as it otherwise means anybody forking QField ATM gets notification spammed :)

I was under the impression, that 3rd party repos need to opt in to enable workflows on forks, but if you say so

m-kuhn (approved):

nirvn:
@m-kuhn , maybe it's because of my user status within the main repo? In any case, I didn't enable anything and I got a bunch of daily notifications after pulling my fork.

PR #5628, opened 2024-09-08
Could we get to green macos on forked repo?

nirvn:
Proof of work.


m-kuhn (approved):
I would already merge that as a first improvement

PR #5637, opened 2024-09-10
Add egeniouss receiver.

mohsenD98:
# Enable Egeniouss Receiver Support

### Overview

This pull request implements support for Egeniouss receivers in the QField application. A new class, `EgenioussReceiver`, allows users to establish a TCP connection and interact with Egeniouss devices.

--- 

### New Class: EgenioussReceiver

The EgenioussReceiver class, derived from AbstractGnssReceiver, has been created. It facilitates a `TCP` connection to Egeniouss devices, using a fixed host address (`localhost`) and port (`1235`).

---

### Activation Instructions

The Egeniouss receiver is automatically enabled upon adding the Egeniouss plugin.

---

### Additional Information

Within the plugin, we have two key functions: `appWideEnabled()` and `appWideDisabled()`. These functions are invoked in QField when the plugin is enabled or disabled. They simply toggle the `egenioussEnabled` property as follows:

```C++
plugin.positioningSettings.egenioussEnabled = true;  // To enable
``` 
or
```C++
plugin.positioningSettings.egenioussEnabled = false; // To disable
```
By utilizing egenioussEnabled within the positioningSettings, we can control the visibility of the EgenioussReceiver.

---

### Download the Plugin
You can download the Egeniouss plugin [here](https://github.com/user-attachments/files/17098758/qfield-egeniouss-plugin.zip).


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for aac785171ee05d99e0190442a40bad26943e285f: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5637-48f49a24ee408601d91ee3f8468440d8932601f1-android-x86.apk)
</details>

src/core/positioning/egenioussreceiver.h
```diff
@@ -0,0 +1,33 @@
...
+    void onErrorOccurred( QAbstractSocket::SocketError socketError );
+
+  private:
+    void processReceivedData();
+
+  private:
+    QTcpSocket *mTcpSocket = nullptr;
+    QString mAddress;
+    int mPort = 0;
+    GnssPositionInformation mLastGnssPositionInformation;
```
nirvn, line 30: You don't need that, it's in the abstractgnssreceiver.h header and you've made this class a friend.

src/core/positioning/egenioussreceiver.h
```diff
@@ -0,0 +1,33 @@
...
+    QByteArray mReceivedData;
+    QByteArray payload;
```
nirvn, line 32: You don't need mReceivedData, let's just keep payload. Rename it to mPayLoad, and make it a QJsonObject (comment above)

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
+#include "egenioussreceiver.h"
+
+#include <QHostAddress>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QJsonValue>
+
+EgenioussReceiver::EgenioussReceiver( const QString &address, const int port, QObject *parent )
+  : AbstractGnssReceiver( parent ), mAddress( address ), mPort( port ), mTcpSocket( new QTcpSocket( this ) )
```
nirvn, line 9:
```suggestion
  : AbstractGnssReceiver( parent ), 
    mAddress( address ), 
    mPort( port ), 
    mTcpSocket( new QTcpSocket( this ) )
```

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+  setValid( !mAddress.isEmpty() && mPort > 0 );
+}
+
+void EgenioussReceiver::handleConnectDevice()
+{
+  connect( mTcpSocket, &QTcpSocket::readyRead, this, &EgenioussReceiver::onReadyRead );
+  connect( mTcpSocket, &QTcpSocket::errorOccurred, this, &EgenioussReceiver::onErrorOccurred );
```
nirvn, line 17: connect this in the constructor, we should connect/disconnect these every time.

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+  if ( mAddress.isEmpty() || mPort == 0 )
+  {
+    return;
+  }
+
+  mTcpSocket->connectToHost( QHostAddress( mAddress ), mPort, QTcpSocket::ReadWrite );
+
+  if ( !mTcpSocket->waitForConnected( 3000 ) )
```
nirvn, line 26: Don't wait for connection, listen to signals to change connected status.

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+void EgenioussReceiver::handleDisconnectDevice()
+{
+  disconnect( mTcpSocket, &QTcpSocket::readyRead, this, &EgenioussReceiver::onReadyRead );
+  disconnect( mTcpSocket, &QTcpSocket::errorOccurred, this, &EgenioussReceiver::onErrorOccurred );
+
+  if ( mTcpSocket->state() == QAbstractSocket::ConnectedState )
+  {
+    mTcpSocket->disconnectFromHost();
+    mTcpSocket->waitForDisconnected();
```
nirvn, line 46: Don't wait, listen to signal

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+QList<QPair<QString, QVariant>> EgenioussReceiver::details()
+{
+  QList<QPair<QString, QVariant>> dataList;
+
+  QJsonDocument jsonDoc = QJsonDocument::fromJson( payload );
+  QJsonObject jsonObject = jsonDoc.object();
```
nirvn, line 55: Don't store the payload as a byte array, store is as a QJsonObject. You already create that object in onReadyRead, let's avoid the double-cost here.

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+  if ( jsonDoc.isNull() || !jsonDoc.isObject() )
```
nirvn, line 57: Then here you can just do if ( mPayload.isEmpty() ) { return detailsList; }

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
...
+  if ( mTcpSocket->state() == QAbstractSocket::ConnectedState )
+  {
+    mTcpSocket->disconnectFromHost();
+    mTcpSocket->waitForDisconnected();
+  }
+}
+
+QList<QPair<QString, QVariant>> EgenioussReceiver::details()
+{
+  QList<QPair<QString, QVariant>> dataList;
```
nirvn, line 52:
```suggestion
  QList<QPair<QString, QVariant>> detailsList;
```

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,128 @@
...
+    {
+      return; // Failed to parse JSON
+    }
+
+    QJsonObject jsonObject = jsonDoc.object();
+
+    mLastGnssPositionInformation = GnssPositionInformation(
+      jsonObject.value( "lat" ).toDouble(),
+      jsonObject.value( "lon" ).toDouble(),
+      jsonObject.value( "alt" ).toDouble(),
```
nirvn, line 106: As the spec is developed, we'll need to make sure a lack of lat doesn't result is a 0.0 value but a quiet_NaN value.

src/core/positioning/positioning.cpp
```diff
@@ -191,20 +192,22 @@ void Positioning::setupDevice()
   }
   else
   {
+    const qsizetype portSeparator = mDeviceId.lastIndexOf( ':' );
+    const QString address = mDeviceId.mid( 4, portSeparator - 4 );
+    const int port = mDeviceId.mid( portSeparator + 1 ).toInt();
```
nirvn, line 197: As discussed in chat, we don't want a port / address configuration for egeniouss. You'll have to move that back into tcp / udp conditional blocks.

src/core/positioning/egenioussreceiver.h
mohsenD98: Correct :+1:

src/core/positioning/egenioussreceiver.h
mohsenD98: Nice :+1:

src/core/positioning/egenioussreceiver.cpp
mohsenD98: Correct :+1:

src/core/positioning/egenioussreceiver.cpp
mohsenD98: Correct :+1:

src/core/positioning/egenioussreceiver.cpp
mohsenD98: Correct :+1:

src/core/positioning/egenioussreceiver.cpp
mohsenD98: Nice :+1:

src/core/positioning/egenioussreceiver.cpp
mohsenD98: Yes :+1:

src/core/positioning/egenioussreceiver.cpp
mohsenD98: how is this : 
`const double latitude = mPayload.value( "lat" ).toDouble() == 0 ? std::numeric_limits<double>::quiet_NaN() : mPayload.value( "lat" ).toDouble();`

src/core/positioning/egenioussreceiver.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
m-kuhn, line 1: We normally do `#ifdef` include guards

src/qml/PluginManagerSettings.qml
```diff
@@ -14,6 +14,9 @@ Popup {
   y: (parent.height - height) / 2
   padding: 0
 
+  signal unLoadClicked(string Uuid)
+  signal loadClicked(string Uuid)
```
nirvn, line 18: Hm, I think it'd be nicer if we implemented something slightly different and have the plugin manager call function _attached to the plugin Item {} itself_ when enabling/disabling a plugin. 

That way we a/ empower plugin authors, and b/ we avoid adding code in qgismobileapp.qml to handle a plugin-specific Uuid handling which I'm not fond of.

src/core/pluginmanager.cpp
```diff
@@ -455,3 +457,27 @@ QString PluginManager::findProjectPlugin( const QString &projectPath )
...
+  const QString pluginPath = mAvailableAppPlugins[uuid].path();
+  if ( !mLoadedPlugins.contains( pluginPath ) )
+  {
+    return;
+  }
+
+  const QPointer<QObject> object = mLoadedPlugins[pluginPath];
+
+  const char *normalizedSignature = QMetaObject::normalizedSignature( ( methodName + "()" ).toStdString().c_str() );
+  const int methodExists = object->metaObject()->indexOfSlot( normalizedSignature );
```
nirvn, line 477: methodIndex would be a better name. Or have methodExists be a bool and add >= 0 at the end of this line.

nirvn:
@domi4484 , I'm good with this now. Do you want to have a look before we merge?

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,124 @@
...
+void EgenioussReceiver::handleDisconnectDevice()
+{
+  mTcpSocket->disconnectFromHost();
+}
+
+void EgenioussReceiver::disconnected()
+{
+  if ( mTcpSocket->state() == QAbstractSocket::ConnectedState )
+  {
+    mSocketState = QAbstractSocket::UnconnectedState;
```
domi4484, line 37: It feels a bit strange to me that subclasses of postioning device have to manage `mSocketState` that way. mSocketState and mSocketStateString have a quite risk to get out of sync. Wouldn't be it a cleaner interface to reimplement `AbstractGnssReceiver::socketState()` and `socketStateString` in subclasses and get rid of the variables mSocketState and mSocketStateString?

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,124 @@
...
+    emit socketStateChanged( mSocketState );
+  }
+}
+
+QList<QPair<QString, QVariant>> EgenioussReceiver::details()
```
domi4484, line 42: You could use `QVariantMap` to have dict like instead of plain pair list

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  emit lastGnssPositionInformationChanged( mLastGnssPositionInformation );
+}
+
+void EgenioussReceiver::handleError( QAbstractSocket::SocketError error )
+{
+  switch ( error )
+  {
+    case QAbstractSocket::HostNotFoundError:
+      mLastError = tr( "Could not find the remote host" );
```
domi4484, line 109: I guess with the current interface here you should set mSocketState and mSocketStateString as well

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,124 @@
...
+QList<QPair<QString, QVariant>> EgenioussReceiver::details()
+{
+  QList<QPair<QString, QVariant>> detailsList;
+
+  if ( mPayload.isEmpty() )
+  {
+    return detailsList;
+  }
+
+  detailsList.append( qMakePair( "q", mPayload.value( "q" ).toDouble() ) );
```
domi4484, line 51: For what stands `q`? Is a key in the json? Eventually to be put in a const to avoid having literals around

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  return detailsList;
+}
+
+void EgenioussReceiver::onReadyRead()
+{
+  QByteArray mReceivedData = mTcpSocket->readAll();
+  if ( mReceivedData.size() < 9 )
```
domi4484, line 59:
```suggestion
  if ( mReceivedData.size() < EgenioussReceiver::MINIMUM_DATA )
```
I would define consts for such kind of things

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,124 @@
...
+  {
+    return; // Received data is too short to process.
```
domi4484, line 61: What should we do here? maybe set mLastError? or just print some warnings?

src/core/positioning/egenioussreceiver.cpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+  if ( static_cast<uint8_t>( mReceivedData[0] ) != 0xFE )
+  {
+    return; // Invalid start byte
```
domi4484, line 65: Here as well missing error handling

src/core/positioning/positioningdevicemodel.cpp
```diff
@@ -171,6 +171,9 @@ const QString PositioningDeviceModel::deviceId( const Device &device ) const
 
     case SerialPortDevice:
       return QStringLiteral( "serial:%1" ).arg( device.settings.value( QStringLiteral( "address" ) ).toString() );
+
+    case EgenioussDevice:
+      return QStringLiteral( "egeniouss:" );
```
domi4484, line 176: I would put those prefixes into static const members. For `serial:`, `udp:`, and so on as well

src/core/positioning/tcpreceiver.cpp
```diff
@@ -119,7 +119,7 @@ void TcpReceiver::handleError( QAbstractSocket::SocketError error )
       mLastError = tr( "The connection was refused by the remote host" );
       break;
     default:
-      mLastError = tr( "UDP receiver error (%1)" ).arg( QMetaEnum::fromType<QAbstractSocket::SocketError>().valueToKey( error ) );
+      mLastError = tr( "TCP receiver error (%1)" ).arg( QMetaEnum::fromType<QAbstractSocket::SocketError>().valueToKey( error ) );
```
domi4484, line 122: In this function also mSocketState and socket state string should be updated

src/qml/EgenioussDeviceChooser.qml
domi4484, line 1: Would put this file into a egeniouss directory. Maybe into positioning/egeniouss and move as well other positioning related files in there to gain some structure

src/qml/PositioningSettings.qml
```diff
@@ -45,4 +45,5 @@ Settings {
   property int digitizingMeasureType: 1
 
   property bool geofencingPreventDigitizingDuringAlert: false
+  property bool enableEgeniouss: false
```
domi4484, line 48:
```suggestion
  property bool egenioussEnabled: false
```
But for `positioningActivated` we use Active instead of Enable. But in the code we see generally more often "Enabled". Sot not sure which one you should choose :smile:

domi4484 (commented):
Great thank you @mohsenD98!
I left some comments, @nirvn  please take also a look maybe you don't agree on all of them :)

src/core/positioning/egenioussreceiver.cpp
mohsenD98: `QVariantMap` does not maintain the order of its elements (i think it is implemented using a `Hash` table) but we need to maintain insertion order.

src/core/positioning/egenioussreceiver.cpp
mohsenD98: Yes, it is in JSON and we assume it stands for `Quality(?!)` however in my tests it was always zero.    

if we are sure it is quality i can go for something like this: 

```C++
const QString KEY_QUALITY = "q";
detailsList.append(qMakePair(KEY_QUALITY, mPayload.value(KEY_QUALITY).toDouble()));
```

@nirvn this `KEY_QUALITY` or `KeyQuality` or `keyQuality` ?

src/core/positioning/egenioussreceiver.h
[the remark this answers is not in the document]
nirvn: @domi4484 , let's maybe not do this right away. ATM we have our other device classes sitting in positioning (Bluetooth, TCP,  UDP, serial) here and having a directory for a single pair of .CPP, .H feels a bit too much. 

If we end up needing another class (eg egenioussutils) let's maybe consider moving all of it into an src/core/egeniouss directory)

src/core/positioning/egenioussreceiver.cpp
nirvn: @domi4484 , not a bad idea. Can we do that in a follow up PR?

src/core/positioning/egenioussreceiver.cpp
nirvn: Correct.

src/core/positioning/egenioussreceiver.cpp
nirvn: At this moment let's leave it as is, that's merely to have an extra detail in. We'll polish this when the protocol progresses and we have actual values to handle.

src/core/positioning/egenioussreceiver.cpp
nirvn: Good idea

src/core/positioning/positioningdevicemodel.cpp
nirvn: Good idea.

src/core/positioning/tcpreceiver.cpp
nirvn: Let's defer to different PR here to focus on egeniouss bits

src/qml/EgenioussDeviceChooser.qml
nirvn: I suggest this structural change be kept in a separate PR

nirvn (approved):

PR #5678, opened 2024-09-24
Possibility to use Difference/Clip processing algorithme

qsavoye:
1.  Select vector geometry features
2. Click on Process Selected Features
3. New algo are availabe especially from Vector Overlay group, Choose Difference algo
4. Select OVERLAY layer you want


src/core/processing/processingalgorithmparametersmodel.cpp
```diff
@@ -283,6 +285,20 @@ QVariant ProcessingAlgorithmParametersModelBase::data( const QModelIndex &index,
         const QgsProcessingParameterEnum *parameterEnum = dynamic_cast<const QgsProcessingParameterEnum *>( mParameters.at( index.row() ) );
         configuration["options"] = parameterEnum->options();
       }
+      else if ( mParameters.at( index.row() )->type() == QStringLiteral( "source" ) )
+      {
+        QMap<QString, QgsMapLayer *> map = QgsProject::instance()->mapLayers();
+        QStringList list;
```
nirvn, line 291: Instead of a string list of layer names here we should return a QVariantMap with keys being the layer _ID_ and the value being the layer name.  Referring to a layer by its ID is more resilient.

nirvn (commented):
@qsavoye , hey there, thanks for the contribution! I've added a couple of comments that'd need addressing. Looking forward to merge this :)

Could you also install the pre-commit hook to insure formatting is done so it passes our CI tests? See documentation here: https://github.com/opengisch/QField/blob/master/doc/dev.md#contribute

src/core/processing/processingalgorithmparametersmodel.cpp
qsavoye: Thanks @nirvn for the advices, indeed it is more resilient to search by ids

src/core/processing/processingalgorithmparametersmodel.cpp
```diff
@@ -283,6 +286,24 @@ QVariant ProcessingAlgorithmParametersModelBase::data( const QModelIndex &index,
...
+      else if ( mParameters.at( index.row() )->type() == QStringLiteral( "source" ) )
+      {
+        QMap<QString, QgsMapLayer *> map = QgsProject::instance()->mapLayers();
+        QVariantList list;
+        QVariantMap supportedLayers;
+        QgsProcessingContext context;
+        context.setProject( QgsProject::instance() );
+        for ( auto it = map.begin(); it != map.end(); ++it )
+        {
+          if ( it.value()->type() == Qgis::LayerType::Vector && QgsProcessingUtils::variantToSource( it.value()->name(), context, mParameters.at( index.row() )->defaultValue() ) )
```
nirvn, line 298: @qsavoye , we didn't need the vartiantToSource here, but what we definitively needed was to do a check against the parameter's dataTypes() to filter out of vector layers that are not part of list of geometry types passed by the algorithms. 

I've done that here: https://github.com/opengisch/QField/pull/5690/files#diff-15f7c96b0c450067ace3ed62a83f37fea47213c6cdefbf28bba20747ff6634d2R302

src/core/processing/processingalgorithmparametersmodel.cpp
```diff
@@ -151,8 +151,11 @@ void ProcessingAlgorithmParametersModelBase::rebuild()
           mHasAdvancedParameters = true;
         }
 
-        mParameters << definition;
-        mValues << definition->defaultValue();
+        if ( definition->name() != "INPUT" )
```
nirvn, line 154: Just to be extra safe, I've added a definition->type() == QStringLiteral( "source" ) condition here:

https://github.com/opengisch/QField/pull/5690/files#diff-15f7c96b0c450067ace3ed62a83f37fea47213c6cdefbf28bba20747ff6634d2R154

nirvn (commented):
Hey @qsavoye , thanks again for this, great way to contribute!

I did some more cleanups here https://github.com/opengisch/QField/pull/5690 (variable name renaming for clarity, and what I've mentioned in the second round of review).

In case you hadn't noticed, it also unlocked split by line, which is a nice alg. to have too!

nirvn:
Superseded by https://github.com/opengisch/QField/pull/5690

PR #5709, opened 2024-10-02
Fix QFieldGuide component: adjust pointerToItem.

mohsenD98:
# PR description: 

Somehow (!) pointer to item of QFieldGuide is out of sync!

image

this commit fixes that with more enhanced logic!

### DIFF
--- a/src/qml/QFieldGuide.qml
+++ b/src/qml/QFieldGuide.qml
@@ -319,16 +319,11 @@ Popup {
 
     x: {
       if (internalObject.target[0]) {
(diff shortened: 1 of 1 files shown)


nirvn:
I suspect this regressed at the 11th hour when we jumped back to qt 6.6.3.

Let's remember to check again when we go up to 6.7.3.

mohsenD98:
> I suspect this regressed at the 11th hour when we jumped back to qt 6.6.3.
> 
> Let's remember to check again when we go up to 6.7.3.

Agree

nirvn (approved):
Also tested on Android; looks good.

PR #5737, opened 2024-10-15
Add the possibility to rotate a single feature (line or polygon)

qsavoye:
https://github.com/user-attachments/assets/449771d6-4520-4001-a967-2e09993ba4b5


qsavoye:
I just remove some submodules depency that I do not had to commit previously, sorry for the incovenient.

nirvn:
@qsavoye , hey there, thanks for this contribution, it's fantastic! I'll get to review it ASAP (I've been a bit under the weather this week).

One thing I can already see as a need here is to change the icon to use a stock one from Google's material symbols & icons library: https://fonts.google.com/icons?icon.query=rotate -- that'll insure the style matches the rest of the app.

nirvn:
@qsavoye , oh, one question: you are aware that the processing toolbox (new feature in QField 3.4) has a rotate algorithm. Is there a reason why you wouldn't be using that instead?

qsavoye:
> @qsavoye , oh, one question: you are aware that the processing toolbox (new feature in QField 3.4) has a rotate algorithm. Is there a reason why you wouldn't be using that instead?

@nirvn It's more a question of user experience. I tried the processing toolbox for rotation on my Iphone, the parameter window hides a big part of the map canvas, which can be unpleasant
Furthermore, in my opinion, it is faster to rotate with drag or mouse pointer on mapcanvas than with a precise angle in degrees. I think both are complementary.

i18n/qfield_en.ts
```diff
@@ -1583,6 +1583,10 @@ The features geometries will be combined into feature &apos;%1&apos;, which will
         <source>Duplicate Feature</source>
         <translation>Duplicate Feature</translation>
     </message>
+    <message>
```
nirvn, line 1586: Can you remove the translation modifications here? Let transifex deal with it automatically.

src/core/featurelistextentcontroller.cpp
```diff
@@ -62,6 +62,29 @@ void FeatureListExtentController::zoomToSelected( bool skipIfIntersects ) const
   }
 }
 
+QgsPoint FeatureListExtentController::getCentroidFromSelected() const
+{
+  QgsPoint point( 0.0, 0.0, 0.0 );
```
nirvn, line 67: Returning a 0,0,0 point is valid, we won't have any way to know whether we're actually being fed a valid centroid or not. Let's just return a QgsPoint() instead.

```suggestion
```

src/core/featurelistextentcontroller.cpp
```diff
@@ -62,6 +62,29 @@ void FeatureListExtentController::zoomToSelected( bool skipIfIntersects ) const
...
+    if ( layer && layer->geometryType() != Qgis::GeometryType::Unknown && layer->geometryType() != Qgis::GeometryType::Null )
+    {
+      QgsGeometry geom = feat.geometry();
+
+      QgsCoordinateTransform transf( layer->crs(), mMapSettings->destinationCrs(), mMapSettings->mapSettings().transformContext() );
+      geom.transform( transf );
+
+      QgsPointXY centroid = geom.centroid().asPoint();
+      point.setX( centroid.x() );
+      point.setY( centroid.y() );
```
nirvn, line 82:
```suggestion
      return QgsPoint( geom.centroid().asPoint() );
```

src/core/featurelistextentcontroller.cpp
```diff
@@ -62,6 +62,29 @@ void FeatureListExtentController::zoomToSelected( bool skipIfIntersects ) const
...
+    }
+  }
+  return point;
```
nirvn, line 85:
```suggestion
  return QgsPoint();
```

src/core/featurelistextentcontroller.cpp
```diff
@@ -62,6 +62,29 @@ void FeatureListExtentController::zoomToSelected( bool skipIfIntersects ) const
...
+  {
+    QgsFeature feat = mSelection->focusedFeature();
+    QgsVectorLayer *layer = mSelection->focusedLayer();
+
+    if ( layer && layer->geometryType() != Qgis::GeometryType::Unknown && layer->geometryType() != Qgis::GeometryType::Null )
+    {
+      QgsGeometry geom = feat.geometry();
+
+      QgsCoordinateTransform transf( layer->crs(), mMapSettings->destinationCrs(), mMapSettings->mapSettings().transformContext() );
+      geom.transform( transf );
```
nirvn, line 78: Let's add some const.
```suggestion
    const QgsFeature feat = mSelection->focusedFeature();
    const QgsVectorLayer *layer = mSelection->focusedLayer();

    if ( layer && layer->geometryType() != Qgis::GeometryType::Unknown && layer->geometryType() != Qgis::GeometryType::Null )
    {
      QgsGeometry geom = feat.geometry();

      const QgsCoordinateTransform transf( layer->crs(), mMapSettings->destinationCrs(), mMapSettings->mapSettings().transformContext() );
      geom.transform( transf );
```

src/core/multifeaturelistmodelbase.cpp
```diff
@@ -386,6 +386,38 @@ bool MultiFeatureListModelBase::canMoveSelection() const
   return true;
 }
 
+bool MultiFeatureListModelBase::canRotateSelection() const
+{
+  if ( mSelectedFeatures.isEmpty() )
+    return false;
+
+  QgsVectorLayer *vlayer = mSelectedFeatures[0].first;
+  if ( !vlayer || vlayer->readOnly() || !( vlayer->dataProvider()->capabilities() & QgsVectorDataProvider::ChangeGeometries ) || vlayer->customProperty( QStringLiteral( "QFieldSync/is_geometry_locked" ), false ).toBool() || ( vlayer->geometryType() == Qgis::GeometryType::Point ) )
```
nirvn, line 395: Actually, as unintuitive as it might sound, points are valid geometries to rotate when you are dealing with multipoint, we should allow for it.

src/core/multifeaturelistmodelbase.cpp
```diff
@@ -624,6 +656,47 @@ bool MultiFeatureListModelBase::moveSelection( const double x, const double y )
...
+  }
+
+  if ( isSuccess )
+  {
+    // commit changes
+    isSuccess = vlayer->commitChanges();
+  }
+
+  if ( !isSuccess )
+  {
```
nirvn, line 692:
```suggestion
  }
  else
  {
```

src/qml/MapCanvas.qml
```diff
@@ -33,6 +33,7 @@ Item {
   property bool hovered: false
   property bool pinched: pinchHandler.active
   property bool freehandDigitizing: false
+  property bool isRotationFeaturesRequested: false
```
nirvn, line 36: I am extremely sensitive to touching MapCanvas and its handlers, can we revert all changes here please.

If you are fixing something that's not related to feature rotation, it's most welcome, but please do so in a separate PR. :)

src/qml/NavigationBar.qml
```diff
@@ -755,6 +756,21 @@ Rectangle {
       onTriggered: duplicateClicked()
     }
 
+    MenuItem {
+      id: rotateFeatureBtn
+      text: qsTr('Rotate Feature')
+      icon.source: Theme.getThemeVectorIcon("ic_rotate_white_24dp")
+      // allow only rotation for line or polygon
```
nirvn, line 763: As said above, we should allow points too, because multipoints can be rotated.

src/qml/qgismobileapp.qml
```diff
@@ -322,6 +322,42 @@ ApplicationWindow {
       }
     }
 
+    DragHandler {
+      id: rotateDragHandler
+      enabled: rotateFeaturesToolbar.rotateFeaturesRequested == true && !freehandButton.visible && !freehandButton.freehandDigitizing
```
nirvn, line 327: Hm, here I don't think we should care about freehand button -- if we have requested a feature rotation, we wouldn't be freehand digitizing anyways.

src/qml/qgismobileapp.qml
```diff
@@ -455,6 +491,7 @@ ApplicationWindow {
       quality: qfieldSettings.quality
       forceDeferredLayersRepaint: trackings.count > 0
       freehandDigitizing: freehandButton.freehandDigitizing && freehandHandler.active
+      isRotationFeaturesRequested: rotateFeaturesToolbar.rotateFeaturesRequested
```
nirvn, line 494: -1 , just add it into the isEnabled property.

src/qml/qgismobileapp.qml
```diff
@@ -805,6 +842,25 @@ ApplicationWindow {
       width: 5
     }
 
+    /* Highlight the currently selected item being rotated */
+    FeatureListSelectionHighlight {
```
nirvn, line 846: Could we instead use only a single FeatureListSelectionHighlight for both moving and rotating features? Thanks.

nirvn (commented):
@qsavoye , hey there, sorry for the delayed review, it's been one of those months :) 

This would make a great feature for QField's next release (3.5). Let me know if you have questions on the review comments.

i18n/qfield_en.ts
qsavoye: Should I remove all translations modification that i made (fr,es,it) ?

nirvn: @qsavoye , yes, please, all translation modifications should be reverted.

nirvn:
@qsavoye , thanks for addressing the review. Will merge as soon as the CIs turn green. Thanks for this contribution :pray:

qsavoye:
@nirvn I think my there was an update between the current master and and the master where i start these pull request on file multifeaturelistmodelbase.cpp

in canRotateSelection function
```c++
if ( !vlayer || vlayer->readOnly() || !( vlayer->dataProvider()->capabilities() & QgsVectorDataProvider::ChangeGeometries ) || vlayer->customProperty( QStringLiteral( "QFieldSync/is_geometry_locked" ), false ).toBool() )
```
must be replace to 

```c++
if ( !vlayer || vlayer->readOnly() || !( vlayer->dataProvider()->capabilities() & Qgis::VectorProviderCapability::ChangeGeometries ) || vlayer->customProperty( QStringLiteral( "QFieldSync/is_geometry_locked" ), false ).toBool() )
```

Do you see other change which can block the merge ?

nirvn:
@qsavoye , no, if that's fixed, I think we can merge. 

There was indeed a QGIS core library update which did change a couple of enum keys, inc. this one.

nirvn:
@qsavoye , can you resolve the conflict too?

PR #5825, opened 2024-11-16
Improve map rotation by mouse event

qsavoye:
When using mouse, the map rotation calculation is improved

### DIFF
--- a/src/qml/MapCanvas.qml
+++ b/src/qml/MapCanvas.qml
@@ -355,14 +355,20 @@ Item {
     grabPermissions: PointerHandler.TakeOverForbidden
     acceptedButtons: Qt.MiddleButton
 
-    property real oldTranslationY: 0
-    property bool translationThresholdReached: false
+    property real pressClickX: 0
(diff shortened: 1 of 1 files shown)


src/qml/MapCanvas.qml
```diff
@@ -355,30 +355,40 @@ Item {
...
     }
 
     onTranslationChanged: {
       if (active) {
-        if (translationThresholdReached) {
-          if (oldTranslationY != 0) {
-            mapCanvasWrapper.rotate(oldTranslationY - translation.y);
+        if (lastRotateAngle != 0) {
+          var newPositionX = pressClickX + translation.x;
+          var newPositionY = pressClickY + translation.y;
```
suricactus, line 381:
```suggestion
          let newPositionX = pressClickX + translation.x;
          let newPositionY = pressClickY + translation.y;
```

nirvn (approved):

nirvn:
@qsavoye , approved, thanks.

PR #5842, opened 2024-11-21
Use curve to digitize Polylines/Polygons

qsavoye:
Use curve to digitize Polylines/Polygons

test

1. Active Curve mode during Digitization of a Polygon
2. Add a first vertex to define the center point of the arc
3. Add a second vertex to stop the arc

Precision of the final arc can be set in QField Parameters page (slider)

### DIFF
--- a/src/qml/DigitizingToolbar.qml
+++ b/src/qml/DigitizingToolbar.qml
@@ -242,7 +242,19 @@ QfVisibilityFadingRow {
       if (Number(rubberbandModel.geometryType) === Qgis.GeometryType.Point || Number(rubberbandModel.geometryType) === Qgis.GeometryType.Null) {
         confirm();
       } else {
-        addVertex();
+        if (settings.valueBool("/QField/Digitizing/CurveEdition", false) == true) {
+          if (currentRubberband.model.isDuringCurveDrawing() == true || currentRubberband.model.vertexCount == 1) {
+            if (currentRubberband.model.vertexCount != 1) {
+              addCurve();
+            } else {
+              addVertex();
+            }
+          } else {
+            addMiddlePointCurve();
+          }
+        } else {
+          addVertex();
+        }
       }
     }
   }
@@ -279,6 +291,14 @@ QfVisibilityFadingRow {
 
   function triggerAddVertex() {
     addVertexButton.clicked();
+  }
+
+  function addCurve() {
+    rubberbandModel.addCurve();
+  }
+
+  function addMiddlePointCurve() {
+    rubberbandModel.addMiddlePointCurve();
   }
 
   function addVertex() {
--- a/src/qml/qgismobileapp.qml
+++ b/src/qml/qgismobileapp.qml
@@ -504,7 +504,19 @@ ApplicationWindow {
               if (Number(currentRubberband.model.geometryType) === Qgis.GeometryType.Point || Number(currentRubberband.model.geometryType) === Qgis.GeometryType.Null) {
                 digitizingToolbar.confirm();
               } else {
-                digitizingToolbar.addVertex();
+                if (settings.valueBool("/QField/Digitizing/CurveEdition", false) == true) {
+                  if (currentRubberband.model.isDuringCurveDrawing() == true || currentRubberband.model.vertexCount == 1) {
+                    if (currentRubberband.model.vertexCount != 1) {
+                      digitizingToolbar.addCurve();
+                    } else {
+                      digitizingToolbar.addVertex();
+                    }
+                  } else {
+                    digitizingToolbar.addMiddlePointCurve();
+                  }
+                } else {
+                  digitizingToolbar.addVertex();
+                }
               }
             }
           } else {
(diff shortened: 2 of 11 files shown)


src/qml/qgismobileapp.qml
```diff
@@ -1559,6 +1571,51 @@ ApplicationWindow {
...
           }
         }
+
+        QfToolButton {
+          id: curveEditionButton
+          width: visible ? 40 : 0
+          height: visible ? 40 : 0
+          padding: 2
+          round: true
+          visible: dashBoard.activeLayer && (dashBoard.activeLayer.geometryType() === Qgis.GeometryType.Polygon || dashBoard.activeLayer.geometryType() === Qgis.GeometryType.Line)
```
m-kuhn, line 1581: I think we want this to be dependent on `QgsWkbTypes.isCurvedType()`

qsavoye: In my example , my layer is not a curve type at all and i use QgsCompoundCurve::curveToLine function to set geometry drawn into my none curved layer.  This helps me to drawn faster instead of set vertex by vertex.

But you are right i am not taking into account the possibility to set directly the qgscompoundcurve geometry into the new feature.

How do you think ?

m-kuhn: Interesting, I was mostly seeing this in the context of surveying / cadastre, where curves are often a requirement. My main motivation here would be to keep the interface as light as possible and ask users to have a curve layer if they want this feature. What do you think?

qsavoye: I am using it to draw underground maps, such as caves or old quarries. My polygon layer corresponds to some galleries and i no need to have a specific curved layer but some times digitizing arcs like in qgis help me.

Maybe we can allow it only for curved layer (QgsWkbTypes.isCurvedType()) or for advanced digitizing user ( which could be set in settings).

Tell me what you think ?

Sorry for the delay, i had some stuff to do last week.

m-kuhn: Interesting job you have!

I'd be interested to hear @nirvn's point of view

nirvn: @m-kuhn , @qsavoye , hey there, sorry for the delay in chipping in. 

I do think the feature is useful; I would indeed prefer if this option was only visible for curved type layers, I feel that'd be a good way to enable this while not cluttering UI by limiting to an explicit intent by the user / project manager.

src/qml/DigitizingToolbar.qml
```diff
@@ -242,7 +242,19 @@ QfVisibilityFadingRow {
       if (Number(rubberbandModel.geometryType) === Qgis.GeometryType.Point || Number(rubberbandModel.geometryType) === Qgis.GeometryType.Null) {
         confirm();
       } else {
-        addVertex();
+        if (settings.valueBool("/QField/Digitizing/CurveEdition", false) == true) {
```
nirvn, line 245: I think we need an additional check here to avoid curved edition when people are freehand drawing, otherwise worlds will collide.

m-kuhn:
@9ls1 I noted you added a :+1: to this, do you have a specific use case in mind for which this would be interesting?

9ls1:
@m-kuhn No, just in general thinking about saving time digitizing shapes like the one shown here and curved roads.

src/qml/qgismobileapp.qml
qsavoye: @m-kuhn @nirvn, 
Ok with your reviews.
I will patch it for only allowing layer with QgsWkbTypes.isCurvedType() as soon as I have time

nirvn:
@qsavoye , you have not been forgotten :) I'll review this in the coming day. Thanks for your patience.

qsavoye:
> @qsavoye , are you planning to work on this further?

@nirvn , yes but i will be unavailable to work on it for at least a month :(

nirvn:
@qsavoye , no problem, take your time. You're contributions are appreciated, just glad to hear you're planning to come back to it :)

src/core/utils/layerutils.cpp
```diff
@@ -383,3 +383,11 @@ FeatureIterator LayerUtils::createFeatureIteratorFromExpression( QgsVectorLayer
   const QgsFeatureRequest request = QgsFeatureRequest( QgsExpression( expression ) );
   return FeatureIterator( layer, request );
 }
+
+bool LayerUtils::isCurvedGeometry( QgsVectorLayer *layer )
```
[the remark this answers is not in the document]
qsavoye: @nirvn 
Do you prefer :  make a patch when compiling qgis to modify QgsWkbTypes.h and erase what I did on LayerUtils or  move the method into GeometryUtils ?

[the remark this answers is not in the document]
nirvn: @qsavoye , I'd love for you to submit a patch in QGIS upstream (we're 5 days away from 3.44 release), then we can update QField to 3.44 when released and we'll be able to make sure of that newly-made invokable function.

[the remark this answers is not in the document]
nirvn: @qsavoye , I've taken care of upstream here: https://github.com/qgis/QGIS/pull/62306

nirvn:
@qsavoye , if you don't mind, I will close this pull request for now, it'll need some work rebasing against latest master, etc.

I do hope you will find some time to get back to coding this feature, your contributions are most welcome!

PR #5905, opened 2024-12-20
QF-1414-Download and writing data to a temporary file in QFieldCloudProjects.

mohsenD98:
**PR Description:**  
  
This PR aims to handle the download progress of a network reply by writing the downloaded data incrementally to a temporary file. The key steps include:

- **Listening to Download Progress:** We connect the `downloadProgress` signal of the `NetworkReply` to a lambda function that processes the download as it progresses.

- **File Writing:** Inside the lambda, we open the temporary file in append mode `(QIODevice::Append)` and write the data to it each time new bytes are received.

- **Error Handling:** If any error occurs while opening the file or during the write operation, we log the error and abort the network reply `(rawReply->abort())`.

- **Reading Data:** For each progress update, we read all available data from the reply using `readAll()` and write it to the file.

**Ui polish:**  
       
1. Allow changing tabs ["My Projects" / "Community"].
2. Fix invisible progressBar under headersection (layer name).
3. Fix spacing between progressBar and next item.

**Summary:**

This approach focuses on incrementally saving downloaded data to a temporary file during each progress update. It includes basic error handling and attempts to append data to the file as it is received.

### DIFF
--- a/src/core/networkreply.cpp
+++ b/src/core/networkreply.cpp
@@ -41,10 +41,7 @@ void NetworkReply::abort()
 
 QNetworkReply *NetworkReply::reply() const
 {
-  if ( mIsFinished )
-    return mReply;
-
-  return nullptr;
+  return mReply;
 }
 
 
--- a/src/core/networkreply.cpp
+++ b/src/core/networkreply.cpp
@@ -39,7 +39,7 @@ void NetworkReply::abort()
 }
 
 
-QNetworkReply *NetworkReply::reply() const
+QNetworkReply *NetworkReply::currentRawReply() const
 {
   return mReply;
 }
@@ -87,6 +87,7 @@ void NetworkReply::initiateRequest()
     case QNetworkAccessManager::UnknownOperation:
       throw QStringLiteral( "Not implemented!" );
   }
+  emit currentRawReplyChanged();
 
   mReply->ignoreSslErrors( mExpectedSslErrors );
 
--- a/src/core/networkreply.h
+++ b/src/core/networkreply.h
@@ -58,7 +58,7 @@ class NetworkReply : public QObject
 
 
     /**
-     * Get the QNetworkReply object. Do not delete it manually.
+     * Get the current `QNetworkReply` object. Note that it might get deleted even if the parent `NetworkReply` is not in case of redirect or internal retry. Do not delete it manually.
      * @return network reply
      */
     QNetworkReply *reply() const;
--- a/src/core/networkreply.h
+++ b/src/core/networkreply.h
@@ -58,7 +58,7 @@ class NetworkReply : public QObject
 
 
     /**
-     * Get the QNetworkReply object once the CloudReply is finilized. Do not delete it manually.
+     * Get the QNetworkReply object. Do not delete it manually.
      * @return network reply
      */
     QNetworkReply *reply() const;
--- a/src/core/networkreply.h
+++ b/src/core/networkreply.h
@@ -59,9 +59,9 @@ class NetworkReply : public QObject
 
     /**
      * Get the current `QNetworkReply` object. Note that it might get deleted even if the parent `NetworkReply` is not in case of redirect or internal retry. Do not delete it manually.
-     * @return network reply
+     * @return network currentRawReply
      */
-    QNetworkReply *reply() const;
+    QNetworkReply *currentRawReply() const;
 
 
     /**
@@ -144,6 +144,11 @@ class NetworkReply : public QObject
      * @param code
      */
     void temporaryErrorOccurred( QNetworkReply::NetworkError code );
+
+    /**
+     * Emitted when reply has changed.
+     */
+    void currentRawReplyChanged();
 
   private:
     /**
--- a/src/core/qfieldcloudprojectsmodel.cpp
+++ b/src/core/qfieldcloudprojectsmodel.cpp
@@ -1736,7 +1736,6 @@ NetworkReply *QFieldCloudProjectsModel::downloadFile( const QString &projectId, 
   return mCloudConnection->get( request, QStringLiteral( "/api/v1/packages/%1/latest/files/%2/" ).arg( projectId, fileName ) );
 }
 
-
 void QFieldCloudProjectsModel::downloadFileConnections( const QString &projectId, const QString &fileName )
 {
   const QModelIndex projectIndex = findProjectIndex( projectId );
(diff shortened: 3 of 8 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for d6091339db6ba6d43717bd8f338433ea2f19b774: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5905-6b2945fed7670bd0a2abdab21edf8774bd8a0ce4-android-x86.apk)
</details>

src/core/networkreply.cpp
```diff
@@ -41,10 +41,7 @@ void NetworkReply::abort()
 
 QNetworkReply *NetworkReply::reply() const
 {
-  if ( mIsFinished )
-    return mReply;
-
-  return nullptr;
+  return mReply;
```
nirvn, line 44: @suricactus , was there a reason why you were not passing on the pointer until finished?

suricactus: Yes, because there is the internal retry and the `QNetworkReply` object might change. if this is going to be changed, please also update the docs here: https://github.com/opengisch/QField/blob/QField-Cloud-Projects/src/core/networkreply.h#L61

src/core/qfieldcloudprojectsmodel.cpp
```diff
@@ -1809,6 +1809,26 @@ void QFieldCloudProjectsModel::downloadFileConnections( const QString &projectId
...
+    QNetworkReply *rawReply = reply->reply();
+    QFile file( temporaryFileName );
+
+    if ( file.open( QIODevice::WriteOnly | QIODevice::Append ) )
+    {
+      file.write( rawReply->readAll() );
+
+      if ( file.error() != QFile::NoError )
+      {
+        QgsLogger::debug( QStringLiteral( "Failed to write on temporary file `%1`." ).arg( temporaryFileName ) );
```
suricactus, line 1822: I would keep as much details as possible about the failure, such as original filename, project id etc. 

Also note that before the failures were logged also to the user in `QgsMessageLog::logMessage`, I would prefer to have this back.

src/core/networkreply.cpp
nirvn: Ok we'll update the documentation then. We definitely need access to it during download process so we offload bits from memory to disk as they come in.

suricactus: Be careful that the `QNetworkReply` might be deleted even if the parent `NetworkReply` is still alive.

src/core/networkreply.h
```diff
@@ -58,7 +58,7 @@ class NetworkReply : public QObject
 
 
     /**
-     * Get the QNetworkReply object once the CloudReply is finilized. Do not delete it manually.
+     * Get the QNetworkReply object. Do not delete it manually.
```
suricactus, line 61:
```suggestion
     * Get the current `QNetworkReply` object. Note that it might get deleted even if the parent `NetworkReply` is not in case of redirect or internal retry. Do not delete it manually.
```

mohsenD98: Thanks Ivan :+1:

mohsenD98: @suricactus  do you think should we handle it like this: 

```
    if( !rawReply ) {
      hasError = true;
      errorMessageDetail = "";
      errorMessage = tr( "" );
    }
```

and never call `rawReply->readAll()`.

suricactus: @mohsenD98 what if we rename the getter to `currentRawReply()` and add a new signal `currentRawReplyChanged()`, so the caller would know at any time if the reply has changed?

If you receive `NetworkReply::downloadProgress` most probably `currentRawReply()` will return an active pointer, if it doesn't, I think a simple `if (!rawReply) return` should be enough?

mohsenD98: @nirvn what do you think about this ?

nirvn: @suricactus , @mohsenD98 , this sounds good to me.

src/core/qfieldcloudprojectsmodel.h
```diff
@@ -496,6 +496,7 @@ class QFieldCloudProjectsModel : public QAbstractListModel
     void downloadFileConnections( const QString &projectId, const QString &fileName );
     void loadProjects( const QJsonArray &remoteProjects = QJsonArray(), bool skipLocalProjects = false );
     void insertProjects( const QList<CloudProject *> &projects );
+    void logFailedDownload( CloudProject *projectName, const QString &projectId, const QString &fileName, const QString &errorMessage, const QString &errorMessageDetail );
```
suricactus, line 499:
```suggestion
    void logFailedDownload( CloudProject *project, const QString &fileName, const QString &errorMessage, const QString &errorMessageDetail );
```

Note you should rename `projectName` to `project`, as it has the `CloudProject` instance (and you use that name in the `.cpp` file). Also no need for explicit `projectId` param, as it is already in the `CloudProject` instance. Even if it is not (I did not check), please add it there.

mohsenD98: There is `project->id` i think that would work.

suricactus (approved):
Code looks good. Did not test. Thanks!

nirvn (approved):

PR #5908, opened 2024-12-23
Add setSubsetString into layerutils

woutergd:
For a plugin in QField we are creating, we want to apply a filter dynamically to a layer. We were however unable to access the setSubsetString method on a vectorlayer from the qml files, so hereby a PR which add this method to the LayerUtils so it becomes accessible through plugins.


nirvn:
@woutergd , greatings -- actually , the setSubsetString function is exposed through a Q_PROPERTY on the vector layer (which is more QMLesque :) ). To retrieve _or_ set the current subset string, you'd go through the property.

E.g. to get the current subset string:

```
let layer = qgisProject.mapLayersByName('my_vector_layer')[0]
console.log(layer.subsetString)
```

To set a subset string, simply do:

```
let layer = qgisProject.mapLayersByName('my_vector_layer')[0]
layer.subsetString = '"TYPE" = 1'
```

Voila :)

That said, thanks for opening a pull request with new code, while this specific PR isn't going to be merged, I certainly hope this will not be the last time we see you guys around :)

Out of curiosity, what functionalities are you adding through your plugin?

nirvn:
@woutergd , if you check the QGIS API documentation , you'll notice each class can feature a Properties section which lists all available Q_PROPERTY within a given class flagged as Q_OBJECT or Q_GADGET.

E.g., the QgsVectorLayer's subsetString : https://api.qgis.org/api/classQgsVectorLayer.html#a82752c93ea29f4789e2973ef2090b9e7

You can also spot invokable functions by looking for the Q_INKABLE tag next to individual functions (not great, but better than nothing).

I'm always delighted to hear about plugin use cases, you can reach out via mathieu@opengis.ch (since I do not have your email at hand :) ).

Issue #5897, opened 2024-12-16
Build IOS error on MacOS sequoia Version 15.2 


mansivincent:
Hello everyone, 
I'm trying to build an IOS environment with the following command: 

cmake -S . -B build-arm64-ios
        -DVCPKG_TARGET_TRIPLET=arm64-ios \
        -DWITH_VCPKG=ON \
        -DVCPKG_BUILD_TYPE=release \
        -DCMAKE_SYSTEM_NAME=iOS \
        -DCMAKE_OSX_SYSROOT=iphoneos \
        -DCMAKE_OSX_ARCHITECTURES=arm64 \
        -DCMAKE_SYSTEM_PROCESSOR=aarch64 \
        -GXcode
However I get this error: 

The following build commands failed:
        PhaseScriptExecution Generate\ vcpkg-populate-prefix/src/vcpkg-populate-stamp/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)/vcpkg-populate-download 
/qfield_project/QField/build-arm64-ios/_deps/vcpkg-subbuild/build/vcpkg-populate.build/Debug/vcpkg-populate.build/Script-7ED1C045FE7F1AA40556E615.sh (in target 'vcpkg-populate' from project 'vcpkg-populate')
(1 failure)

CMake Error at /usr/local/share/cmake/Modules/FetchContent.cmake:1918 (message):
  Build step for vcpkg failed: 65
Call Stack (most recent call first):
  /usr/local/share/cmake/Modules/FetchContent.cmake:1609 (__FetchContent_populateSubbuild)
  /usr/local/share/cmake/Modules/FetchContent.cmake:2145:EVAL:2 (__FetchContent_doPopulation)
  /usr/local/share/cmake/Modules/FetchContent.cmake:2145 (cmake_language)
  /usr/local/share/cmake/Modules/FetchContent.cmake:2384 (__FetchContent_Populate)
  cmake/VcpkgToolchain.cmake:34 (FetchContent_MakeAvailable)
  CMakeLists.txt:26 (include)

after some modifications suggested by users in other issues, i still can't configure my build
the command line I placed came from this pull request: https://github.com/opengisch/QField/pull/5016/files

nirvn:
This looks like an issue that was fixed on master yesterday. Can you make sure you pull all changes up to this very moment?

mansivincent:
thank you for your reply 
[...]

PR #5909, opened 2024-12-23
Update IOS cmake command in docs

woutergd:
Update the cmake-command in the docs for the iOS build, as a parameter is missing to get the build successfully passed. See also #5897.

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -213,18 +213,31 @@ xcodebuild -downloadPlatform iOS
 export PATH="$(brew --prefix flex)/bin:$(brew --prefix bison)/bin:$PATH"
 
 # Configure using CMake
+#
+# In the command below, the following line depends on the host system, i.e. the system
+# you are building on. In case you are building on a x64 (Intel) based system instead of
+# an Apple Silicon-based system (M1/M2/..), change this line:
+#
+#   -DVCPKG_HOST_TRIPLET=arm64-osx \
+#
+# into
+#
+#   -DVCPKG_HOST_TRIPLET=x64-osx \
+#
 
 cmake -S . -B build-arm64-ios \
+	-DVCPKG_HOST_TRIPLET=arm64-osx \
 	-DVCPKG_TARGET_TRIPLET=arm64-ios \
 	-DWITH_VCPKG=ON \
 	-DVCPKG_BUILD_TYPE=release \
 	-DCMAKE_SYSTEM_NAME=iOS \
 	-DCMAKE_OSX_SYSROOT=iphoneos \
 	-DCMAKE_OSX_ARCHITECTURES=arm64 \
-        -DCMAKE_SYSTEM_PROCESSOR=aarch64 \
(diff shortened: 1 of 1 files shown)


doc/dev.md
```diff
@@ -213,18 +213,23 @@ xcodebuild -downloadPlatform iOS
 export PATH="$(brew --prefix flex)/bin:$(brew --prefix bison)/bin:$PATH"
 
 # Configure using CMake
+#
+# Make sure to update the `VCPKG_HOST_TRIPLET` parameter below if you are using an x64 MacOS system to the value `x64-osx`. There are
+# also known issues on MacOS, check this issue for the current status: https://github.com/opengisch/QField/issues/5897.
 
 cmake -S . -B build-arm64-ios \
+   -DVCPKG_HOST_TRIPLET=arm64-osx
```
nirvn, line 221: Spacing issue here (as well as below with DCMAKE_SYSTEM_PROCESSOR , can it be fixed?

doc/dev.md
```diff
@@ -213,18 +213,23 @@ xcodebuild -downloadPlatform iOS
 export PATH="$(brew --prefix flex)/bin:$(brew --prefix bison)/bin:$PATH"
 
 # Configure using CMake
+#
+# Make sure to update the `VCPKG_HOST_TRIPLET` parameter below if you are using an x64 MacOS system to the value `x64-osx`. There are
+# also known issues on MacOS, check this issue for the current status: https://github.com/opengisch/QField/issues/5897.
```
nirvn, line 218: Can we remove the github issue? I think we want to keep our documentation as time-proof as possible :)

doc/dev.md
woutergd: Updated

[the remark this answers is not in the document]
woutergd: Thanks for the suggestion, applied

doc/dev.md
woutergd: Updated it based on the suggestion below, which clarifies it even more.

woutergd:
@m-kuhn @nirvn PR is updated based on the comments above

m-kuhn (approved):
Thanks a lot for this @woutergd !

nirvn:
@woutergd , @m-kuhn , this looks a bit weird ATM, it looks like the VCPKG_HOST_TRIPLET=arm64-osx is to be executed, not part of the comment:

image

I'd suggest something like:

```
# In case you are building on a x64 (Intel) based system, change this definition
# VCPKG_HOST_TRIPLET=arm64-osx to VCPKG_HOST_TRIPLET=x64-osx
```

woutergd:
@nirvn new try, good point. Extended the explanation a bit and added a bit more whitespace around the line to change in the comment just to make it clear what need to be changed. Adding it to a single line would make it a bit more obfuscated in my opinion that something, and especially what, need to be changed.

nirvn:
Perfect thanks

PR #5917, opened 2024-12-27
Background tracking support

nirvn:
This PR implements background tracking support in QField :partying_face: 

Implementation notes:
- Users must return to the QField app for the collected positions while QField was suspended to be saved into the vector layer tracks. This is a requirement to insure features will get proper expression context et al and avoid serious issues with vector layers co-existing in multiple processes.
- The improvements fix an interesting out-of-sync situation where the position information was not necessarily matching the projected position (which is used to add vertices to the track feature).

Plenty of code refactoring has happened, hope it doesn't make it too much harder to review this.

### DIFF
--- a/src/core/positioning/positioningsource.cpp
+++ b/src/core/positioning/positioningsource.cpp
@@ -150,7 +150,35 @@ void PositioningSource::setBackgroundMode( bool backgroundMode )
 
   mBackgroundMode = backgroundMode;
 
+  if ( mBackgroundMode )
+  {
+    QFile file( QStringLiteral( "%1.information" ).arg( backgroundFilePath ) );
+    file.open( QFile::WriteOnly );
+    file.close();
+  }
+
   emit backgroundModeChanged();
+}
+
+QList<GnssPositionInformation> PositioningSource::getBackgroundPositionInformation() const
+{
+  QList<GnssPositionInformation> positionInformationList;
+
+  QFile file( QStringLiteral( "%1.information" ).arg( backgroundFilePath ) );
+  if ( file.exists() )
+  {
+    file.open( QFile::ReadOnly );
+    QDataStream stream( &file );
+    while ( !stream.atEnd() )
+    {
+      GnssPositionInformation positionInformation;
+      stream >> positionInformation;
+      positionInformationList << positionInformation;
+    }
+    file.close();
+  }
+
+  return std::move( positionInformationList );
 }
 
 void PositioningSource::setElevationCorrectionMode( ElevationCorrectionMode elevationCorrectionMode )
@@ -297,6 +325,14 @@ void PositioningSource::lastGnssPositionInformationChanged( const GnssPositionIn
     {
       emit averagedPositionCountChanged();
     }
+  }
+  else
+  {
+    QFile file( QStringLiteral( "%1.information" ).arg( backgroundFilePath ) );
+    file.open( QFile::Append );
+    QDataStream stream( &file );
+    stream << mPositionInformation;
+    file.close();
   }
 }
 
--- a/src/core/positioning/positioningsource.cpp
+++ b/src/core/positioning/positioningsource.cpp
@@ -145,7 +145,6 @@ void PositioningSource::setLogging( bool logging )
 
 void PositioningSource::setBackgroundMode( bool backgroundMode )
 {
-  qDebug() << "sss setBackgroundMode " << ( backgroundMode ? "true" : "false" );
   if ( mBackgroundMode == backgroundMode )
     return;
 
@@ -165,7 +164,6 @@ void PositioningSource::setBackgroundMode( bool backgroundMode )
 
 QList<GnssPositionInformation> PositioningSource::getBackgroundPositionInformation() const
 {
-  qDebug() << "sss getBackgroundPositionInformation";
   QList<GnssPositionInformation> positionInformationList;
 
   QFile file( QStringLiteral( "%1.information" ).arg( backgroundFilePath ) );
@@ -181,7 +179,6 @@ QList<GnssPositionInformation> PositioningSource::getBackgroundPositionInformati
     }
     file.close();
   }
-  qDebug() << "sss collected positions " << positionInformationList.size();
 
   return std::move( positionInformationList );
 }
@@ -333,7 +330,6 @@ void PositioningSource::lastGnssPositionInformationChanged( const GnssPositionIn
   }
   else
   {
-    qDebug() << "sss writing to information file";
     QFile file( QStringLiteral( "%1.information" ).arg( backgroundFilePath ) );
     file.open( QFile::Append );
     QDataStream stream( &file );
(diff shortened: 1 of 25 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 2015f356b12b5ffddf9ca25d8acd96cab9efed6b: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-5917-97712f8887030c8d073171d57c64dd69f8ed38d5-android-x86.apk)
</details>

src/core/platforms/android/androidplatformutilities.cpp
```diff
@@ -754,13 +754,13 @@ void AndroidPlatformUtilities::vibrate( int milliseconds ) const
 
 void AndroidPlatformUtilities::requestBackgroundPositioningPermissions()
 {
-  checkAndAcquirePermissions( QStringLiteral( "android.permission.ACCESS_BACKGROUND_LOCATION" ) );
+  checkAndAcquirePermissions( QStringList() << QStringLiteral( "android.permission.ACCESS_BACKGROUND_LOCATION" ) );
```
m-kuhn, line 757:
```suggestion
  checkAndAcquirePermissions( { QStringLiteral( "android.permission.ACCESS_BACKGROUND_LOCATION" ) } );
```

Do initializer lists work here?

nirvn: @m-kuhn , let me check.

src/core/positioning/positioningsource.cpp
```diff
@@ -150,9 +150,37 @@ void PositioningSource::setBackgroundMode( bool backgroundMode )
 
   mBackgroundMode = backgroundMode;
 
+  if ( mBackgroundMode )
+  {
+    QFile file( QStringLiteral( "%1.information" ).arg( backgroundFilePath ) );
+    file.open( QFile::WriteOnly );
+    file.close();
```
m-kuhn, line 157: Is this done in order to truncate the file? Should this be done at the end of `getBackgroundPositionInformation` instead? Should `QIODeviceBase::Truncate` be used instead?

src/core/qgsquick/qgsquickcoordinatetransformer.cpp
```diff
@@ -107,11 +107,22 @@ QgsCoordinateTransformContext QgsQuickCoordinateTransformer::transformContext()
   return mCoordinateTransform.context();
 }
 
+QgsPoint QgsQuickCoordinateTransformer::transformPosition( const QgsPoint &position )
```
m-kuhn, line 110: I think this should be a `const` method (and `processPosition` as well)

src/core/qgsquick/qgsquickcoordinatetransformer.cpp
```diff
@@ -242,52 +253,54 @@ void QgsQuickCoordinateTransformer::setVerticalGrid( const QString &grid )
...
+      {
+        // Invalid grid file, skip
+        mVerticalGridPath.clear();
+      }
+      GDALClose( hDataset );
     }
-    GDALClose( hDataset );
   }
+
+  emit verticalGridChanged();
```
m-kuhn, line 305: Random discovery when reviewing this: we currently create a proj object for each call to this, should we rather cache this when the vertical grid is set? https://github.com/opengisch/QField/pull/5917/files?diff=unified&w=1#diff-d7c2354458a68177df2d0253ffddfc748ca33192354477b781dd176e04f1fc37L152-R170

src/core/platforms/android/androidplatformutilities.cpp
nirvn: @m-kuhn , it does work.

src/core/qgsquick/qgsquickcoordinatetransformer.cpp
nirvn: @m-kuhn , interesting, I'm game to do this in a separate PR to reduce the noise here. I also spotted a similar optimization with the tracker I want to do in a separate PR.

werererer:
I highly appreciate you for considering background tracking, acknowledging its importance and actually putting in the work.

However every great progress needs also to be evaluated critically.

Certain types of users will prefer a 100% automatic solution, because regular manual Intervention will intervene in other important tasks.

Two types of users would be affect most positively by automatic tracking.

First, hickers who want to track their trails. They are always overwhelmed by problems, like seeing nature, making manual intervention easily forgotten leading to incomplete data. Missing trail data will make them feel that this app is inadequate even though it easily can be.

Second, people who track themselves long term. While for the first few days excitement over your app exists, long term users might want to track for years. One day they wont be excited anymore about the possibility or tracking and only want their data to work. This leads to manual intervention being forgotten.

If, after a few months or years the user wants to remember an important memory from the past because of lacking automation this data will be Missing. Not only will these users lose trust in the app, they will also have lost a part of important past memories. This can leave a lasting negative impression.

Therefore it is clear that any kind of manual intervention will lead to long term user dissatisfaction.

However I have to admit that i don't have a solution of how to do it exactly in the app.

Thank you again for your effort. I hope my feedback will help you decide the best future for QField.

nirvn:
Interesting lesson learned while debugging an issue reported through Groupement Forestier Quebec: running multiple services on the main process (i.e. ":qt") leads to worlds colliding, and one service just gets pushed out.

In our case, it meant that launching the service that does attachment upload to QFC in the background was failing, leaving a notification with a indefinite progress bar living in the notification panel forever.

PR #5933, opened 2025-01-06
Use visibility signal as driver for focus stack focus

nirvn:
After further debugging, using itemFocusChanged signal is really not ideal (e.g., it throws constant interference when focusing on feature form items!). Instead, we can rely on the visibleChanged signal.

### DIFF
--- a/src/core/focusstack.cpp
+++ b/src/core/focusstack.cpp
@@ -17,8 +17,8 @@
 
 void FocusStack::addFocusTaker( QObject *object )
 {
-  QVariant visible = object->property( "visible" );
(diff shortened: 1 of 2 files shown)


mohsenD98 (approved):
We need to be cautious about having an `item` placed under another `item` with `visibility = true`. Using the `gammaray` application, we can detect if any `item` with `visibility = true` is positioned under another item.

nirvn:
@mohsenD98 it'll only ever be a problem for the key handler item vs welcome screen in our case, which is resolved immediately after we load a project so I I'm confident we're good here.

PR #6005, opened 2025-02-15
Implement Multi-Selection for Local Data Picker Screen

mohsenD98:
**Description:**
In this PR we are implementing multi-selection functionality in the local project and dataset view.

**Changes Implemented:**
1- Selection mode triggers with **long press** on file / folder.
2- It is not possible to trigger selection mode in **root** (Home).   
3- In selection mode, **3 dot menu** for each item is invisible.  
4- In selection mode, back button will turn in to **cancel button** that clear selection.

**Main C++ Changes:**
1- Adding `inSelectionMode` Q_PROPERTY which indicates is in selection mode or not.   
2- Add `ItemCheckedRole`.
3- Add `clearSelection` function that `uncheck` selected mode for all items.  

**Screen cast:**

[Screencast From 2025-02-18 20-50-42.webm](https://github.com/user-attachments/assets/f449d458-14a3-4e86-a6a1-565499e50ebf)

PR #6040, opened 2025-02-25
Upload multiple files to webDav or QFC

mohsenD98:
**Description:**

This pull request implements enhancements based on issue #6005, enabling the ability to push multiple files to both `QFieldCloud` and `WebDAV`.

**Enhancements for Pushing Multiple Files to QFieldCloud:**
We have updated the `addPendingAttachment` function to `addPendingAttachments`, which now accepts `const QStringList &fileNames` instead of `const QString &fileName` as its input. Additionally, we have refactored all instances where this function is used.

**Enhancements for Pushing Multiple Files to WebDAV:**
Similarly, we have revised the `uploadPath` function to `uploadPaths`, which now utilizes `const QStringList &localPaths` instead of `const QString &localPath`. Consequently, we have also refactored all related usages of this method.

**UI Notes:**
1. A menu button will be displayed when the selection mode is active, provided that the selected items are either pushable to QFieldCloud or are configured for WebDAV.
2. We have introduced `determineMenuWidth` to eliminate duplicate code.
3. Once an operation is completed, we will clear the multi-selection for a smoother user experience.

**Demo:**   
  

[Screencast From 2025-03-01 18-51-08.webm](https://github.com/user-attachments/assets/2d16bb61-8f22-4e86-9578-842cbcb87f9d)

### DIFF
--- a/src/core/webdavconnection.cpp
+++ b/src/core/webdavconnection.cpp
@@ -652,70 +652,73 @@ void WebdavConnection::downloadPath( const QString &localPath )
   }
 }
 
-void WebdavConnection::uploadPath( const QString &localPath )
+void WebdavConnection::uploadPaths( const QStringList &localPaths )
 {
-  QFileInfo fi( QDir::cleanPath( localPath ) );
-  QDir dir( fi.isFile() ? fi.absolutePath() : fi.absoluteFilePath() );
-
-  bool webdavConfigurationExists = dir.exists( "qfield_webdav_configuration.json" );
-  QStringList remoteChildrenPath;
-  while ( !webdavConfigurationExists )
+  for ( const QString &localPath : localPaths )
   {
-    remoteChildrenPath.prepend( dir.dirName() );
-    if ( !dir.cdUp() )
-      break;
+    QFileInfo fi( QDir::cleanPath( localPath ) );
+    QDir dir( fi.isFile() ? fi.absolutePath() : fi.absoluteFilePath() );
 
-    webdavConfigurationExists = dir.exists( "qfield_webdav_configuration.json" );
-  }
-
-  if ( webdavConfigurationExists )
-  {
-    QFile webdavConfigurationFile( dir.absolutePath() + QDir::separator() + QStringLiteral( "qfield_webdav_configuration.json" ) );
-    webdavConfigurationFile.open( QFile::ReadOnly );
-    QJsonDocument jsonDocument = QJsonDocument::fromJson( webdavConfigurationFile.readAll() );
-    if ( !jsonDocument.isEmpty() )
+    bool webdavConfigurationExists = dir.exists( "qfield_webdav_configuration.json" );
+    QStringList remoteChildrenPath;
+    while ( !webdavConfigurationExists )
     {
-      QVariantMap webdavConfiguration = jsonDocument.toVariant().toMap();
-      setUrl( webdavConfiguration["url"].toString() );
-      setUsername( webdavConfiguration["username"].toString() );
-      setStorePassword( isPasswordStored() );
+      remoteChildrenPath.prepend( dir.dirName() );
+      if ( !dir.cdUp() )
+        break;
 
-      mProcessRemotePath = webdavConfiguration["remote_path"].toString();
-      if ( !remoteChildrenPath.isEmpty() )
+      webdavConfigurationExists = dir.exists( "qfield_webdav_configuration.json" );
+    }
+
+    if ( webdavConfigurationExists )
+    {
+      QFile webdavConfigurationFile( dir.absolutePath() + QDir::separator() + QStringLiteral( "qfield_webdav_configuration.json" ) );
+      webdavConfigurationFile.open( QFile::ReadOnly );
+      QJsonDocument jsonDocument = QJsonDocument::fromJson( webdavConfigurationFile.readAll() );
+      if ( !jsonDocument.isEmpty() )
       {
-        mProcessRemotePath = mProcessRemotePath + remoteChildrenPath.join( "/" ) + QStringLiteral( "/" );
-      }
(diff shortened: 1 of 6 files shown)


src/qml/QFieldLocalDataPickerScreen.qml
```diff
@@ -808,6 +817,54 @@ Page {
...
+        font: Theme.defaultFont
+        width: parent.width
+        height: enabled ? 48 : 0
+        leftPadding: Theme.menuItemLeftPadding
+
+        text: qsTr("Upload file(s) to WebDAV")
+        onTriggered: {
+          console.log("Send it to web dav!");
+        }
+      }
```
nirvn, line 841: TODO? :) Do you want to do it in this PR or merge this and work on the WebDAV-specific bits in a different PR?

src/qml/QFieldLocalDataPickerScreen.qml
```diff
@@ -808,6 +817,54 @@ Page {
...
+        leftPadding: Theme.menuItemLeftPadding
+
+        text: qsTr("Push file(s) to QFieldCloud")
+        onTriggered: {
+          var fileNames = [];
+          for (let i = 0; i < table.selectedList.length; ++i) {
+            const item = table.itemAtIndex(table.selectedList[i]);
+            const pushableToCloud = item.itemMetaType == LocalFilesModel.Dataset && item.itemType == LocalFilesModel.RasterDataset && cloudProjectsModel.currentProjectId;
+            if (pushableToCloud)
+              fileNames.push(item.itemPath);
```
nirvn, line 861: Just for the sake of readability 

```suggestion
            if (pushableToCloud) {
              fileNames.push(item.itemPath);
            }
```

src/qml/QFieldLocalDataPickerScreen.qml
```diff
@@ -808,6 +817,54 @@ Page {
...
+          }
+          QFieldCloudUtils.addPendingAttachments(cloudProjectsModel.currentProjectId, fileNames);
```
nirvn, line 863: I'd add a fileNames.length > 0 check here , and if we end up with no file to upload (i.e. just folders selected for e.g.), you can throw a warning toaster to users.

mohsenD98: Nice idea, sure :+1:

mohsenD98: what warning text you suggest ?

src/qml/QFieldLocalDataPickerScreen.qml
```diff
@@ -47,6 +47,21 @@ Page {
         parent.finished(false);
       }
     }
+
+    onOpenMenu: {
+      selectionMenu.open();
+    }
+  }
+
+  function determineMenuWidth(menu) {
```
nirvn, line 56: I like the deduplication of code, but not sure I like the fact that it ends as a function -- what about adding a QfMenu and use that code for width, it'd allow us to use it for most of our Menu {} out there.

If you agree, let's defer that as a follow up PR, this can do for now.

src/core/webdavconnection.cpp
```diff
@@ -652,70 +652,73 @@ void WebdavConnection::downloadPath( const QString &localPath )
...
-void WebdavConnection::uploadPath( const QString &localPath )
+void WebdavConnection::uploadPaths( const QStringList &localPaths )
 {
-  QFileInfo fi( QDir::cleanPath( localPath ) );
-  QDir dir( fi.isFile() ? fi.absolutePath() : fi.absoluteFilePath() );
-
-  bool webdavConfigurationExists = dir.exists( "qfield_webdav_configuration.json" );
-  QStringList remoteChildrenPath;
-  while ( !webdavConfigurationExists )
+  for ( const QString &localPath : localPaths )
```
nirvn, line 657: Please remove parts that shouldn't be repeated again and again as you loop through the local paths.

The whole discovery of the qfield_webdav_configuration.json file should be done only one time. Similarly setting the mProcessRemotePath and mProcessLocalPath should be done only once. You'll probably still need to loop through all the files to find the item path that is parent-most  (i.e. take the following list:

- /project/attachments
- /project/my.txt

The mProcessRemotePath and mProcessLocalPath should be /project , not project/attachments

Also, thinking out-loud here, we'll need to detect and bail out when someone provides items from different webdav configuration, that wouldn't work at all ATM.

src/core/webdavconnection.cpp
```diff
@@ -652,70 +652,73 @@ void WebdavConnection::downloadPath( const QString &localPath )
...
+        emit progressChanged();
 
-      mIsUploadingPath = true;
-      emit isUploadingPathChanged();
-      const QUrl url( mUrl );
-      emit confirmationRequested( url.host(), mUsername );
+        mIsUploadingPath = true;
+        emit isUploadingPathChanged();
+        const QUrl url( mUrl );
+        emit confirmationRequested( url.host(), mUsername );
```
nirvn, line 720: All of this ATM is emitted for every item, that's not healthy :)

src/core/webdavconnection.h
```diff
@@ -87,7 +87,7 @@ class WebdavConnection : public QObject
 
     Q_INVOKABLE void importPath( const QString &remotePath, const QString &localPath );
     Q_INVOKABLE void downloadPath( const QString &localPath );
-    Q_INVOKABLE void uploadPath( const QString &localPath );
+    Q_INVOKABLE void uploadPaths( const QStringList &localPaths );
```
nirvn, line 90: Note to self: add documentation! :wink: That's on me, I'll take care of it after we merge this.

nirvn (approved):
Good looking good, and builds tested worked as expected. Good job pushing this through the finish line @mohsenD98 .

PR #6072, opened 2025-03-08
Show QFieldCloud current storage usage

MidnightCoder-m:
**Description:**  
This PR implements a `ProgressBar` that indicates the amount of storage being used in `QFieldCloud`.

---

**UI Notes:**  
- Usage under 90%: Blue color  
- Usage between 90% and 97.5%: Warning (Yellow) color  
- Usage above 97.5%: Red color with a hyperlink that says: “Upgrade to more storage here”  

---

**Upcoming:**  
The API call will be integrated in future PRs, so this feature will remain hidden until those merges are completed.

---

**Glance:**
image

### DIFF
--- a/src/qml/QFieldCloudScreen.qml
+++ b/src/qml/QFieldCloudScreen.qml
@@ -36,15 +36,17 @@ Page {
     Layout.fillHeight: true
     spacing: 2
 
-    RowLayout {
+    Item {
       id: connectionInformation
-      spacing: 2
       Layout.fillWidth: true
+      Layout.minimumHeight: cloudAvatarRect.height + (storageBar.visible * storageBar.height) + 8
+      Layout.leftMargin: 10
+      Layout.rightMargin: 10
+      Layout.topMargin: 12
       visible: cloudConnection.status === QFieldCloudConnection.LoggedIn || table.count > 0
 
       Label {
-        Layout.fillWidth: true
-        padding: 10
+        anchors.left: parent.left
         opacity: projects.visible ? 1 : 0
         text: switch (cloudConnection.status) {
         case 0:
@@ -64,8 +66,7 @@ Page {
 
       Rectangle {
         id: cloudAvatarRect
-        Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter
-        Layout.margins: 10
+        anchors.right: parent.right
         width: 48
         height: 48
         border.color: Theme.mainColor
@@ -120,6 +121,45 @@ Page {
               projects.visible = true;
               refreshProjectsListBtn.forceActiveFocus();
             }
+          }
+        }
+      }
+
+      Label {
+        id: storageText
+        anchors.bottom: storageBar.top
+        anchors.bottomMargin: 4
+        color: Theme.mainTextColor
+        font.italic: true
+        textFormat: Text.RichText
+        onLinkActivated: link => Qt.openUrlExternally(link)
+      }
+
+      ProgressBar {
+        id: storageBar
+        anchors.top: cloudAvatarRect.bottom
+        anchors.topMargin: 8
+        width: parent.width
+        visible: false
+        from: 0
+        to: 1
+        value: usage
+
+        // The `value` property is being animated, so we need the actual value at all times.
+        property double usage: 0
+
+        Material.accent: {
+          if (usage < .9)
+            return Theme.qfieldCloudBlue;
+          else if (usage < .975)
+            return Theme.warningColor;
+          else
+            return Theme.bookmarkRed;
+        }
+
+        Behavior on value  {
+          NumberAnimation {
+            duration: 1000
           }
         }
       }
@@ -620,8 +660,34 @@ Page {
       } else {
         projects.visible = true;
         connectionSettings.visible = false;
+
+        // uncomment when storage bar api is ready
+        // showStorageBar()
       }
+    } else
+    // uncomment when storage bar api is ready
+    // hideStorageBar()
+    {
     }
+  }
+
+  function showStorageBar() {
+    const usedStorage = 1;
+    const totalStorage = 1;
+    storageBar.usage = usedStorage / totalStorage;
+    storageText.text = qsTr(`${usedStorage} GB of ${totalStorage} GB used`);
+    if (storageBar.usage >= .975) {
+      const upgradeStorageText = qsTr("upgrade to more storage here");
+      storageText.text += `; <a href="https://apps.qfield.cloud/plans">${upgradeStorageText}</a>`;
+    }
+    storageText.visible = true;
+    storageBar.visible = true;
+  }
+
+  function hideStorageBar() {
+    storageBar.usage = 0;
+    storageBar.visible = false;
+    storageText.visible = false;
   }
 
   Component.onCompleted: {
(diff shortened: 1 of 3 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for f72294026c500e6b12b4524be5e0de558ee04a80: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6072-eba2024e2d87d32524ee0e10d2b8325e8be43ef2-android-x86.apk)
</details>

src/qml/QFieldCloudScreen.qml
```diff
@@ -123,6 +124,45 @@ Page {
...
+        visible: false
+        from: 0
+        to: 1
+        value: usage
+
+        // The `value` property is being animated, so we need the actual value at all times.
+        property double usage: 0
+
+        Material.accent: {
+          if (usage < .9)
```
suricactus, line 152: btw, is there a shift that we prefer to write the floats without a leading 0? I always find it less readable, but I don't want to force my opinion on the QField devs.

---

One more comment, I would aso add the `{` and `}` around all the blocks that contain JS, personally I find it less error prone. Probably too much python makes me miss them.

---

As for the value itself, I would suggest that it is derived from QFieldCloud, as we want to use a centralized way to inform the users.

In the worst case, I would move it as a constant in the QField source code rather than a hardcoded value.

src/qml/imports/Theme/Theme.qml
```diff
@@ -106,6 +106,7 @@ QtObject {
   property color bookmarkOrange: "orange"
   property color bookmarkRed: "#c0392b"
   property color bookmarkBlue: "#64b5f6"
+  property color qfieldCloudBlue: "#4a6fae"
```
suricactus, line 109: "QFieldCloud" or "qfieldcloud" or "qFieldCloud" or "qfc"? I would vote for either "qfieldcloud" or "qfc".

src/qml/QFieldCloudScreen.qml
mohsenD98: Thanks @suricactus , Personally, I find that when the type is clearly defined as double or float, using either 0 or 0.0 is equally clear. However, to enhance clarity, I will update it :+1: 

also about value itself, its not finished yet, so it will change in upcoming PRs.

src/qml/imports/Theme/Theme.qml
mohsenD98: That's a nice point you've raised. I want to point out that `QFieldCloud` is not possible because properties in QML must start with a lowercase letter. Given that 'qfield' and 'cloud' are two distinct words, I believe `qFieldCloudBlue` is more appropriate for clarity. That said, I’m open to feedback. What do you think, @nirvn ?

suricactus: The main motication to suggest this is that in QFieldSync and QFieldCloud itself we always assume that `qfieldcloud` is a single word and always write it lowercase (unless we can write it the marketing ways of QFieldCloud).

nirvn: Sounds good to me

nirvn:
@suricactus , OK to merge this?

nirvn (approved):

suricactus:
Is there a reason to rush with merging this? There is no API that is stabilized yet on the QFC side and also there might be more info we would like to show? 

Also, there are still 3 occurances of float/doubles that does not have leading 0, which was never used in the codebase before.

My plan is when we finilize the yearly payments in QFC in the next month or so to open the endpoint that serves this data to the client.

nirvn:
_>> Also, there are still 3 occurances of float/doubles that does not have leading 0, which was never used in the codebase before._

That's valid, we need to address this @mohsenD98 .

m-kuhn:
I would vote to show the "upgrade to get more storage" earlier, ideally users will want to upgrade _before_ they run out of space.

nirvn:
We need something like this first: https://github.com/opengisch/QFieldCloud/pull/1099

nirvn:
(oups, ping @m-kuhn for reply above)

mbernasocchi:
From my perspective, the upgrade should appear with the yellow case already, and I think the yellow threshold should be lower. Maybe 80%

suricactus:
Just an update where we are at here. Since we are introducing quite some changes in billing, we better wait for these changes and then expose the API needed for this PR. Once there, I believe the threshold should be a QFieldCloud setting, that is shared across all clients - mobile, web, sync.

m-kuhn:
> From my perspective, the upgrade should appear with the yellow case already, and I think the yellow threshold should be lower. Maybe 80%

Actually I would always show it. We can change the wording from "Manage storage" to "Upgrade to more storage" at 80%.

nirvn:
@m-kuhn , +1 to always showing and tweaking the label.

I can't state how important I feel this improvement is in my mind. We're operating a storage space without a usage meter and no way for people to easily upgrade on either QFS or QF :)

nirvn:
Google threw this at me today, I love it:

image

MidnightCoder-m:
@nirvn, Long time open PR.

nirvn:
This has fallen into merge conflict state. Let's re-open when our fellow QFC devs have had time to develop the needed API.

suricactus:
By the way, think that you have storage that is coming from multiple accounts (shared projects, organizations, etc) so it's different values to show, based on the project. See https://github.com/opengisch/QFieldCloud/pull/1521 for the current implementation, planned to be merged in the next week or so.

PR #6132, opened 2025-03-25
Update qfield_fr.ts - Trad "Not NULL"

AntoElCrackito:
Hello,
For users who speak French and are not geomaticians, it may be confusing to interpret the message of the ‘Not NULL’ constraint, which is why I suggest translating it to the following string: ‘Champs obligatoire’ or mandatory fields.

Ajout de la traduction 'Not NULL' → 'Champs obligatoire'
Add translation 'Not NULL' → 'Champs obligatoire'.

Hoping that this will soon be available in the application, it seems to me that it would be a great help to French users.
Thanks in advance !
(It's my first time using GitHub pull request, I hope I've done this properly).
NotNULLConstraint

### DIFF
--- a/i18n/qfield_fr.ts
+++ b/i18n/qfield_fr.ts
@@ -138,7 +138,7 @@
     </message>
     <message>
         <source>Not NULL</source>
-        <translation>Not NULL</translation>
+        <translation>Champs obligatoire</translation>
     </message>
     <message>
         <source>Unique</source>


nirvn:
@AntoElCrackito , bonjour :) FYI, translations are done on the transifex platform: https://explore.transifex.com/opengisch/qfield-for-qgis/ -- QField's French translation is already at 91%, feel most welcome to chip in!

On the "not null" being translated as "champ obligatoire", it'll be interesting to see if other French translator will agree or have objections. While I can see the logic in using "champ obligatoire", it's not quite what the constraint is stating here. If you consider NULL as a form of value statement, what we're saying here is that the field cannot be set to NULL value :)

In any case, hop over transifex, there are better French speakers that can guide you there :)

AntoElCrackito:
@nirvn, hello, thank you for your feedback :). I'm already involved in translating QField on transifex, but it seems that I haven't found the Not NULL string in the strings to be translated for the app so that's why I've posted this pull request. I also hope that other French contributors will be able to validate this phrase or suggest another, as you can see from the screenshot, this is generally how the fields to be filled in are indicated : 

ChampsObligatoireExemple

Finally, as my English isn't perfect, I've probably misunderstood this part of your post : 

> On the "not null" being translated as "champ obligatoire", it'll be interesting to see if other French translator will agree or have objections. While I can see the logic in using "champ obligatoire", it's not quite what the constraint is stating here. If you consider NULL as a form of value statement, what we're saying here is that the field cannot be set to NULL value :)

In my opinion, the fact that QField App shows in the forms "Not NULL" in the field in this way tells the user that he has to indicate a value because the field can't be empty/NULL, which is why I thought that this translation was relevant. Hope some french speakers will help us, or maybe the string will be posted or validated on transifex !

Thanks !

PR #6153, opened 2025-04-05
Add file content read/write utilities for QML extensions

t-book:
**Problem**
While developing QField extensions, I discovered that the codebase lacks functions to read and write file content directly from QML. This absence makes several common extension scenarios difficult to implement:

- Reading file content to send via POST requests to APIs
- Storing local data when offline
- Implementing custom file-based caching with validation
- Processing local files (CSV, JSON, etc.) directly in QML

After examining FileUtils, PlatformUtilities, and QFieldCloudConnection, I found no existing functions that expose file content reading/writing capabilities to QML.

**Solution**
I've added three utility functions to the FileUtils class:

```
// Read entire file content as a byte array
Q_INVOKABLE static QByteArray readFileContent(const QString &filePath);

// Write content to a file with directory creation and permissions checks
Q_INVOKABLE static bool writeFileContent(const QString &filePath, const QByteArray &content);

// Get comprehensive file info including content, MD5, and metadata 
Q_INVOKABLE static QVariantMap getFileInfo(const QString &filePath);
```
**These functions enable QML extensions to:**

- Read files for processing or uploading
- Write files for local storage
- Get file metadata and content in a single call
- Validate files using MD5 hashes

**Implementation Details**

- The functions feature robust error handling and detailed logging
- Platform-specific considerations are included for Android and iOS security restrictions
- Permissions issues are gracefully handled with informative messages
- Documentation includes guidance on safe file locations

**QML Usage Examples**
Here are some practical examples of how these functions can be used in QML extensions:

```
// Read a file and send it to an API
var fileContent = FileUtils.readFileContent(filePath);
var xhr = new XMLHttpRequest();
xhr.open("POST", "https://api.example.com/upload");
xhr.send(fileContent);

```

**Note:** 
Due to problems with my local build environment, I have not been able to test these functions in a running QField instance. The implementation is based on studying the existing codebase patterns. I would be happy to provide unit tests if needed to ensure the functionality works as expected.
These additions will significantly enhance QField's extensibility without adding much complexity to the codebase.

### DIFF
--- a/src/core/utils/fileutils.cpp
+++ b/src/core/utils/fileutils.cpp
@@ -496,14 +496,28 @@ bool FileUtils::writeFileContent( const QString &filePath, const QByteArray &con
   }
 }
 
-QVariantMap FileUtils::getFileInfo( const QString &filePath )
+QVariantMap FileUtils::getFileInfo( const QString &filePath, bool fetchContent )
 {
   QVariantMap info;
+
+  // Initialize all possible keys with empty defaults
+  info["exists"] = false;
+  info["error"] = "";
+  info["fileName"] = "";
+  info["filePath"] = "";
+  info["fileSize"] = 0;
+  info["lastModified"] = QDateTime();
+  info["suffix"] = "";
+  info["mimeType"] = "";
+  info["md5"] = "";
+  info["md5Error"] = "";
+  info["content"] = QByteArray();
+  info["readable"] = false;
+  info["readError"] = "";
 
   if ( !isWithinProjectDirectory( filePath ) )
   {
     qWarning() << QStringLiteral( "Security warning: Attempted to access file info outside project directory: %1" ).arg( filePath );
-    info["exists"] = false;
     info["error"] = QStringLiteral( "Access denied: File is outside the current project directory" );
     return info;
   }
@@ -536,18 +550,20 @@ QVariantMap FileUtils::getFileInfo( const QString &filePath )
     if ( file.open( QIODevice::ReadOnly ) )
     {
       file.close();
+      info["readable"] = true;
 
-      QByteArray content = readFileContent( filePath );
+      if ( fetchContent )
+      {
+        QByteArray content = readFileContent( filePath );
 
-      if ( !content.isEmpty() || fileInfo.size() == 0 )
-      {
-        info["content"] = content;
-        info["readable"] = true;
-      }
-      else
-      {
-        info["readable"] = false;
-        info["readError"] = QStringLiteral( "Failed to read file content" );
+        if ( !content.isEmpty() || fileInfo.size() == 0 )
+        {
+          info["content"] = content;
+        }
+        else
+        {
+          info["readError"] = QStringLiteral( "Failed to read file content" );
+        }
       }
     }
     else
@@ -560,10 +576,6 @@ QVariantMap FileUtils::getFileInfo( const QString &filePath )
         info["readError"] = info["readError"].toString() + QStringLiteral( " - This may be due to platform security restrictions" );
       }
     }
-  }
-  else
-  {
-    info["exists"] = false;
   }
 
   return info;
--- a/src/core/utils/fileutils.cpp
+++ b/src/core/utils/fileutils.cpp
@@ -26,12 +26,15 @@
 #include <QMimeDatabase>
 #include <QPainter>
 #include <QPainterPath>
+#include <QStandardPaths>
 #include <qgis.h>
 #include <qgsexiftools.h>
 #include <qgsfileutils.h>
+#include <qgsproject.h>
 #include <qgsrendercontext.h>
 #include <qgstextformat.h>
 #include <qgstextrenderer.h>
+
 
 FileUtils::FileUtils( QObject *parent )
   : QObject( parent )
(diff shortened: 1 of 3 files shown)


nirvn:
@t-book , let's see if CI will give us builds :)

Nice contribution BTW!

t-book:
thanks @nirvn I'm just having a look at the formating issues and other failing tests

t-book:
@nirvn added a commit, to hopefully solve the formatting issues. Would you kindly trigger the workflows again?

nirvn:
@t-book , done.

t-book:
Thanks a lot:  @nirvn 

So from what I see Format checks / pre-commit (pull_request) fails because of a whitespace.
image

I would fix this and commit again, okay?

Further MacOS / Create dmg fails because of a 404 not related to my PR? I opened an issue that might solve it. (untested)
https://github.com/opengisch/QField/issues/6156

Thank you for your patience!

nirvn:
@t-book , the code is looking good. One thing though is that we very much need to think of a way to restrict these APIs access to device storages within a reasonable scope that provides enough fencing to avoid non-project data leakage.

On that front, I think we should restrict the paths to PlatformUtilities::applicationDirectory() and additionalApplicationDirectories() while also denying access to the QField subfolder of these paths (where we store additional information including credentials).

So basically this would be permitted:

```
const QStringList allowedPaths = QStringList() << PlatformUtilities::applicationDirectory() << PlatformUtilities::additionalApplicationDirectories();
bool allowed = true;
for ( const QString &allowedPath : allowedPaths )
{
  if ( !filePath.startsWith( PlatformUtilities::applicationDirectory() || filePath.startsWith(  QStringLiteral( "%1/QField" ) ) )
  {
    allowed = false;
    break;
}

if ( !allowed )
  return;
```

That should be a whitelist check applied to all three functions here (and we probably want to use the QDir's cleanPath to aovid people working around the restrain by adding double slashes, etc.)

With that safeguard in, I think we'd be good to go. @m-kuhn , do you have any additional thoughts on the safety of these functions?

m-kuhn:
For proper safeguards we'll need to do a bit more than that (to avoid patterns like `applicationDirectory/../../../protectedDir`)  and make sure we always compare canonical filepaths.

nirvn:
@m-kuhn , good thinking.

t-book:
@nirvn @m-kuhn thanks a lot for your feedback!

I completely agree with implementing path restrictions to prevent access outside permitted application directories. IOS and Android should prevent this out of the box? … but other environments like Windows etc will not …
I'm happy to update the PR (will take a moment)!

nirvn:
@t-book , please implement the logic across all platforms, inc. windows / macos / linux. It'll insure plugin authors can better understand restrictions across the board.

nirvn:
Sounds good to me, I'd rather start with a really narrow accepted read/write access and expand in the future if that's ever needed.

t-book:
Thanks a lot @nirvn (again)!
As this addition is crucial for security, I would try to add a unit test so that we can clearly see what is expected. Being new to the project, is this the correct file? https://github.com/opengisch/QField/blob/master/test/test_featureutils.cpp and are tests currently used?

nirvn:
@t-book , testing most welcome; I'd add it in here: https://github.com/opengisch/QField/blob/master/test/test_fileutils.cpp

t-book:
Good evening @nirvn,
I've marked the PR as a Draft temporarily while I work on the final touches. The safeguard is now in place, and I’ve added some very simple [test cases](https://github.com/opengisch/QField/pull/6153/files#diff-a110cad3a14fbd8bf35ecfd3da72c9b99dbefdd0d3a65f7047d4355ef30f01bc):

I would wait now until https://github.com/opengisch/QField/pull/6165 is merged. Doing so we should get an OSX build again and I can test with the created artifact. Also I would like to add some more checks for isWithinProjectDirectory() regarding common attack patterns. As I'm having problems builing locally and developing is currently a bit of a blindflight, do you know if by chance some docker image or VM exists that has a developent environment ready?

nirvn:
@t-book , artifact created.

t-book:
@nirvn This turned out to be trickier than expected — I can now confidently say: I hate Windows separators 😄

I’ve added several tests for the new functions and verified reading, writing, and security cases with a quick and [dirty test plugin](https://gist.github.com/t-book/01ff93049177717fb4791811e3bbd5d9). It demonstrates the expected behavior using the latest Linux build artifact:

2025-04-10 12 20 28 PM

I’m removing the draft status — from my side, everything looks good now, builds finish successfull and format checks pass.
The only odd thing: when downloading the macOS build artifact, the system reports the package as broken. I’m quite sure this isn’t related to this PR (or hopefully not to [#6165](https://github.com/opengisch/QField/pull/6165) either). But it’s worth a closer look. If it’s not expected behavior or there for longer, I’ll open a separate issue.

Would be happy about your review!

src/core/utils/fileutils.cpp
```diff
@@ -318,3 +321,250 @@ void FileUtils::addImageStamp( const QString &imagePath, const QString &text )
...
+#ifdef Q_OS_ANDROID
+  QString appStorage = QStandardPaths::writableLocation( QStandardPaths::AppDataLocation );
+  isLikelyWritable = filePath.startsWith( appStorage );
+#elif defined( Q_OS_IOS )
+  QString appStorage = QStandardPaths::writableLocation( QStandardPaths::AppDataLocation );
+  isLikelyWritable = filePath.startsWith( appStorage );
+#else
+  QFileInfo dirInfo( directory.absolutePath() );
+  isLikelyWritable = dirInfo.isWritable();
+#endif
```
nirvn, line 458: Do we need this? What would be the scenario(s) where we would be within a project directory but _not_ able to write?

src/core/utils/fileutils.cpp
```diff
@@ -318,3 +321,250 @@ void FileUtils::addImageStamp( const QString &imagePath, const QString &text )
...
+    }
+    else
+    {
+      info["readable"] = false;
+      info["readError"] = file.errorString();
+
+      if ( file.error() == QFile::PermissionsError )
+      {
+        info["readError"] = info["readError"].toString() + QStringLiteral( " - This may be due to platform security restrictions" );
+      }
```
nirvn, line 561: Could we add a bool fetchContent parameter (false by default) to this function? I feel like there are quite a few scenarios when we won't want to get the content at all.

src/core/utils/fileutils.cpp
```diff
@@ -318,3 +321,250 @@ void FileUtils::addImageStamp( const QString &imagePath, const QString &text )
...
+
+QVariantMap FileUtils::getFileInfo( const QString &filePath )
+{
+  QVariantMap info;
+
+  if ( !isWithinProjectDirectory( filePath ) )
+  {
+    qWarning() << QStringLiteral( "Security warning: Attempted to access file info outside project directory: %1" ).arg( filePath );
+    info["exists"] = false;
+    info["error"] = QStringLiteral( "Access denied: File is outside the current project directory" );
```
nirvn, line 507: Are you deliberately not adding an "error" key when no error to use:

if info.error != undefined { ... } 

?

Asking as I'm wondering whether we should insure that all keys are always available (and when there are no errors, these error keys have empty string values).

src/core/utils/fileutils.cpp
```diff
@@ -318,3 +321,250 @@ void FileUtils::addImageStamp( const QString &imagePath, const QString &text )
...
+    info["mimeType"] = db.mimeTypeForFile( filePath ).name();
+
+    QByteArray md5Hash = fileChecksum( filePath, QCryptographicHash::Md5 );
+    if ( !md5Hash.isEmpty() )
+    {
+      info["md5"] = md5Hash.toHex();
+    }
+    else
+    {
+      info["md5Error"] = QStringLiteral( "Could not calculate MD5 hash - possibly due to permission restrictions" );
```
nirvn, line 533: Same as above. I am wondering if we should insure all keys always have associated values.

nirvn (commented):
Only a few comments. Thanks for this, I'm sure it'll be wildly used in the future :)

t-book:
Great thanks a lot @nirvn I will comment under your comments and push a comment that respects your ideas.

src/core/utils/fileutils.cpp
t-book: The idea here was to early catch situation like Permission issues where the project directory is for example on a read-only file system … or disk full errors … chmod permissions on linux etc … but I can definitly remove this check if this makes the code more clear.

src/core/utils/fileutils.cpp
t-book: Very good! Will add it!

src/core/utils/fileutils.cpp
t-book: Will respect it in the following commit. Will take a moment as I first check if tests pass before pushing to upstream. Thanks @nirvn !

t-book:
@nirvn 


- Added a new fetchContent parameter (default: false) to make content retrieval optional, improving performance for cases where only metadata is needed
- Ensured all keys in the returned QVariantMap are always present with default values for a more consistent API
- Added empty default values for error-related keys to simplify client code (especially QML) by eliminating the need for key existence checks
- Updated documentation to reflect the new parameter and behavior in h file
- removed additional check if project directory is writeable

nirvn (approved):

t-book:
@nirvn Thanks for your approval! Just checking—are we ready to merge?

nirvn:
@t-book , done! :) thanks for your contribution, looking forward to plenty more :wink:

t-book:
Perfect ☀️, thanks a you su much @nirvn!
> Looking forward to plenty more.

I’ll do my best to look into some bug reports from the GitHub issues. :)

Quick question—do devs have a place where they mingle, like on @gitter or maybe a mailing list?

PR #6158, opened 2025-04-07
Refresh and Update README.md

mbernasocchi:
- added new screnshot
- link to DPG
- link to releases page
- link to QFieldCloud and its repo
- formatt download links to look nicer

PR #6160, opened 2025-04-07
update and refresh readme

mbernasocchi:
added new screnshot
link to DPG
link to releases page
link to QFieldCloud and its repo
formatt download links to look nicer


mbernasocchi:
same submodule error as #6158

PR #6168, opened 2025-04-09
Cloud project's localized datasets support

nirvn:
This PR adds logic to support downloading of localized datasets from a special localized datasets project hosted on QFieldCloud.

### DIFF
--- a/src/core/qfieldcloudprojectsmodel.cpp
+++ b/src/core/qfieldcloudprojectsmodel.cpp
@@ -1708,6 +1708,7 @@ void QFieldCloudProjectsModel::projectListReceived()
     beginResetModel();
     mProjects.clear();
     endResetModel();
+    mLocalizedDatasets.clear();
   }
 
   QByteArray response = rawReply->readAll();
@@ -2141,6 +2142,7 @@ void QFieldCloudProjectsModel::loadProjects( const QJsonArray &remoteProjects, b
     if ( projectDetails.value( "name" ) == QStringLiteral( "localized_datasets" ) )
     {
       // Protected name, skip
+      mLocalizedDatasets[projectDetails.value( "owner" ).toString()] = projectDetails.value( "id" ).toString();
       continue;
     }
 
--- a/src/core/qfieldcloudprojectsmodel.cpp
+++ b/src/core/qfieldcloudprojectsmodel.cpp
@@ -1810,7 +1810,7 @@ NetworkReply *QFieldCloudProjectsModel::downloadFile( const QString &projectId, 
   request.setAttribute( QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::RedirectPolicy::UserVerifiedRedirectPolicy );
   mCloudConnection->setAuthenticationDetails( request );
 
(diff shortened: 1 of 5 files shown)


src/core/qfieldcloudprojectsmodel.cpp
```diff
@@ -2170,6 +2252,16 @@ void QFieldCloudProjectsModel::loadProjects( const QJsonArray &remoteProjects, b
       }
     }
 
+    if ( projectDetails.contains( "localized_datasets" ) )
+    {
+      const QVariantList localizedDatasets = projectDetails.value( "localized_datasets" ).toList();
+      for ( const QVariant &localizedDataset : localizedDatasets )
+      {
+        const QVariantMap localizedDatasetDetails = localizedDataset.toMap();
+        cloudProject->localizedDatasets << localizedDatasetDetails.value( "datasource" ).toString().mid( 10 );
```
suricactus, line 2261: I do not follow this bit here? You skip the first 10 chars because they are `localized:` I guess?

I guess a comment can be helpful here.

Docs reference: https://doc.qt.io/qt-6/qstring.html#mid

src/core/qfieldcloudprojectsmodel.cpp
```diff
@@ -1730,16 +1804,16 @@ void QFieldCloudProjectsModel::projectListReceived()
   }
 }
 
-NetworkReply *QFieldCloudProjectsModel::downloadFile( const QString &projectId, const QString &fileName )
+NetworkReply *QFieldCloudProjectsModel::downloadFile( const QString &projectId, const QString &fileName, bool fromPackage )
```
suricactus, line 1807:
```suggestion
NetworkReply *QFieldCloudProjectsModel::downloadFile( const QString &projectId, const QString &fileName, const QString &packageId )
```

Note one can download from multiple packages. This soon will be the case when we support multiple password credentials and therefore packages for endusers. So values could be:

- `fromPackage.isEmpty()` - download from standard files
- `! QUuid.fromString( fromPackage ).isNull()` - download from given package
- `fromPackage == QStringLiteral( "latest" )` - download from latest
- else - panic!

If you want to keep the boolean logic for now, rename it to `bool fromLatestPackage`.

src/core/qfieldcloudprojectsmodel.cpp
```diff
@@ -1081,7 +1081,15 @@ void QFieldCloudProjectsModel::projectDownload( const QString &projectId )
       }
       else
       {
-        if ( !layer.value( QStringLiteral( "is_valid" ) ).toBool() && !layer.value( QStringLiteral( "is_localized" ) ).toBool() )
+        if ( layer.value( QStringLiteral( "is_localized" ) ).toBool() )
+        {
+          const QString localizedDataset = layer.value( "datasource" ).toString().mid( 10 );
```
suricactus, line 1086: I think it would be even beter if the cloud returns you the proper localized filename within the project, so there is only one place that calculates this.

nirvn: @suricactus , maybe -- see https://github.com/opengisch/QFieldCloud/pull/1207

src/core/qfieldcloudprojectsmodel.cpp
nirvn: See comment below

mohsenD98 (approved):

PR #6249, opened 2025-05-09
make QFC standout

mbernasocchi:
image


mbernasocchi:
@anjakyra the png would need to be regenerated from the SVG, I just color replaced the png directly

anjakyra:
Nice. I think it looks good

PR #6267, opened 2025-05-16
gh pages revision

anjakyra:
- homepage: qfieldcloud sync icon png swapped with svg
- carousel on homepage: stays the same size but I repositioned some pictures (some are already cropped so no possibility to change the position)
- carousel on other pages: smaller and I repositioned some pictures


nirvn:
@mbernasocchi , seems like we need to add .DS_STORE in the list of ignored files :)

PR #6302, opened 2025-05-26
Implement dynamic sorting of QFieldCloud projects list

nirvn:
Instead of using the display role to sort the cloud projects list, let's do it via c++ code, allows us to then implement a richer sorting (for e.g., featured projects first? :wink: )


mbernasocchi (approved):
All works from functional side

PR #6343, opened 2025-06-03
Add vendor command support

messplay:


nirvn:
@messplay , this feels like AI generated, it wouldn't work in practice, and its parts are not integrated (e.g. the UI additions are not exposed, the positioning device invokable functions not reachable, etc.).  On top of which, there's no description, context, etc. 

Did something go wrong here?

nirvn:
@messplay , re-reading myself and hoping I didn't come through as too harsh :) we can discuss this further in an enhancement request on a draft PR (with code that'll compile) if you are interested in pushing this functionality forward.

PR #6378, opened 2025-06-11
Navigation → Azimuth : Printing a positive value when bearing > 180°

AntoElCrackito:
Added line 295 to force values returned negative in the range 180.1° to 360° to appear positive (modulo of the result forcing it to be in the range 0° to 360°).

Example : If the value returned is -45°, it's 315° which is returned. 

For users, this more intuitive and similar to a compass use.


nirvn:
@AntoElCrackito , nice papercut fix, thanks!

nirvn:
I'd like to get this in the next point release, so I went ahead and fixed the proposed fix here: https://github.com/opengisch/QField/pull/6399. Thanks for your contribution!

PR #6436, opened 2025-07-03
Improve carrousel looks on mobile devices

nirvn:
Before (left) vs. PR right:

image

image

image

On desktop screens, the look remain identical.


anjakyra:
Thank you @nirvn! it looks nice

PR #6502, opened 2025-07-29
improve correctness of GNSS datetime handling

edgecase14:
This prevents initial positioning window not being greyed out even when there has never been a GNSS fix.

This also avoids pretending the system time is the GNSS time in EXIF metadata.  That could potentially break someone's batch image processing, so maybe add to release notes.

### DIFF
--- a/src/core/positioning/internalgnssreceiver.cpp
+++ b/src/core/positioning/internalgnssreceiver.cpp
@@ -177,7 +177,9 @@ void InternalGnssReceiver::handleSatellitesInUseUpdated( const QList<QGeoSatelli
                                                               mSatellitesInfo, 0, 0, 0,
                                                               mLastGnssPositionInformation.hacc(),
                                                               mLastGnssPositionInformation.vacc(),
-                                                              mLastGnssPositionInformation.utcDateTime(),
+                                                              mLastGnssPositionInformation.utcDateTime().isValid() ? 
+                                                                mLastGnssPositionInformation.utcDateTime() : 
+                                                                QDateTime::currentDateTimeUtc(),
                                                               QChar(), 0, -1, static_cast<int>( mSatellitesID.size() ), QChar( 'A' ), mSatellitesID, mSatelliteInformationValid,
                                                               mLastGnssPositionInformation.verticalSpeed(),
                                                               mLastGnssPositionInformation.magneticVariation(),
--- a/src/core/utils/fileutils.cpp
+++ b/src/core/utils/fileutils.cpp
@@ -373,8 +373,11 @@ void FileUtils::addImageMetadata( const QString &imagePath, const GnssPositionIn
     metadata["Exif.GPSInfo.GPSSpeedRef"] = "K";
   }
 
-  metadata["Exif.GPSInfo.GPSDateStamp"] = positionInformation.utcDateTime().date();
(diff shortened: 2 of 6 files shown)


edgecase14:
Additional comments in issue #6571

gdt:
I'm an outsider, but were I reviewing for merge i'd want to see the multiple commits rebased into tthe series of commits you would have done had you done it perfectly the first time.    Not a criticism of getting to working incrementlaly, <metoo>, but On projects I maintain I prefer to merge clean commits, vs squashing things that don't necessarily go together.  I don't know how qfield feels.

src/core/utils/fileutils.cpp
```diff
@@ -373,8 +373,12 @@ void FileUtils::addImageMetadata( const QString &imagePath, const GnssPositionIn
     metadata["Exif.GPSInfo.GPSSpeedRef"] = "K";
   }
 
-  metadata["Exif.GPSInfo.GPSDateStamp"] = positionInformation.utcDateTime().date();
-  metadata["Exif.GPSInfo.GPSTimeStamp"] = positionInformation.utcDateTime().time();
+  // slight change in behaviour, but more correct
```
gdt, line 376: I prefer to see changes in commit messages, and comments in the code describe how it is.   I find "slight change in behavior" unhelpful, and it will not age well.   Change from when?  Why?  What actual change?

src/core/utils/fileutils.cpp
```diff
@@ -373,8 +373,12 @@ void FileUtils::addImageMetadata( const QString &imagePath, const GnssPositionIn
...
+  if ( positionInformation.utcDateTime().isValid() )
+  {
+    metadata["Exif.GPSInfo.GPSDateStamp"] = positionInformation.utcDateTime().date();
+    metadata["Exif.GPSInfo.GPSTimeStamp"] = positionInformation.utcDateTime().time();
+  }
```
gdt, line 381: I see what you're doing, but elsewhere you are using system time instead of gnss time if gnss time is not valid.   Why is this different?  I'm not saying you don't have a reason  but that reason is what should be in the comment.

Are you finding that there is a valid position but not valid time in gnss data?

src/core/positioning/internalgnssreceiver.cpp
```diff
@@ -177,7 +177,7 @@ void InternalGnssReceiver::handleSatellitesInUseUpdated( const QList<QGeoSatelli
                                                               mSatellitesInfo, 0, 0, 0,
                                                               mLastGnssPositionInformation.hacc(),
                                                               mLastGnssPositionInformation.vacc(),
-                                                              mLastGnssPositionInformation.utcDateTime(),
+                                                              mLastGnssPositionInformation.utcDateTime().isValid() ? mLastGnssPositionInformation.utcDateTime() : QDateTime::currentDateTimeUtc(),
```
gdt, line 180: This code is repeated a bunch of times.  That makes me wonder:
  - should it be a procedure?
  - why check if datetime is not valid, but use hacc/vacc unconditionally?   Is there really a situation with vllid accuracy but not valid time?

It seems clear something tricky is going on, but I don't understand it from the comments you didn't add, and I don't understand it from the commit message.     Maybe GNSS data should not be used at all if the associated datetime is not valid, and that's the real bug.

nirvn:
This has felt into merge conflict state. I'll close this to clean up the PR queue, feel most invited to re-open when you have additional time to address comments @edgecase14 .

PR #6573, opened 2025-08-15
add headings to Positioning Settings

edgecase14:
I added headings and tried to re-group the existing items similar to the General settings tab.

The purpose is to make it easier to find things, and also to prepare for a group of settings for my upcoming NTRIP client, and other planned features like the setting of CRS for the GNSS receiver.


edgecase14:
oh and if there is a better place to discuss UI changes, please let me know.

the "Enforce Accuracy Requirement" option still gets hidden based on "Show Position Information", which is a bit wierd.

image

nirvn:
@edgecase14 , nice improvement. Let's keep the [x] enforce accuracy requirement as a sub option to accuracy indicator. It's not ideal but it's much better than having it disconnected in a different section where its visibility is harder to understand.

We'd also need a header for the positioning device here, otherwise it just looks odd. The left/right margins of that section also needs adjusting to match the rest of the elements on that page.

nirvn:
Since we've had quite a few UI tweaks prior to the next QField update, I'll close this for now.

PR #6575, opened 2025-08-15
add GNSS menu button and Pos. Info panel to Welcome Screen

edgecase14:
This was inspired by having GNSS receivers that required a lot of fiddling around, combined with providing phone support to field workers.  The premise is to make application launch match our physical workflow:

1) office GIS staff prepare your backpack, test GNSS receiver with a quick power up, launch QFIELD, and immediately see system status.
2) field staff put on bug spray, test GNSS receiver and Android tablet by launching QFIELD and see system status immediately, walk 1km (with limited cell phone coverage in many cases), start collecting data.  
3) office staff doing equipment maintenance don't need to wait for QfieldCloud menus to load

I feel most people with external GPS receiver would have a similar experience, and appreciate having GPS info quick at hand.

image


nirvn:
@edgecase14 , hey there -- let's close this PR in favor of quickstarting a discussion in an enhancement request / issue instead? As it stands, I think the idea doesn't integrate super well with the welcome screen, which we're keen to keep lean :) 

I'm happy to brainstorm with you on what scenarios would require early positioning details, and how we could integrate that. One alternative for e.g. would be to add some details in the settings' positioning tab to help configuring GNSS devices.

edgecase14:
Where should the discussion happen?  The Positioning Info part is an option, only the GNSS button is an imposition.  I see a struggle between pretty UI and functional for busy people, I'd be happy to compromise by perhaps hiding that too behind a UI configuration option?  But the number of options starts heading to infinity.  Really all the conflicting requests could use a QML UI design tool, each user builds their own...

edgecase14:
Oh and one of our staff I showed the screenshot to was excited... perhaps because they're the one who sets up QFieldCloud projects for the rest of the company, packs up the equipment, day after day.

PR #6588, opened 2025-08-20
QField / i18n - ptBR translations

eduardokanegae:
+140 strings translated to pt_BR - those tagged with *unfinished*


nirvn:
@eduardokanegae , this is a fantastic amount of work, thanks for this! 

We do not use github to manage translation but rather transifex. You can see QField's page here: https://explore.transifex.com/opengisch/qfield-for-qgis/ -- this is where you can join in and add your super nice effort towards getting 100% translation status :)

eduardokanegae:
sorry. Just uploaded it at transifex

PR #6681, opened 2025-09-30
Gut project *out* of the delta file wrapper code to allow for some good stuff to come

nirvn:
First step to allow for each QFieldCloudProject to have a delta file wrapper member and unlock push changes / synchronization of cloud projects _other than_ the one currently opened.

### DIFF
--- a/src/core/qfieldcloud/deltafilewrapper.cpp
+++ b/src/core/qfieldcloud/deltafilewrapper.cpp
@@ -423,7 +423,6 @@ QMap<QString, QString> DeltaFileWrapper::attachmentFileNames() const
   for ( const QJsonValue &deltaJson : std::as_const( mDeltas ) )
   {
     QVariantMap delta = deltaJson.toObject().toVariantMap();
-    qDebug() << delta;
     const QString localLayerId = delta.value( QStringLiteral( "localLayerId" ) ).toString();
     const QString method = delta.value( QStringLiteral( "method" ) ).toString();
     const QString localPk = delta.value( QStringLiteral( "localPk" ) ).toString();
--- a/src/core/qfieldcloud/deltafilewrapper.cpp
+++ b/src/core/qfieldcloud/deltafilewrapper.cpp
@@ -41,8 +41,7 @@ Q_GLOBAL_STATIC( CacheAttachmentFieldNamesMap, sCacheAttachmentFieldNames );
 Q_GLOBAL_STATIC( QSet<QString>, sFileLocks );
 
 
-DeltaFileWrapper::DeltaFileWrapper( const QgsProject *project, const QString &fileName )
-  : mProject( project )
+DeltaFileWrapper::DeltaFileWrapper( const QString &projectId, const QString &fileName )
 {
   QFileInfo fileInfo = QFileInfo( fileName );
 
@@ -61,7 +60,7 @@ DeltaFileWrapper::DeltaFileWrapper( const QgsProject *project, const QString &fi
 #endif
 
   if ( mErrorType == DeltaFileWrapper::ErrorType::NoError )
-    mCloudProjectId = QFieldCloudUtils::getProjectId( mProject->fileName() );
+    mCloudProjectId = projectId;
 
   if ( mErrorType == DeltaFileWrapper::ErrorType::NoError && mCloudProjectId.isNull() )
     setError( DeltaFileWrapper::ErrorType::NotCloudProjectError );
@@ -424,10 +423,10 @@ QMap<QString, QString> DeltaFileWrapper::attachmentFileNames() const
   for ( const QJsonValue &deltaJson : std::as_const( mDeltas ) )
   {
     QVariantMap delta = deltaJson.toObject().toVariantMap();
+    qDebug() << delta;
     const QString localLayerId = delta.value( QStringLiteral( "localLayerId" ) ).toString();
     const QString method = delta.value( QStringLiteral( "method" ) ).toString();
     const QString localPk = delta.value( QStringLiteral( "localPk" ) ).toString();
-    const QStringList attachmentFieldNamesList = attachmentFieldNames( mProject, localLayerId );
 
     if ( method == QStringLiteral( "delete" ) || method == QStringLiteral( "patch" ) )
     {
@@ -453,12 +452,12 @@ QMap<QString, QString> DeltaFileWrapper::attachmentFileNames() const
 
         for ( const QString &fieldName : attributeNames )
         {
-          if ( !attachmentFieldNamesList.contains( fieldName ) )
-            continue;
-
           const QString fileName = attributes.value( fieldName ).toString();
-
-          Q_ASSERT( filesChecksum.contains( fileName ) );
+          if ( !filesChecksum.contains( fileName ) )
+          {
+            // Not a file attachment, skip ahead
+            continue;
+          }
 
           const QString key = QStringLiteral( "%1//%2//%3" ).arg( localLayerId, localPk, fieldName );
           const QString fileChecksum = filesChecksum.value( fileName ).toString();
@@ -487,7 +486,7 @@ QMap<QString, QString> DeltaFileWrapper::attachmentFileNames() const
   return fileNameChecksum;
 }
 
-void DeltaFileWrapper::addPatch( const QString &localLayerId, const QString &sourceLayerId, const QString &localPkAttrName, const QString &sourcePkAttrName, const QgsFeature &oldFeature, const QgsFeature &newFeature, bool storeSnapshot )
+void DeltaFileWrapper::addPatch( const QgsProject *project, const QString &localLayerId, const QString &sourceLayerId, const QString &localPkAttrName, const QString &sourcePkAttrName, const QgsFeature &oldFeature, const QgsFeature &newFeature, bool storeSnapshot )
 {
   QVariant localPk = oldFeature.attribute( localPkAttrName );
   QVariant sourcePk = oldFeature.attribute( sourcePkAttrName );
@@ -495,8 +494,8 @@ void DeltaFileWrapper::addPatch( const QString &localLayerId, const QString &sou
     {
       { "localPk", QgsVariantUtils::isNull( localPk ) ? QString() : localPk.toString() },
       { "localLayerId", localLayerId },
-      { "localLayerCrs", crsByLayerId( mProject, localLayerId ) },
-      { "localLayerName", nameByLayerId( mProject, localLayerId ) },
+      { "localLayerCrs", crsByLayerId( project, localLayerId ) },
+      { "localLayerName", nameByLayerId( project, localLayerId ) },
       { "method", "patch" },
       { "sourcePk", QgsVariantUtils::isNull( sourcePk ) ? QString() : sourcePk.toString() },
       { "sourceLayerId", sourceLayerId },
@@ -624,7 +623,7 @@ void DeltaFileWrapper::addPatch( const QString &localLayerId, const QString &sou
 
     QJsonObject oldFileChecksums;
     QJsonObject newFileChecksums;
-    std::tie( newFileChecksums, oldFileChecksums ) = addAttachments( localLayerId, tmpNewAttrs, tmpOldAttrs );
+    std::tie( newFileChecksums, oldFileChecksums ) = addAttachments( project, localLayerId, tmpNewAttrs, tmpOldAttrs );
     if ( !oldFileChecksums.isEmpty() )
     {
       oldData.insert( QStringLiteral( "files_sha256" ), oldFileChecksums );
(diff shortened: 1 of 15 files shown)


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for e6facafd65b170f8e2d0c8e9c9185725a7b96b00: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6681-03cb778680575204ecd19b543e90f69343640257-android-x86.apk)
</details>

MidnightCoder-m:
Nice work @nirvn , Thanks!

src/core/qfieldcloud/deltafilewrapper.cpp
```diff
@@ -424,10 +423,10 @@ QMap<QString, QString> DeltaFileWrapper::attachmentFileNames() const
   for ( const QJsonValue &deltaJson : std::as_const( mDeltas ) )
   {
     QVariantMap delta = deltaJson.toObject().toVariantMap();
+    qDebug() << delta;
```
suricactus: qDebug()

src/core/qfieldcloud/deltafilewrapper.cpp
```diff
@@ -424,10 +423,10 @@ QMap<QString, QString> DeltaFileWrapper::attachmentFileNames() const
...
     const QString localLayerId = delta.value( QStringLiteral( "localLayerId" ) ).toString();
     const QString method = delta.value( QStringLiteral( "method" ) ).toString();
     const QString localPk = delta.value( QStringLiteral( "localPk" ) ).toString();
-    const QStringList attachmentFieldNamesList = attachmentFieldNames( mProject, localLayerId );
```
suricactus, line 430: can you elaborate why we removed `attachmentFieldNamesList` here?  We added it as a double assertion that we check the right thing and more.

test/test_digitizinglogger.cpp
suricactus, line 1: LOL, you are removing my author headers even though  I did not have a single line in that file  😆

src/qml/LocatorSettings.qml
```diff
@@ -16,7 +16,7 @@ Popup {
   width: Math.min(400, mainWindow.width - Theme.popupScreenEdgeMargin)
   x: (parent.width - width) / 2
   y: (parent.height - height) / 2
-  padding: 0
+  padding: 10
```
suricactus: Is this by accident? I don't see it related to the PR.

src/core/qfieldcloud/qfieldcloudprojectsmodel.cpp
```diff
@@ -864,7 +866,7 @@ QVariant QFieldCloudProjectsModel::data( const QModelIndex &index, int role ) co
       return project->deltaFilePushStatusString();
 
     case LocalDeltasCountRole:
-      return project->deltasCount();
+      return project->deltaFileWrapper() ? project->deltaFileWrapper()->count() : 0;
```
suricactus, line 869: I see you use this extensively in QML and also here, probably "hiding" the logic behind `deltasCount` proxy method on the project/layerObserver is not a terrible idea to reduce checks in the code which IMO are error prone.


int XXX::deltasCount() const
{
    return project->deltaFileWrapper() ? project->deltaFileWrapper()->count() : 0;
}

src/core/qfieldcloud/qfieldcloudproject.cpp
```diff
@@ -2220,7 +2219,12 @@ void QFieldCloudProject::restoreLocalSettings( QFieldCloudProject *project, cons
   // generate local export id if not present. Possible reasons for missing localExportId are:
   // - the cloud project download aborted halfway
   // - the local settings were somehow deleted, but not the project itself (unlikely)
-  if ( project->lastLocalExportId().isEmpty() )
+  if ( !project->lastLocalExportId().isEmpty() )
```
suricactus, line 2226: Can you switch the if and else parts here for better readability? I think having `if (!true) { //1 } else { //2 }` is harder to read than `if (true) { //2 } else { //1 }`.

src/core/qfieldcloud/qfieldcloudproject.cpp
```diff
@@ -1693,12 +1688,12 @@ void QFieldCloudProject::push( LayerObserver *layerObserver, bool shouldDownload
...
-        deltaFileWrapper->resetId();
-        deltaFileWrapper->setIsPushing( false );
+        mDeltaFileWrapper->reset();
+        mDeltaFileWrapper->resetId();
+        mDeltaFileWrapper->setIsPushing( false );
 
-        if ( !deltaFileWrapper->toFile() )
-          QgsMessageLog::logMessage( QStringLiteral( "Failed to reset delta file. %1" ).arg( deltaFileWrapper->errorString() ) );
+        if ( !mDeltaFileWrapper->toFile() )
+          QgsMessageLog::logMessage( QStringLiteral( "Failed to reset delta file. %1" ).arg( mDeltaFileWrapper->errorString() ) );
```
suricactus, line 1699:
```suggestion
        if ( !mDeltaFileWrapper->toFile() )
        {
          QgsMessageLog::logMessage( QStringLiteral( "Failed to reset delta file. %1" ).arg( mDeltaFileWrapper->errorString() ) );
        }
```

src/core/qfieldcloud/qfieldcloudproject.cpp
```diff
@@ -1679,10 +1674,10 @@ void QFieldCloudProject::push( LayerObserver *layerObserver, bool shouldDownload
...
       case DeltaErrorStatus:
         delete networkDeltaStatusCheckedParent;
-        deltaFileWrapper->resetId();
-        deltaFileWrapper->setIsPushing( false );
+        mDeltaFileWrapper->resetId();
+        mDeltaFileWrapper->setIsPushing( false );
 
-        if ( !deltaFileWrapper->toFile() )
+        if ( !mDeltaFileWrapper->toFile() )
           QgsMessageLog::logMessage( QStringLiteral( "Failed update committed delta file." ) );
```
suricactus, line 1682:
```suggestion
        if ( !mDeltaFileWrapper->toFile() )
        {
          QgsMessageLog::logMessage( QStringLiteral( "Failed update committed delta file." ) );
        }
```

src/core/qfieldcloud/deltafilewrapper.cpp
```diff
@@ -1025,17 +1038,69 @@ void DeltaFileWrapper::mergePatchDelta( const QJsonObject &delta )
       }
     }
 
-    // add the checksums of the files in the current and old delta combined
-    QJsonObject oldFileChecksums;
-    QJsonObject newFileChecksums;
-    std::tie( newFileChecksums, oldFileChecksums ) = addAttachments( localLayerId, existingDeltaNewAttrs, existingDeltaOldAttrs );
-    if ( !oldFileChecksums.isEmpty() )
```
suricactus, line 1032: Can you add a comment what is going on from these lines until the end of the function? If I get it right, it does the same as before, but it is 3 times the code so I might be missing something.

src/core/qfieldcloud/deltafilewrapper.cpp
nirvn: @suricactus , we already did this check in the addAttachments() function when adding a delta when triggered by a project vector layer signal. Duplicating the check in DeltaFileWrapper::attachmentFileNames() prevents us from using that function outside of a context where the delta file wrapper is not associated with the currently opened project, blocking push and sync of non-opened cloud projects.

test/test_digitizinglogger.cpp
nirvn: I copy / pasted a test file of yours, hence why you were there :)

nirvn:
@suricactus , thanks for the review. 

> We no longer pass the singleton project instance in the DeltaFileWrapper constructor (good move), but the projectId (ok). When I was writing this code back then I always assumed that in QGIS 4 having more than 1 project instances will no longer be a blasphemy. But probably having 200 project instances in the memory is not smart either for all the projects that a user has access to.

Yeah so on that front a/ QGIS still has some parts that's tied to QgsProject::instance(), so we can't pop the bottle of champagne yet, but more fundamentally we would not want to have separate project instances for cloud project objects and their delta file wrappers. Beyond the memory cost, it would just not work well (i.e., when you open a cloud project, do you hot swap the project instance for the main instance displayed on the map canvas where features are edited and signals emitted? nope)

> Your approach is to pass the QgsProject instance in few special occasions - the addCreate/Delete/Patch, apply* and addAttachment methods. I find this a bit confusing. Can't we determine what is the project instance directly from the projectId? This mixture between when to pass the project or the projectId makes me feel this can be done in another way.

I'm confident enough this is a good step in the right direction (other than the apply*() functions, but those will be dealt with in due time separately). For the add{Create,Delete,Patch,Attachments} functions, it's a natural fit as we already pass on all of the other context required to form the delta (i.e. localLayerId, sourceLayerId, localPkAttrName, sourcePkAttrName, and oldFeature).

src/qml/LocatorSettings.qml
nirvn: Correct.

src/core/qfieldcloud/qfieldcloudprojectsmodel.cpp
nirvn: We can make things nicer on the QML side of things by getting rid of the hyper-reliance on the QFieldCloudProjectsModel class (e.g. in the QFC popup, we should use the cloud project pointer directly), which will make things nicer.

As for the null check here (and in other places), IMHO it's the right approach. It'll feel fluid as we gradually reduce our usage of the QFC projects model class.

src/core/qfieldcloud/qfieldcloudproject.cpp
nirvn: It's something that's done throughout the code. For me instinctively, I see isEmpty() as _false/negative_.

Essentially here we say:

if ( we_have_a_good_id ) {
  do something;
} else {
  abort;
}

I prefer dealing with double negatives as a positive and show it first :wink:

src/core/qfieldcloud/deltafilewrapper.cpp
nirvn: @suricactus , it doesn't quite do the same, it stops relying on a QgsProject instance (i.e. addAttachments()) as we now trust that if a file was in a delta to begin with, it was part of the project :)

It merges the old and new checksums to insure that extra files added when merging two deltas are combined. I'll add a note.

src/core/qfieldcloud/qfieldcloudproject.cpp
suricactus: ok, I see your logic. I thought it just slipped when you were refactoring. You are the boss in this repo and you know best what works for you:)

Just to let you know that I instinctively tend to reorder the double negation, so I guess some of my contributions might do not follow your logical path. Feel free to highlight them in my future PRs :) .

Just for the Python repos (QFS/QFC) makes reading python much easier.

src/core/qfieldcloud/qfieldcloudprojectsmodel.cpp
suricactus: Thanks for the explanation. Good with it.

src/core/qfieldcloud/deltafilewrapper.cpp
suricactus: I see, the potential mismatch between the current QgsProject instance and cloud project gives a clear answer. As for the checking in another place, IMO the more we check, the better.

suricactus (approved):
@nirvn thanks for the detailed answers. Good to go from my side, I did not test it myself, but I don't have more questions regarding the source code anymore.

nirvn:
@suricactus , :pray: thanks for the review!

src/core/qfieldcloud/qfieldcloudproject.cpp
nirvn: @suricactus , roger that.

PR #6782, opened 2025-11-01
When making bookmark addition/update/deletion, immediately store the changes to the device

nirvn:
The QgsBookmarkManager class only writes its content to disk when the QgsApplication is exiting. For Android, this can be an issue as quite a few users never actually  quit QField but rather suspend QField, which results in the OS eventually killing it to regain memory. A killed QField does _not_ go through its exit code.

To avoid users losing their bookmarks, the solution is to immediately store the content of the bookmark manager after an addition, update, or deletion operation.

### DIFF
--- a/src/core/bookmarkmodel.cpp
+++ b/src/core/bookmarkmodel.cpp
@@ -204,7 +204,6 @@ void BookmarkModel::store()
     QFile f( filePath );
     if ( !f.open( QFile::WriteOnly | QIODevice::Truncate ) )
     {
-      f.close();
       return;
     }
 


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 1f7eb60a0bf106e6d3b4b89f7c4c94ee06687379: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6782-ec0b4367dfc65a3993e1c7815fd7c8f9e9c90df2-android-x86.apk)
</details>

src/core/bookmarkmodel.cpp
```diff
@@ -189,3 +195,25 @@ QgsCoordinateReferenceSystem BookmarkModel::getBookmarkCrs( const QString &id )
...
+
+void BookmarkModel::store()
+{
+  const QString filePath = QStringLiteral( "%1/bookmarks.xml" ).arg( QgsApplication::qgisSettingsDirPath() );
+  if ( !filePath.isEmpty() )
+  {
+    QFile f( filePath );
+    if ( !f.open( QFile::WriteOnly | QIODevice::Truncate ) )
+    {
+      f.close();
```
MidnightCoder-m: If the file isn't open, then `close()` would be unnecessary, right?

src/qml/BookmarkProperties.qml
```diff
@@ -28,7 +29,15 @@ QfPopup {
     colorContainer.value = bookmarkGroup;
   }
 
-  function saveBookmark() {
+  onAboutToHide: {
+    if (!bookmarkDeleted) {
+      bookmarkModel.store();
```
MidnightCoder-m, line 34: Question: Can we move `store()` to `updateBookmarkDetails()` and avoid it here?
Then we can remove its `Q_INVOKABLE` too!

src/qml/BookmarkProperties.qml
```diff
@@ -70,7 +79,7 @@ QfPopup {
         text: ''
 
         onTextChanged: {
-          saveBookmark();
+          updateBookmark();
         }
```
MidnightCoder-m, line 83: Here, `updateBookmark()` fires on every keystroke, so two alternatives:
1. Using `onTextEdited`
2. Using `onFocusChanged`  

To be honest, I’m not entirely sure which one would work best for the Android release.

src/core/bookmarkmodel.cpp
nirvn: I'm mimicking the QgsBookmarkManager::store() function here, ultimately I'd like for that one to be exposed upstream so we can use that instead. 

But it the meantime, we can indeed tweak a bit :)

src/qml/BookmarkProperties.qml
nirvn: I'd rather not here. In case we're dealing with 1000s of bookmark accumulated over time, I think there will be a cost that my start to translate into noticeable milliseconds.

src/qml/BookmarkProperties.qml
nirvn: Updating in-memory is not costly here, I think it's reasonable to keep it simple :)

src/core/bookmarkmodel.cpp
nirvn: @mohsenD98 , done.

PR #6796, opened 2025-11-10
Add drag-to-resize behavior for feature form indicator

mohsenD98:
### 🚀 Description

Adds interactive drag-to-resize behavior to the feature form indicator so users can drag the handle to resize or expand the feature form. This improves discoverability and provides immediate visual feedback while adjusting the form size.

### ✨ Key Changes
* Adds a small visual handle to the status indicator to indicate draggability

### 🎥 Demo

https://github.com/user-attachments/assets/fa347c18-16d7-46cd-88ec-f4c76055b214


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 52d5ff0ca34ff2e5dc2b2ae7eef7036ad4ed8191: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6796-f0bd71fb43884864ab1b1d679817f6a64103e35a-android-x86.apk)
</details>

nirvn:
@mohsenD98 , this will require some more work to be ready.

First, using mouseX and mouseY is not possible in this case as they are relative the MouseArea, which means that every time you move the window, you end up with a new relative mouseX / mouseY, which created really bad UX feedback (the page would jump up and down somewhat arbitrarily).

A better way here is to use the draghandler {} and use the centroid.scenePosition value, which gives us an absolute value that will be reliable as we move the feature form around.

The comment regarding the grab indicator moving to the left of the feature form when the feature form is on the right edge of a landscape window orientation also meant to suggest we should drag from that grab indicator. Merely moving it visually makes it even _more_ confusing as the actual drag remains on the header itself. This needs addressing. We could decide to make the drag indicator stay at the top all the time, and while a bit weird it'd still be better than nothing at all.

As discussed, we also need this in the feature addition item, so that'll need to be added too.

Lastly, the top margin handling on Android was not working properly (bad alignment and all). One trick when you develop an item that is impacted by these margins is to manually edit the value on linux to give you the ability to check without having to try an APK for every single change. You do it on these lines (https://github.com/opengisch/QField/blob/master/src/qml/qgismobileapp.qml#L56-L59). You can put ridiculous values like 100, that'll help you test your changes.

MidnightCoder-m:
@nirvn, Yes, thanks! I actually learned today how much more stable `DragHandler` is, so I really appreciate the tip. Also, following our chat, I’ve moved the indicator (the drag handle) to the top.

src/qml/OverlayFeatureFormDrawer.qml
```diff
@@ -3,16 +3,21 @@ import QtQuick.Controls
 import QtQuick.Controls.Material
 import QtQuick.Window
 import org.qfield
+import Theme
 
 /**
  * \ingroup qml
  */
-Drawer {
+Pane {
```
nirvn: Was the change from drawer to pane necessary here?

nirvn (commented):
We're getting closer. Tried locally and I can see two issues:
- the actual bar that indicates the ability to drag isn't draggable, that's a problem here as people with stylus (or on desktop) will want to click on that bar and then nothing happens
- while you're doing a better job at handling the top margin, we should have the feature be full screen, i.e. we shouldn't see the map below. That was the case before, so you'll just have to revert to previous handling there

src/qml/OverlayFeatureFormDrawer.qml
mohsenD98: @nirvn, Because the FeatureListForm layer is implemented using a **Pane**, and we already custom-implemented the slide-up/slide-down behavior, I didn’t want us to maintain a separate algorithm just for a **Drawer**. future changes become much easier and everything stays consistent in terms of format. We might even be able to introduce a dedicated `Qf----` Component for this in the future.

MidnightCoder-m:
> We're getting closer. Tried locally and I can see two issues:
> 
> * the actual bar that indicates the ability to drag isn't draggable, that's a problem here as people with stylus (or on desktop) will want to click on that bar and then nothing happens
> * while you're doing a better job at handling the top margin, we should have the feature be full screen, i.e. we shouldn't see the map below. That was the case before, so you'll just have to revert to previous handling there


both addressed. Thank you.

src/qml/OverlayFeatureFormDrawer.qml
MidnightCoder-m: Reverted to the drawer.

nirvn (approved):

nirvn:
@mohsenD98 , OK, I think this is ready to go. I've updated the control images to make the tests pass again.

I feel like we might need a bit more spacing around that drag indicator, but for now I'd like to see it live on unstable to get a good sense of it.

MidnightCoder-m:
> @mohsenD98 , OK, I think this is ready to go. I've updated the control images to make the tests pass again.
> 
> I feel like we might need a bit more spacing around that drag indicator, but for now I'd like to see it live on unstable to get a good sense of it.

Agree, Thanks.

PR #6817, opened 2025-11-19
Allow cloud project QR codes to embed username and url properties

nirvn:
This PR adds the ability to embed a QFC endpoint URL and username pair of properties into the url used to generate cloud project QR codes.

For example, this URL would look for a cloud project matching project_id on a QFieldCloud instance running on https://custom.qfield.cloud/ and enforce my_user username:

image

_qfield://cloud?project=project_id&url=custom.qfield.cloud&username=my_user_

QField will conduct a small set of checks to insure that the right endpoint URL and (if present) the right user is enforced. If a need for re-connection occurs, users will be shown the login panel with prefilled properties.

@spwoodcock , as discussed last(?) month.

### DIFF
--- a/src/core/qfieldcloud/qfieldcloudprojectsmodel.cpp
+++ b/src/core/qfieldcloud/qfieldcloudprojectsmodel.cpp
@@ -78,15 +78,7 @@ void QFieldCloudProjectsModel::setCloudConnection( QFieldCloudConnection *cloudC
     connect( mCloudConnection, &QFieldCloudConnection::urlChanged, this, &QFieldCloudProjectsModel::urlChanged );
 
     mUsername = mCloudConnection->username();
-
-    if ( !mProjects.isEmpty() )
-    {
-      beginResetModel();
-      qDeleteAll( mProjects );
-      mProjects.clear();
-      endResetModel();
-    }
-    loadProjects();
+    resetProjects();
   }
 
   emit cloudConnectionChanged();
@@ -366,18 +358,15 @@ void QFieldCloudProjectsModel::connectionStatusChanged()
 void QFieldCloudProjectsModel::usernameChanged()
 {
   mUsername = mCloudConnection->username();
-
-  if ( !mProjects.isEmpty() )
-  {
-    beginResetModel();
-    qDeleteAll( mProjects );
-    mProjects.clear();
-    endResetModel();
-  }
-  loadProjects();
+  resetProjects();
 }
 
 void QFieldCloudProjectsModel::urlChanged()
+{
+  resetProjects();
+}
+
+void QFieldCloudProjectsModel::resetProjects()
 {
   if ( !mProjects.isEmpty() )
   {
@@ -386,6 +375,8 @@ void QFieldCloudProjectsModel::urlChanged()
     mProjects.clear();
     endResetModel();
   }
+
+  // Load locally stored projects
   loadProjects();
 }
 
--- a/src/core/qfieldcloud/qfieldcloudprojectsmodel.h
+++ b/src/core/qfieldcloud/qfieldcloudprojectsmodel.h
@@ -236,6 +236,7 @@ class QFieldCloudProjectsModel : public QAbstractListModel
 
     void loadProjects( const QJsonArray &remoteProjects = QJsonArray(), bool skipLocalProjects = false );
     void insertProjects( const QList<QFieldCloudProject *> &projects );
+    void resetProjects();
 
     inline QString layerFileName( const QgsMapLayer *layer ) const;
 


qfield-fairy:
🎉 Ta-daaa, freshly created APKs are available for 8930f4a638554011c011e98eefcfaa509179c021: [**arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-arm64.apk)
<details>

- [arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-arm-neon.apk)
- [x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-x64.apk)
- [x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-x86.apk)
- [**all access arm64-android**](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-arm64.apk)
- [all access arm-neon-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-arm-neon.apk)
- [all access x64-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-x64.apk)
- [all access x86-android](https://qfieldapks.sos-ch-dk-2.exoscale-cdn.com/ci-builds/qfield_all_access-dev-6817-5128ea41b98c91cbb4b9a96016bf95ffdca155e0-android-x86.apk)
</details>

spwoodcock:
This is great! Thanks @nirvn 😁

With this, from an external app we could:
- Create QFC users in advance via API.
- Generate a QRCode to open QField with the project and user preconfigured.
- The user will need to enter the password, so we need some way to communicate it to them.

I'm assuming that `password` wasn't used as a URL param due to it being pretty bad security practice.

This PR definitely goes a good way to making a more intuitive flow linking to QField from an external app! Really appreciated 🙌

The remaining thing for us to consider is the best approach to achieve a 'longinless' type experience from a user perspective. Essentially clicking 'Map In QField' in an external app, and having it open to start mapping immediately. We discussed an approach of QRCode generation possibly calling the API to create a new project-scoped user, and other ideas. I'll test out this APK though & go from there!

spwoodcock:
I was thinking, to avoid the issues with loginless workflows, do you think the following could work:
- Create a service account `public-contributor` for a project, scoped with permissions only for a single project.
- Modify this PR to allow for the `&password=xxx` URL param.
- From an external app, display a QRCode for the `public-contributor` user, for a specific project, with user/password included.

This way, we can have both:
1. Logged in users that want OSM attribution for their edits (once we work on the OSM OAuth integration).
2. Unauthenticated users that can still contribute, for projects where we configure it.

Can you foresee any issues with this approach?

>[!NOTE]
> Apologies I'm sounding like a broken record here about the loginless approach 😅
>
> The key reason is because some of our users don't even have an email address, let alone a QFC login!
>
> But given a good satellite image, a 2.5D map pitch, users are definitely capable of identifying geometries in their neighbourhood and adding local knowledge to the map.

nirvn:
>> I'm assuming that password wasn't used as a URL param due to it being pretty bad security practice.

Yeah, discussing this with the team while creating this PR, we didn't want to open the door for users to allow for plain text password within QR codes. That'd be a security nightmare for a manager to prevent their users to do that _in an environment where this would not be acceptable_. 

That being said, the behavior when scanning a QR code that embeds an endpoint URL and username is as follow:
- If QField detects that either the endpoint URL or username does not match the currently connected user, we automatically bring the user to the login page, with the endpoint and username _already pre-filled_.
- The only thing left to do is to enter the password (here, in your case, you could actually just show the password in plain text _below_ the QR code on your web front-end accompanied with a simple "type the following password: '123'" instruction to it)
- When the password is entered and users click on log in, they will automatically be brought to the cloud project specified in the QR code and be present a big download button.

This is probably as frictionless as we can possibly make this throuhg a QR code without an environment that requires a password.

This is hopefully enough of an improvement for now as we are discussing further to get to a wikipedia-like loginless mode for QFC. 

One possibility we were thinking of was to possibly allow for passwordless usernames, which would then make it quite easy to make the QR code auto-login with an empty password string.

spwoodcock:
Thanks for the context! Agree with all the rationale. Presenting a password for the user to enter isn't a huge problem, so I'm pretty happy with this flow.

Its different to what mappers might expect, coming from an ODK ecosystem, but its definitely only a small thing to get used to 😃

Thanks so much for progressing this!

src/core/qfieldcloud/qfieldcloudprojectsmodel.cpp
```diff
@@ -356,6 +366,27 @@ void QFieldCloudProjectsModel::connectionStatusChanged()
...
+{
+  if ( !mProjects.isEmpty() )
+  {
+    beginResetModel();
+    qDeleteAll( mProjects );
+    mProjects.clear();
+    endResetModel();
+  }
+  loadProjects();
 }
```
MidnightCoder-m, line 381: Let's not repeat, using a helper function like `resetProjectsList()`

MidnightCoder-m: also useable in line 81-88.

nirvn: Good idea.

PR #6825, opened 2025-11-24
Fix: trackpad pan/zoom interaction on MapCanvas

kaustuvpokharel:
**On laptop with a touchpad, pan and feature clicks could stop working**
after performing a two-finger scroll or pinch gesture. Zoom continued to work, but the map became effectively non-interactive for panning and tapping. (Note that: the button zoom was not effecting this, only the track-pad zoom gestures).
like in the screen records, as how panning was working for everything, before I used 2 finger scroll (track-pad) to zoom, and then the map becomes un-interactive for panning and tapping (zoom still working though).

https://github.com/user-attachments/assets/5a5df167-6ead-42d9-aee8-a9de0fdfbfca

Root cause found:
- Qt reports laptop touchpad events as PointerDevice.TouchPad.
- MapCanvas TapHandler/DragHandler instances only accepted Stylus/Mouse or TouchScreen devices. After a touchpad gesture, subsequent events were coming from a TouchPad device and were ignored by all pan/click handlers.

Fix used:
- Treat touchpad like mouse on desktop for the "stylus" handlers, and keep the "main" handlers dedicated to touch (or mouse-as-touchscreen) only.
- Update acceptedDevices as follows:
  * stylusTapHandler / stylusDragHandler:
    - When qfieldSettings.mouseAsTouchScreen == false:
      accept Stylus | Mouse | TouchPad.
    - When qfieldSettings.mouseAsTouchScreen == true:
      accept Stylus only (mouse/touchpad handled by main* handlers).
  * mainTapHandler / mainDragHandler:
    - When qfieldSettings.mouseAsTouchScreen == false:
      accept TouchScreen only (original behavior).
    - When qfieldSettings.mouseAsTouchScreen == true:
      accept TouchScreen | Mouse | TouchPad.
- PinchHandler is restricted back to PointerDevice.TouchScreen to avoid
  accidentally grabbing platform-specific touchpad gestures on desktop.

Behavior after the fix:
- On desktop/laptop with a touchpad:
  * Left-click + drag pans immediately after opening.
  * Feature tap/click works before and after any two-finger scroll zoom.
  * Two-finger scroll (wheel) still zooms via WheelHandler.

This keeps the original MapCanvas interaction model intact, but extends it to correctly handle PointerDevice.TouchPad without breaking existing touch and mouse behavior.

### DIFF
--- a/src/qml/MapCanvas.qml
+++ b/src/qml/MapCanvas.qml
@@ -256,9 +256,7 @@ Item {
     id: stylusTapHandler
     enabled: interactive
     grabPermissions: PointerHandler.CanTakeOverFromHandlersOfDifferentType | PointerHandler.ApprovesTakeOverByAnything | PointerHandler.ApprovesCancellation
-    acceptedDevices: !qfieldSettings.mouseAsTouchScreen
-                        ? PointerDevice.Stylus | PointerDevice.Mouse | PointerDevice.TouchPad
-                        : PointerDevice.Stylus
+    acceptedDevices: !qfieldSettings.mouseAsTouchScreen ? PointerDevice.Stylus | PointerDevice.Mouse | PointerDevice.TouchPad : PointerDevice.Stylus
     acceptedButtons: Qt.LeftButton | Qt.RightButton
 
     property bool longPressActive: false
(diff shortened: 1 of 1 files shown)


nirvn:
@kaustuvpokharel , interesting, linux (and/or wayland) registers touchpad as a generic pointer device, and doesn't classify it as PointerDevice.TouchPad. 

Did you test this change on an Android phone to make sure we're not regressing there already?

nirvn:
(The APKs to test the changes are served as artifacts to the Android CI here: https://github.com/opengisch/QField/actions/runs/19624335931?pr=6825)

kaustuvpokharel:
@nirvn 
It doesn't affect the existing controls, I do not have android device, so I tested in IOS, and seems to work. Attached video with:

https://github.com/user-attachments/assets/d437e448-5498-4b00-83d9-565ab939877f
 
 It was a bit challenging to build ios for QFIELD, and found some changes needed for the project, will open a pr and issue for it next.

PR #6833, opened 2025-11-26
Fix iOS build: patched sentry-cocoa outdated code and disabled SerialPort for IOS build

kaustuvpokharel:
### CASE

This PR fixes the iOS build of QField when using vcpkg, by:

* Patching `sentry-cocoa` to avoid a crashy `std::vector<const T>` usage.
* Documenting / disabling `WITH_SERIALPORT` on iOS so CMake can actually configure.

### 1. sentry-cocoa build failure (vcpkg, iOS)

When building the iOS target with vcpkg’s `sentry-cocoa` **7.23.0** in xcode16.3+, the build fails because Old sentry-cocoa version using std::allocator<const T> inside profiling code (ThreadMetadataCache), the error looks like this in the log:
Screenshot 2025-11-24 at 10 25 37 PM

xcode 16.3 changed the C++ toolchain, as mentioned explicitly here;
https://developer.apple.com/documentation/xcode-release-notes/xcode-16_3-release-notes#:~:text=libc%2B%2B%20no%20longer,allocator%2Dconst%20enabled.

Upstream has already fixed this in **8.46.0+** as the issue link has the same issue;
https://github.com/getsentry/sentry-cocoa/issues/5172
and comment has the ticket confirming the fix.

I applied a targeted patch (`const-vector.patch`) to remove the `const` qualifier from vector element types in `SentryThreadMetadataCache.hpp` for now, beacause rather than upgrading to a newer sentry-cocoa version (which could introduce breaking changes, new dependencies, or behavioral differences, which I tried and encountered) this tiny patch is best alternative to keep it running for now.

So we effectively get the upstream fix while still using the vcpkg-provided version **<7.23.0>**.
(Alternative for the future: upgrading vcpkg’s `sentry-cocoa` port to **8.46.0+** and drop this patch.) Which I tried but was quiet daunting to make it work with vcpkg port change, noting my little knowledge within vcpkg.

---

### 2. Qt SerialPort and iOS build

On iOS, vcpkg marks `qtserialport` as unsupported, so CMake fails with:
Screenshot 2025-11-25 at 7 28 17 PM

QField’s CMake currently does:

```cmake
if(WITH_SERIALPORT)
  list(APPEND QFIELD_CORE_SRCS positioning/serialportmodel.cpp
       positioning/serialportreceiver.cpp)
  list(APPEND QFIELD_CORE_HDRS positioning/serialportmodel.h
       positioning/serialportreceiver.h)
endif()
```
And is on global boundary, so without the flagging it off the ios will run into respected error.
##Also considering SERIAL PORT not supported in ios
https://forum.qt.io/topic/121465/serialport-for-qt-ios

For iOS we therefore disable serialport support and document that the iOS configure line must include:

```bash
-DWITH_SERIALPORT=OFF
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,9 +13,6 @@ set(APK_VERSION_CODE "1" CACHE STRING "Apk Version Code (Example: 1)")
 set(APP_VERSION "0" CACHE STRING "Application Version (Example: v1.0.0)")
 set(APP_VERSION_STR "local - dev" CACHE STRING "Application Version Name (Example: 1.0.0 - Homerun)")
 set(APP_PACKAGE_NAME "qfield" CACHE STRING "Package name suffix. E.g. qfield --> ch.opengis.qfield")
-if(IOS)
-    set(WITH_SERIALPORT OFF CACHE BOOL "Disable serial port support on iOS" FORCE)
-endif()
 
 string(REGEX REPLACE "v" "" CLEAN_APP_VERSION "${APP_VERSION}")
 
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -226,26 +226,47 @@ export PATH="$(brew --prefix flex)/bin:$(brew --prefix bison)/bin:$PATH"
 #
 
 cmake -S . -B build-arm64-ios \
-	-DVCPKG_HOST_TRIPLET=arm64-osx \
-	-DVCPKG_TARGET_TRIPLET=arm64-ios \
-	-DWITH_VCPKG=ON \
-	-DVCPKG_BUILD_TYPE=release \
-	-DCMAKE_SYSTEM_NAME=iOS \
-	-DCMAKE_OSX_SYSROOT=iphoneos \
-	-DCMAKE_OSX_ARCHITECTURES=arm64 \
-	-DCMAKE_SYSTEM_PROCESSOR=aarch64 \
-	-GXcode
+  -DVCPKG_HOST_TRIPLET=arm64-osx \
+  -DVCPKG_TARGET_TRIPLET=arm64-ios \
+  -DWITH_VCPKG=ON \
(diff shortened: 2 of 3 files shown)


cmake/AddQtIosApp.cmake
```diff
@@ -4,8 +4,6 @@
 # add_qt_ios_app help you deploy iOs application with Qt.
 #
 
-cmake_minimum_required(VERSION 3.0.0 FATAL_ERROR)
```
nirvn, line 7: Is this change needed?

kaustuvpokharel: Yes, I believe it should be removed as it is older and doesnt support current changes, if this not absolutely necessary,
The cmake docs:
https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#:~:text=Changed%20in%20version%203.19,CMake%203.19%20and%20above.

The error explicitly said this too:
`cmake/AddQtIosApp.cmake:7 (cmake_minimum_required): Compatibility with CMake < 3.5 has been removed from CMake. Update the VERSION argument <min> value. Or, use the <min>...<max> syntax to tell CMake that the project requires at least <min> but has been updated to work with policies introduced by <max> or earlier. Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway. Call Stack (most recent call first): src/app/CMakeLists.txt:165 (include) -- Configuring incomplete, errors occurred! kaustuvpokharel@Kaustuvs-MacBook-Pro-10 QFIELD_RUN %`

Since the project already sets a higher minimum in the toplevel CMakeListstxt, the correct fix is to remove this redundant call from the helper module which is incompatible, but if it is absolutely needed using version flag will solve it but I think it is best to not have it. Let me know what you think @nirvn 😊

nirvn:
@kaustuvpokharel , thanks for this!

PR #6871, opened 2025-12-08
Remove unread message indicator from map canvas + add notification badge to “Message Log” menu entry

kaustuvpokharel:
This PR removes the message log icon from the map canvas and replaces it with a badge on the “Message Log” menu entry. Unread messages are now indicated centrally in the main menu instead of on the canvas.

**Details**
* `MessageLog` page:
  - Added `unreadCount` property to track how many messages are added while the page is not visible.
  - Kept the existing `unreadMessages` boolean for backwards compatibility
  - Reset both `unreadMessages` and `unreadCount` when the message log becomes visible.

* Menu:
  - Added a `QfBadge` to the “Message Log” `MenuItem`.
  - The badge is shown only when `messageLog.unreadMessages` is `true`, like an active unread notification.
  - The badge text displays `messageLog.unreadCount`.

**Behaviour**
* When new messages are appended while the Message Log page is closed, the “Message Log” menu entry shows a badge with the unread count.
* Opening the Message Log clears the unread state and hides the badge.
* Other parts of the code relying on `messageLog.unreadMessages` continue to behave as before; the new `unreadCount` is an additive, non-breaking change.

**Improvement Feed and questions**
-> badge is quite small on the screen with the `Theme.mainColor` background with the `Theme.lightColor` text of the count of unread message, which is not properly visible when digits goes to 2 or 3 number i.e 10+ or 99+, 
Potential solution could be instead of circle, we change the width+ for a pill shape badge when the text.length > 2 or 3 to make it more clear. Plus, font could use bold or higher weight, this shouldn't break what already exists. @nirvn if you could suggest potential improvement, I would want to set it 

 image

### DIFF
--- a/src/qml/qgismobileapp.qml
+++ b/src/qml/qgismobileapp.qml
@@ -3187,7 +3187,7 @@ ApplicationWindow {
         alignment: QfBadge.Alignment.TopRight
         visible: messageLog.unreadMessages
         color: Theme.mainColor
-        badgeText.text: messageLog.unreadMessagesCount
+        badgeText.text: messageLog.unreadMessagesCount >= 10 ? "+" : messageLog.unreadMessagesCount
         badgeText.color: Theme.light
         border.color: "transparent"
       }


src/qml/qgismobileapp.qml
```diff
@@ -3191,6 +3178,19 @@ ApplicationWindow {
...
+
+      QfBadge {
+        width: 16
+        height: width
+        topMargin: 5
+        rightMargin: 5
+        alignment: QfBadge.Alignment.TopRight
+        visible: messageLog.unreadMessages
+        color: Theme.mainColor
+        badgeText.text: messageLog.unreadMessagesCount
```
nirvn: Let's show a '+' sign if the unread count is greater than 10. That nb could grow quite a lot over a period of time.

Using the '+' sign for count >= 10 also matches the behavior we've established with the QFieldCloud pending changes badge in our dashboard (https://github.com/opengisch/QField/blob/master/src/qml/DashBoard.qml#L231).

kaustuvpokharel: Noted, thanks for pointing me to the file!

PR #6874, opened 2025-12-10
Added unit tests for IdentifyTool and improved Catch2 test configuration on macOS

kaustuvpokharel:
This PR adds a new Catch2 test suite for `IdentifyTool` and adjusts the CMake test configuration to make Catch2 tests more reliable on macOS.

**Changes**

 Added test_identifytool.cpp with tests covering:
  - Calling identify() with only mapSettings or only model set (no crashes, safe early return).
  - Deactivated state: when deactivated is true, identify() is a no-op and does not emit identifyFinished
  - Property signals: mapSettingsChanged, modelChanged, and searchRadiusMmChanged are emitted when their properties change
  - identifyFinished is emitted once when identify() is called with valid mapSettings and model
  - searchRadiusMm behaviour, including that the change signal is only emitted when the value actually changes

 Vector layer identification:
 
  - Features inside the search radius are returned for an in-memory point layer.
  - nullptr and non-spatial layers return no results.
  - A layer with a different CRS than the map can be identified without errors.
  - All tests pass locally.

**CMake / Catch2 test configuration**

- Use DISCOVERY_MODE PRE_TEST for all Catch2 tests to avoid fragile post-build test discovery.
- On macOS with the Xcode generator, ad-hoc sign Catch2 test executables after build so they can be executed correctly by CTest/CMake on recent macOS plus SDKs
- Fixes the `"Result: Subprocess killed"` issue coming from CatchAddTests.cmake when running tests on macOS with the Xcode generator

All tests pass locally

### DIFF
--- a/src/core/identifytool.cpp
+++ b/src/core/identifytool.cpp
@@ -395,6 +395,9 @@ void IdentifyTool::setModel( MultiFeatureListModel *model )
 
 void IdentifyTool::setDeactivated( bool deactivated )
 {
+  if ( mDeactivated == deactivated )
+    return;
+
   if ( deactivated && mModel )
     mModel->clear();
 


src/core/identifytool.cpp
```diff
@@ -395,8 +395,9 @@ void IdentifyTool::setModel( MultiFeatureListModel *model )
 
 void IdentifyTool::setDeactivated( bool deactivated )
 {
-  if ( deactivated )
+  if ( deactivated && mModel )
     mModel->clear();
+
   mDeactivated = deactivated;
 }
```
nirvn, line 405: Let's check whether we're actually changing the mDeactivated value first. So something like

if ( mDeactivated == deactivated )
{
  return;
}

if ( deactivated && mModel )
{
  mModel->clear();
}

...
}

nirvn (commented):
Minor comment, nice test addition, thanks.

PR #7048, opened 2026-02-11
[fix] 최신 원본을 내프로젝트로 가져오기

mapinusRND:


nirvn:
@mapinusRND , I'm not quite sure what you are trying to do here but if you are looking into translating QField into Korean, you can do so via https://explore.transifex.com/opengisch/qfield-for-qgis/ , your translation will be bundled with QField on app/play store.

PR #7364, opened 2026-04-23
Reset video card to thumbnail state after playback ends

kaustuvpokharel:
Improvement, when a gallery video finished playing, the card remained on the last frame (or went blank on some backends), requiring a click to return to a playable state
Detect natural end-of-playback via playbackState transitioning to StoppedState, then reload the Video element to restore the initial thumbnail + play-icon state. Media focus is released so the next playback starts cleanly

## Follow up problem :

Thanks to @nirvn for the idea of serialising video thumbnail captures via a timer, it turned out to be the right approach for the decoder contention problem

The problem we had was opening a gallery with many videos tried to activate every video element at once so the simultaneous activation exhausted decoders, thumbnails would fail to load, and on ios, mac this could come into a crash, especially during fast scrolling where GridView rapidly creates and destroys delegates

## The fix:
- videoQueue and videoQueueTimer at the root. Video loaders start with active: false and register themselves via queueVideoLoad()
- The timer activates one loader at a time with a 150ms gap between each. Decoders open sequentially instead of all at once
- Each loader carries a simple queued flag whhen GridView recycles or destroys a delegate, onSourceUrlChanged resets the flag, and the queued callback bails on its next turn, no dangling references to destroyed QML objects, which was the source of the fast-scroll crashes as well

https://github.com/user-attachments/assets/2de1af0f-7099-464e-8524-3b1fdb1bcd75

### DIFF
--- a/src/qml/editorwidgets/relationeditors/gallery_relation_editor.qml
+++ b/src/qml/editorwidgets/relationeditors/gallery_relation_editor.qml
@@ -871,6 +871,17 @@ RelationEditorBase {
         property url sourceUrl: attachmentIsVideo ? UrlUtils.fromString(attachmentFullPath) : ""
         property bool firstFrameDrawn: false
 
+        Timer {
+          id: reloadTimer
+          interval: 10
+          repeat: false
+          onTriggered: {
+            videoThumbLoader.firstFrameDrawn = false;
+            videoThumbLoader.active = false;
+            videoThumbLoader.active = true;
+          }
+        }
+
         sourceComponent: Component {
           Video {
             anchors.fill: parent
@@ -889,6 +900,15 @@ RelationEditorBase {
               if (!videoThumbLoader.firstFrameDrawn && playbackState === MediaPlayer.PlayingState) {
                 videoThumbLoader.firstFrameDrawn = true;
                 thumbnailPauseTimer.start();
+              }
+            }
+
+            onPlaybackStateChanged: {
+              // Detect natural end-of-playback
+              if (cardContainer.videoPlaying && playbackState === MediaPlayer.StoppedState && duration > 0 && position >= duration * 0.1) {
+                cardContainer.videoPlaying = false;
+                relationEditor.releaseMediaFocus(cardContainer);
+                reloadTimer.start();
               }
             }
 
--- a/src/qml/editorwidgets/relationeditors/gallery_relation_editor.qml
+++ b/src/qml/editorwidgets/relationeditors/gallery_relation_editor.qml
@@ -85,6 +85,31 @@ RelationEditorBase {
     externalStorage.fetch(next.url, next.authId);
   }
 
+  property var videoQueue: []
+  property bool videoQueueBusy: false
+
+  Timer {
+    id: videoQueueTimer
+    interval: 150
+    onTriggered: {
+      while (videoQueue.length > 0) {
+        if (videoQueue.shift()()) {
+          videoQueueTimer.start();
+          return;
+        }
+      }
+      videoQueueBusy = false;
+    }
+  }
+
+  function queueVideoLoad(fn) {
+    videoQueue.push(fn);
+    if (!videoQueueBusy) {
+      videoQueueBusy = true;
+      videoQueueTimer.start();
+    }
+  }
+
   QtObject {
     id: dummyTarget
   }
@@ -576,12 +601,29 @@ RelationEditorBase {
 
       Loader {
         id: listVideoThumbLoader
-        active: attachmentIsVideo
+        active: false
         parent: listVideoContainer
         anchors.fill: parent
 
         property url sourceUrl: attachmentIsVideo ? UrlUtils.fromString(attachmentFullPath) : ""
         property bool firstFrameDrawn: false
+        property bool queued: false
+
+        onSourceUrlChanged: {
+          active = false;
+          firstFrameDrawn = false;
+          queued = false;
+          if (attachmentIsVideo && sourceUrl.toString() !== "") {
+            queued = true;
+            relationEditor.queueVideoLoad(() => {
+              if (!listVideoThumbLoader || !listVideoThumbLoader.queued)
+                return false;
+              listVideoThumbLoader.queued = false;
+              listVideoThumbLoader.active = true;
+              return true;
+            });
+          }
+        }
 
         sourceComponent: Component {
           Video {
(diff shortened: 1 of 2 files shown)


qfield-fairy:
Download a [MacOS DMG universal build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/24915637736/qfield-dmg.zip).
*(Built from commit 278125abb988e6aeca4b0deb1332d3a256898ad4)*




### 📱 Android builds
Download an [Android arm64 build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-arm64.zip).
*(Built from commit 278125abb988e6aeca4b0deb1332d3a256898ad4)*
<details>

- [arm-neon](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-arm-neon.zip)
- [x64](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-x64.zip)
- [x86](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-x86.zip)
- [**all access arm64**](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield_all_access-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-arm64.zip)
- [all access arm-neon](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield_all_access-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-arm-neon.zip)
- [all access x64](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield_all_access-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-x64.zip)
- [all access x86](https://nightly.link/opengisch/QField/actions/runs/24915637764/qfield_all_access-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-android-x86.zip)
</details>




### 🪟 Windows builds
Download a [Windows build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/24915637700/QField-x64.zip) and for [arm64](https://nightly.link/opengisch/QField/actions/runs/24915637700/QField-arm64.zip).

*(Built from commit 278125abb988e6aeca4b0deb1332d3a256898ad4)*




### 🐧 Linux AppImage builds
Download a [Linux AppImage build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/24915637734/qfield-dev-7364-cc21ec885ff3e64c3e3980e9baa3cb558e83168a-linux-x64.AppImage.zip).
*(Built from commit 278125abb988e6aeca4b0deb1332d3a256898ad4)*

kaustuvpokharel:
@nirvn 
While working on this, It hit me with this idea that for every visible video card we  keep a live Video element or decoder alive just to display a paused first frame. For the external resource widget this is probably a non issue since only one attachment form is open at a time, single live decoder, no cost technically. For the gallery relation editor Im less sure how often surveyors end up with many video attachments per feature, but the concurrent-decoder limit exists across all platforms QField targets, just with different numbers and potential impact modes, on research:

on IOS: undocumented cap on concurrent AVPlayer render pipelines, typically reported as ~16 on modern devices ([Apple Developer Forums](https://developer.apple.com/forums/thread/67382)) and as low as 4 on older hardware ([Person article](https://www.beckyhansmeyer.com/2017/08/30/too-many-avplayers/)). Attempting more give AVFoundationErrorDomain Code=-11839 "Cannot Decode" error 

on android: MediaCodec has a per-codec concurrent-instances limit declared in each device's /etc/media_codecs.xml here [AOSP OEM req](https://source.android.com/docs/core/media/oem) show CTS explicitly checking this, with vendor configs commonly around 13-16 for H.264. 

on macos: shares VideoToolbox/AVFoundation with iOS which hardware decode sessions are limited and [only one process gets VideoToolbox hardware acceleration at a time](https://github.com/mpv-player/mpv/issues/6749). Software fallback works but seems to cost CPU

Windows: [DXVA](https://learn.microsoft.com/en-us/windows/win32/medfound/about-dxva-2-0) / Media Foundation don't publish a hard limit, but seems GPU-dependent as well, i am not sure but some drivers seems to become unstable beyond a handful of concurrent decode surfaces ([Mozilla 847267](https://bugzilla.mozilla.org/show_bug.cgi?id=847267))

on linux or ffmpeg software decode I see no API cap, scales with CPU and memory but the decoder thread should still cost noticeable CPU, nothing found on research but it should work similarly.

A followup solution I could think of is a helper FileUtils.extractVideoThumbnail() that extracts the first frame once and caches it as an image, so idle cards render a cached JPG via a plain Image element and a Video is only instantiated during actual playback. On IOS/Android that keeps us clear of the decoder caps; on desktop it removes the per-card CPU/memory overhead

Happy to know your take on this, if its worth pursuing, wanted to flag it here for your input on whether its a real concern for qfield typical gallery usage

nirvn:
@kaustuvpokharel , a VideoThumbnailGenerator element sounds good. Definitively a better method than doing the play then pause hack.

src/qml/editorwidgets/relationeditors/gallery_relation_editor.qml
```diff
@@ -892,6 +951,16 @@ RelationEditorBase {
...
 
+            onPlaybackStateChanged: {
+              // Detect natural end-of-playback
+              if (cardContainer.videoPlaying && playbackState === MediaPlayer.StoppedState && duration > 0 && position >= duration * 0.1) {
+                cardContainer.videoPlaying = false;
+                relationEditor.releaseMediaFocus(cardContainer);
+                seek(0);
+                pause();
+              }
+            }
```
nirvn, line 962: Do we need to add that for the list version?

kaustuvpokharel: No actually, the reason being that the listview never plays videos to completion, it captures the first frame and pauses immediately as a static thumbnail (see onPositionChanged -> listThumbnailPauseTimer.start())
End-of-playback handling only helps for the gridview because users can actually tap to play through the thumbnail

nirvn (approved):

nirvn:
@kaustuvpokharel , an enhancement means a new feature, not a fix. This PR does a great job at fixing a limitation :)

kaustuvpokharel:
> @kaustuvpokharel , an enhancement means a new feature, not a fix. This PR does a great job at fixing a limitation :)

Understood, thankyou @nirvn 😄

PR #7025, opened 2026-02-05
Add Ntrip client to qfield

natuition:
# Add NTRIP client integration for RTK corrections

Add complete NTRIP client implementation to provide real-time kinematic
corrections to external GNSS receivers via Bluetooth.

### Core Implementation
- Add **NtripSocketClient** for low-level TCP connection and NTRIP protocol
- Add **NtripClient** as high-level wrapper with byte counter tracking
- Integrate NTRIP client into **PositioningSource** with configuration properties

### Properties & State Management
- Add NTRIP properties: **host**, **port**, **mountpoint**, **username**, **password**
- Replace `ntripStatus` (QString) with **ntripState** (enum: Connected/Disconnected)
- Add **ntripLastError** property for error reporting (similar to deviceLastError)
- Add **streamConnected** and **streamDisconnected** signals
- Track **NTRIP bytes sent/received** for monitoring

### Bug Fixes
- Fix TCP header fragmentation by buffering data until headers complete

## Requirements

The NTRIP client is only enabled when an **external receiver** is active and
properly configured with all required connection parameters.

## Testing

**Tested on macOS** with **NavX GNSS receiver**.

### Screenshots

RTK positioning active with NTRIP corrections (using Centipede network) : 
Capture d’écran 2026-02-05 à 16 05 52

NTRIP client configuration panel :
Capture d’écran 2026-02-05 à 16 06 06

PR #6574, opened 2025-08-15
NTRIP client POC

edgecase14:
Blindly stuffs TCP packets received over Bluetooth.  RTCM or SPARTN.  TCP message fragments aren't reassembled.

Tested on Linux, with PointPerfect(thingstream) and SparkFun RTK Surveyor firmware on custom harware.  RTK Fixed is achieved.

See #5387.  Comments requested.  Android testing encouraged.

PR #7274, opened 2026-04-05
NTRIP squash review, part 2

nirvn:
Merely pushing to get a build against latest master.

PR #7435, opened 2026-05-15
QField NTRIP implementation

### LINKED ISSUE #5387: Enhancement: NTRIP client to connect Sparkfun RTK Surveyor
Usage is Sparkfun RTK Surveyor (F9P) via Bluetooth from cellphone (Android). Tested with SW Maps and this is working but would like to use QField. The direct Bluetooth connection is recommended by QField documentation but there is no method from QField to send RTK correction data back to receiver over Bluetooth or stream from a public sector data source UNAVCO with RTCM protocol.

Possible workarounds:

Fake GPS instance on Android controlled as listed in QField documentation. The documentation recommends instead a direct Bluetooth connection and avoid this method when possible. As listed the recommended software is not Open Source software and anyways this implements a specific feature of Android so it is not a robust method.

Configure Sparkfun RTK Surveyor to connect to WiFi hotspot and integrated RTCM / NTRIP client, then connecting from QField to the Sparkfun RTK Surveyor using Bluetooth as recommended. This works only when there is a full size AP with a good antenna close by the RTK Surveyor unit. This method is not reliable using cellphone built-in hotspot, the signal is weak to the device and besides this the cellular carrier assesses phone hotspot Internet sharing differently than data used from apps within the phone device so there are carrier-level complications.

Running own base station for corrections data and a serial data link over RF.  Extra cost, setup, and not convenient where there is a source of NTRIP / RTCM data valid for the area being surveyed.

I searched the discussions and forum finding some references to F9P hardware, and even the Sparkfun Facet, but nothing sure about how to use QField with a direct Bluetooth connection as recommended and also streaming correction data that is sent over that Bluetooth link back to the RTK Surveyor;  as SW Maps is doing successfully.

Suggestion is an enhancement:  Add an NTRIP client to QField and support bi-directional communication of this RTK correction data with the Bluetooth connected receiver.

nirvn:
This PR implements NTRIP corrections within QField itself. It is still a work in progress but there is enough in there to enjoy :)

It builds on top of the work done by the community, including:
- https://github.com/opengisch/QField/pull/7025
- https://github.com/opengisch/QField/pull/6574

How the UI looks within the settings's positioning panel:

ui

TO DO:
- [x] more UI polishing
- [x] documentation
- [x] testing

Fixes https://github.com/opengisch/QField/issues/5387


qfield-fairy:
### 📱 Android builds
Download an [Android arm64 build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-arm64.zip).
*(Built from commit 11d3deb348e12c9b404bf0fe3ea12e1efeeb5b4e)*
<details>

- [arm-neon](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-arm-neon.zip)
- [x64](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-x64.zip)
- [x86](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-x86.zip)
- [**all access arm64**](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield_all_access-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-arm64.zip)
- [all access arm-neon](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield_all_access-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-arm-neon.zip)
- [all access x64](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield_all_access-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-x64.zip)
- [all access x86](https://nightly.link/opengisch/QField/actions/runs/26137134413/qfield_all_access-dev-7435-dfc322794807f48064144226c5a6de14727ca294-android-x86.zip)
</details>




Download a [MacOS DMG universal build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/26137134412/qfield-dmg.zip).
*(Built from commit 11d3deb348e12c9b404bf0fe3ea12e1efeeb5b4e)*




### 🪟 Windows builds
Download a [Windows build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/26137134409/QField-x64.zip) and for [arm64](https://nightly.link/opengisch/QField/actions/runs/26137134409/QField-arm64.zip).

*(Built from commit 11d3deb348e12c9b404bf0fe3ea12e1efeeb5b4e)*




### 🐧 Linux AppImage builds
Download a [Linux AppImage build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/26137134420/qfield-dev-7435-dfc322794807f48064144226c5a6de14727ca294-linux-x64.AppImage.zip).
*(Built from commit 11d3deb348e12c9b404bf0fe3ea12e1efeeb5b4e)*

bettellam:
Tested this build on Android with:

**Receiver:** Unicore UM982 (dual UART Bluetooth HC-05)
**Connection:** Bluetooth SPP → UM982 BT_UART
**Network:** caster.centipede.fr or crtk.net
**Everything OK, no problems!**

If I switch to:
**Network:** [Rete GNSS Veneto](https://retegnssveneto.cisas.unipd.it/Web/index.php)(Italian NTRIP network - VRS)
where sending the GGA phrase is required, 

the connection becomes unstable:
- the connection to the caster doesn't always initiate (NTRIP client disconnected);
- the connection initiates with a significant delay (NTRIP client connected);
- the send GGA counter doesn't update;
- after a short connection period, it drops (NTRIP client disconnected).

**I think the problem is actually sending GGA!**

nirvn:
@bettellam , his this checkbox enabled?

image

And, do you see the bytes sent counter go up when connecting to the caster here?

image

I've just pushed a new commit that adds a bit of flexibility detecting GGA strings (https://github.com/opengisch/QField/pull/7435/commits/a80e737e21094990435a8a66662d3844157b423d). If you were not seeing the bytes sent counter go up, maybe this latest commit will work for you.

Big thanks for testing. Did you find the added UI easy enough to understand / navigate?

gdt:
Is this sending every GGA, or at some  lower rate?     For high-rate output (say 10 Hz), sending every position would be spamming the caster.   I would say that there's no good reason a caster needs a position more than once a minute -- since it's only about selecting "NEAR" or doing some kind of VRS -- but I suspect some insist on more.

Overall, I'd say limiting to 30s (as in, once sent, don't send another until 30s later) would be good, absent evidence that it's not.

nirvn:
@gdt , ATM it'll send no more than one location per second. I'm wondering if we should make the rate configurable as suggested in a different PR.

eshattow:
I was not able to find the NTRIP settings view on my own, as it is narrowly sandwiched between two things that toggle on/off NTRIP client.  I think if you want to keep this setting where it is (visually it seems okay to me just difficult for discovery) then make the action on the text of the option go to the NTRIP settings view and have the on/off toggle be the spatially narrow activation (or bury the on/off toggle into the "NTRIP settings" view). At that, can just have this be "NTRIP corrections client"  and not the usual "Enable NTRIP corrections" since we're overloading that context having a few different intentions in the same line item.

NTRIP port setting should default to 2101, technically it's 80 or 2101 in the standard. It seems to be most commonly 2101, and the fallback_port is defined as 0 for no fallback.

I'm not sure if it is a real bug or not, SW Maps allows to select NTRIP Version and when configured for V1 or V2 either is successful there, yet in QField the NTRIP V1 fails to connect and V2 is successful. I don't know what SW Maps is doing there, if it is failing over silently or if there is some actual connection issue with the QField NTRIP implementation.

Notably SW Maps allows to configure a list of NTRIP settings groups and select the group for use, which is useful and not implemented here. I like the clean layout in QField for this initial implementation however the utility of selecting a group of NTRIP client settings is a real non-theoretical need, as preparing configuration groups of casters or mountpoints is best done "in the office" before heading to the field. I do not do this kind of work professionally so in my situation when I'm going to help a friend with some RTK-like things and they're located in a region that benefits from a different caster and zone settings, I would like to set this up ahead of time with a few likely possible configurations and simply select my "home" NTRIP profile when I'm back on my own projects.

I would also like to be able to quickly toggle between Instrument Height Offset, I think as part of the Positioning device profile.  I have the antenna mounted on a 2.000m rover rod for recording accurate points, and a 0.3900m stakeout rod that is more manageable to quickly stakeout previously recorded points;  the same receiver is attached to one of these rods depending on what I'm doing for my project and though I know good practice would be to measure everything as part of a checklist and enter that number each use it would be more convenient for me to just select whichever positioning device profile I've already measured for;  it could be the same receiver again under a different name  e.g. "SparkFun Surveyor on 39cm Stakeout Rod" or "SparkFun Surveyor on 2m Rover Rod".

The text of checkmark box ui option layout "Forward NMEA sentences to NTR..." is all I see it doesn't wrap to the next line on a small display screen. SW Maps has this (I would suppose it is the same anyhow) as per-NTRIP group option "Send NMEA GGA to Base Station". Neither option makes any sense to me unless I'm already having some kind of prior knowledge about NMEA, which sentences are being sent? what is a "GGA"? At least "NMEA GGA" is web searchable, but still kind of meaningless, and the generic "NMEA sentences" is too broad of a search term so I think it is not clear anyways what is being accomplished by turning on and off this option. This is a "miss" for me for such a long description that I don't know what it does even then. I think make it shorter, more distinct for searching (even if that ends up being cryptic), and stick it in documentation so it will be discoverable.

I would like to be able to access the text of the Mount point for natural editing, for typing entry or long-press context copy/paste text. The pre-fill / autocomplete manipulation may be better served as a filter on the drop-down results instead of interfering with the cursor position. The quantity of mount points is enormous on the caster I am typically using and so scrolling through much of that list is not efficient. It would be more natural to type in the first letters of the mount point and then have that narrow the filter applied to the results in the drop-down. Currently I am able to long-press to paste additional text but there is some kind of manipulation of the cursor happening and it prevents the select-all copy sequence. The "autofill" behavior is not ideal because the visual information needed to know what might be valid to type next is in the drop-down selection view that is inherently not visible while typing in the input field form of that ui element. Good attempt, just not quite exactly what it needs to be and ends up blocking select-all copy (or select-all paste) sequences.

I would like to see just a little more detail about the NTRIP connection. The RX / TX counters are good, but I still would like more information to help me understand the difference between a connection is established and the state of useful data being parsed SW Maps has for example:
```
NTRIP Status
Connected
Base Antenna: TRM59800.99    SCIT
dZ: 85.690000 mm
Base Latitude: nn° nn' nn.nnn" NS
Base Longitude: nn° nn' nn.nnn" EW
Base elevation: nnnn.nnn ft
Antenna Offset Not Applied
```

or "Antenna Offset Applied" if the corresponding ui checkbox element is state-on in the NTRIP connections groups view.

Briefly when the connection is being initiated this is simplified as:
```
NTRIP Status
Connected
Base Antenna: ADVNULLANTENNA
dZ: 0.000000 mm
```

It is useful to know no meaningful data has been parsed in this way.

If the NTRIP client connection is active, and I change NTRIP client settings, when are those applied? does it spam new connection attempts everytime I change a settings field?  Does the new setting apply on dismissal of the NTRIP settings view? Does the new setting apply when off/on cycle the state ?

Possibly the on/off NTRIP client toggle ui element could be moved to top-right corner of "NTRIP Settings" and then disable the client whenever a setting is changed that cannot reasonably be applied to the active connection. The user would see that the client is disconnected in this way in the same view of the settings they are applying. It would not be surprising because "NTRIP client disconnected" or the RX / TX counters are visible from the parent settings in "Positioning" view.

On to the other user experience stuff I made notes is not all strictly in-scope:

I suggest:
- drop the "Add" button, make that "Clone" instead
- change "Internal device" to be editable / discoverable so the user can clone and edit with configurable name
- immutable device "No device" which when clone is effectively the same as a new non-cloned item is presently, is the only selection choice for which "Connect" "Disconnect" is input-disabled
- "Remove" action may be relocated to within the profile view and consider if necessary to have with a long-press to avoid accidental use.
- Each position device profile can retain own settings and overrides of parent (Positioning) settings where applicable
- Parent (Positioning) settings input-disables (yet keeps visible and reflecting the active state of) per-profile override settings; The capability to edit these general level settings is preserved i.e. for "in the office" preparations by doing so with the "No device" device selected, which by definition cannot have any overrides so nothing to input-disable in the parent (Positioning) settings.
- "Connect" / "Connecting" button should also overload to "Disconnect" action when appropriate, and as they are directly acting on the selected profile that button does not need to repeat the name. Admittedly this is a tricky balance between efficient use of small display and keeping the imperative contextual intention (button text that is what the user expects them to do);  The button name should be "Disconnect" if acting on that button will do so, but that itself is not very useful to convey information that there is a connection attempt or active connection currently. Anyways without a "Disconnect" action available that keeps the profile selected it needs consideration.
- Automatic connection should be configurable whether to do so, and per-device setting may override in the positive.
- Positioning device TX/RX counters (and maybe a "stale" timer for TX and for RX when the last change occurred) distinct from NTRIP TX/RX counters.
- "Instrument height" setting distinct from antenna PCO should be configured per-position device profile and as a general setting

Per-positioning device profile settings I think we would be overriding only the lack of turning on a feature or entering text input in the parent (Positioning) settings context. This should make sense in most of the settings, for example not entering an Antenna offset in the Bluetooth RTK Positioning device profile can default to the parent (Positioning) setting, while entering such a setting would (from the view of parent (Positioning) settings) be input-disabled but showing the value of that child setting;  the user may then edit or strike the override within that device profile and also if so determined to the user may select the "No device" profile to edit that general-level setting itself (since there are by definition no overrides in "No device") without disrupting the per-device profile override.

Features I note are present in SW Maps we may consider:
- Specific to bluetooth and perhaps Android, the connection history should be possible to select from so a configuration can be prepared "in the office" without the device present
-  "Instrument Model" where there are profiles for various common RTK devices, for example "SparkFun RTK" / "Generic NMEA" / "u-blox" / "Topcon GR3/GR5" / "Pentax G3100 R1/R2" / "ComNav" present in SW Maps as an option;   being closed source I have no idea what that does exactly, but it is notable that there would be a need for such options.
- Instrument height numeric value configuration parented above selection of Positioning device selection
- Multiple NTRIP profiles and selecting the actively connected profile, only one connection may active, the option whether or not to apply base antenna PCO sits parented above the profile selection.

Positives:
- Adding connection to positioning device is discoverable (with minor nit about contextual clues, above)
- Nice to see the selection of bluetooth in a list of many connection options among tcp, udp, and serial options
- Connection type selection of Bluetooth yields a prompt by Android asking for permission to access location, when confirmed there is no interruption the device is detected after a moment as is usually the situation with such bluetooth devices
-  Momentary confusion on selecting the detected bluetooth device but adapt quickly to understand that the checkmark at top of screen is in context with the "cancel" action. This appears consistent with recent Android design guide intent and not specific to QField.

Recommended: I would suggest the "Name" text input field appear above the "Connection type" drop-down selector although this may be a regional preference that I personally expect headings to be positioned above non-headings. The possible confusion is in assumption that the "Name" is a manual input required by the "Connection type" (such as a device name) when instead it is actually the Profile Name. When navigating "Connection type" selections and their parameters, the "Name" text input is an obstruction that needs to be avoided to get to the actual settings for that Connection type, it should be at the top (in my opinion).

gdt:
I guess the question is whether there are servers that are either 1) unreasonable or 2) have a good reason to need a high update rate.    I see the point of configurable, but it's work to do and cognitive load for the user, who given that this is GIS seems likely not to really understand the reasons and consequences (to be fair this is quite complicated and hard to figure out).

1s seems way too fast to me; the only point of the server knowing the client's location is to choose a base station, in 'nearest' single base mode, or for VRS/iMax etc. where the idea is to combine carrier phase measurements from a number of stations to form a synthetic close station.    I don't believe (contrary evidence welcome) that being even 500m from the VRS is much worse than being at it, and you just don't get that far in 30s.

So, I'd set it to 30s, and see if there are problems, and investigate them.

I just looked at https://ntrip-catalog.org/ and picked the first entry alphabetically, ACORN (CT, US).   I don't see any metadata about needing GGA, even though they offer VRS.  I read their FAQ, and it doesn't mention it either.

gdt:
I have posted a query to ntrip-catalog and gpsd-users@ about GGA frequency.

anathan:
We highly encourage all cllients to send GGA at least every 30 seconds and that this should be done by default. It does no harm to send it for single base networks and is absolutely required for any kind of VRS or network RTK service.

It is also best practice to send the GGA as soon as possible upon connection. Some servers hang up quickly if a GGA is not received promptly.

We own and operate one of the largest worldwide networks (@pointonenav) and appreciate you adding this support properly!

gdt:
@anathan I have seen comments (hexagon) about disconnection if GGA is not received in 30s, and it was vague about a disconnect after not receiving something with a nomimal interval of 30s (so triggering after 35s), or whether 30s is the timeout and you really need to send more often.    Can you explain if your service disconnects or stops working, and if so what the interval from last received GGA to failure is, perhaps both specified and current actual if different?

As for "no harm" to send it when not required, there are privacy concerns and people may wish to use single-base mountpoints on purpose to avoid this.   While I'm not arguing any particular provider behaves badly (e.g., by keeping records of the upstream GGA or disclosing them), open source privacy norms require this to be default off.  Sending location data off device is a bright line that may only be crossed with explicit user request.   Your website has a "do not share or sell my information button" and there's no easily findable privacy policy link -- which is a clue that your company discloses at least some information -- so this seems like an entirely valid concern.

Separately from the question of what your disconnect timers are, can you explain if there is a new decision about which base (for selecting closest single base), or if a new VRS/iMAX location is chosen, on each GGA?  Or is the GGA timer simply a proxy for "is this client still present"?   I genuinely do not understand why this should be so frequent.

src/core/positioning/ntripclient.cpp
```diff
@@ -0,0 +1,531 @@
...
+  mLogStream.writeRawData( data.constData(), data.size() );
+
+  mLogBlockCount++;
+  if ( mLogBlockCount % 10 == 0 )
+  {
+    mLogFile.flush();
+  }
+}
+
+void NtripClient::nmeaSentenceReceived( const QString &sentence )
```
bettellam, line 187: I would do the opposite: first I would check that the sentence is GGA and then I would trigger the check on epoch:
```
 // filter first — only GGA sentences are relevant
  const thread_local QRegularExpression rxSentence( u"^\\$([A-Z]{2})([A-Z]{3})"_s );
  const QRegularExpressionMatch match = rxSentence.match( sentence );
  if ( match.captured( 2 ) != QStringLiteral( "GGA" ) )
    return;

  // rate limit applies only to GGA
  const qint64 epoch = QDateTime::currentMSecsSinceEpoch();
  if ( mLastNtripGgaSent != 0 && ( epoch - mLastNtripGgaSent ) < 1000 )
    return;
  mLastNtripGgaSent = epoch;

  sendNmeaSentence( sentence );

```
## Related suggestion: GGA interval instead of boolean
Following up on the earlier comment on PR #7274 — now that `NtripSettings` is a proper structured object, this might be a good moment to consider replacing `forwardNmeaSentences` (bool) with a `ggaIntervalSeconds` (int):

```
ggaIntervalSeconds = 0   → never forward GGA (current false  behaviour)
ggaIntervalSeconds = N   → forward GGA at most every N seconds
```

This would give users explicit control over the interval, which is useful
for casters that accept GGA but prefer a slower update rate, and would make the rate-limiting logic in `nmeaSentenceReceived` map directly to the setting value.

src/core/positioning/ntripclient.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+    logData( data );
+  } );
+
+  connect( mSocket, &NtripSocket::errorOccurred, this, [this]( const QString &msg, bool isPermanent ) {
+    qInfo() << msg;
+    emit errorOccurred( msg, isPermanent );
+  } );
+
+  connect( mSocket, &NtripSocket::streamConnected, this, [this]() {
```
bettellam, line 91: Force immediate GGA sending upon connection with:
`mLastNtripGgaSent = 0;`

anathan:
Completely fair, it's just the technology requires an approximate position to do what it does. Some customer truncate or randomize the GGA as a way around the privacy concerns. 

For our service and most like ours, we hang up the tcp socket if we don't receive an initial packet within 60 seconds. Partly this is because in general single sided tcp connections and nearly indistinguishable from an unclean disconnect (very common with cellular users).

The update rate of the position is in theory only needed if you travel more than 20km.

And for what it's worth we absolutely do not monetize user positions data, we actually make it available to our users to help debug their systems and monitor their fleets. Our full policy is here:  https://pointonenav.com/privacy/ that said, I can imagine there are other folks with less clear intentions and respect the private by default stance, as long as it's clear that will just not work for a large class of services for legitimate technical reasons.

gdt:
Thanks - that all sounds reasonable.   So your "you should every 30s" is tuned against a 60s timeout, with a big margin.

I do understand the need for the server to have a position of some sort for VRS and NEAR.   Drifting off topic, but it would be interesting to see a paper on the increased errors of positions as a results of truncating to 0.1 degree.   My impression is that there's a surprising degree of iono/tropo non-uniformity and VRS with stations in multiple directions helps tremendously.

@nirvn I take away from this that every 30s is likely reasonable.

I also would like to see a "blur GGA", perhaps a setting k with rounding to 10^-k degrees in lat/lon.   But I don't think any other NTRIP implementation supports that, and it isn't fair to add that requirement onto this PR.

src/core/positioning/ntripclient.cpp
nirvn: @bettellam , I've tweaked the time check. It's still happening before the more costly GGA string match, but the mLastNtripGgaSent member is updated _only when a string has been sent_. 

With the latest APKs in this PR now, are you able to connect to the second NTRIP service you were trying out?

nirvn:
@anathan , if you want to give the builds here a try against your NTRIP service, I'd be curious to hear if it works fine.

nirvn:
> The text of checkmark box ui option layout "Forward NMEA sentences to NTR..." is all I see it doesn't wrap to the next line on a small display screen.

@eshattow , that was fixed. Thanks for sharing your opinion. I'm not going to address everything mentioned but it's good food for thoughts. Your enthusiasm is appreciated :)

Are you in a position to try this NTRIP functionality with devices you have at-hand? Any conformation GNSS model XXX working would be most welcome.

src/core/positioning/ntripclient.cpp
bettellam: **Receiver:** ZED-F9P + QField
**Receiver:** Unicore UM982 + Esp32 GNSS Manager
to compare results with the same antenna with a splitter.

**Network:** Connection to VRS/IMAX caster
**GGA Send:** 30 seconds (I consider this too high!)

The connection to the caster is not always active.
I need to constantly check if it is receiving RTCM. It would be helpful to move/replicate the ggasend/rtcmreceived data in the Position interface by adding the line: **RTCM: 123/45678** when Enable NTRIP corrections = true
Now I have poor satellite constellation positioning and I can't achieve RKT FIX
(inconsistent values ​​in the two GNSS receivers)
I'll try another test in 5-10 hours.

nirvn: @bettellam , last time you mentioned the Rete GNSS Veneto NTRIP network was simply not working. Is it at least working sometimes now? Or are you trying something else.

One thing comes to mind: ATM, we do not have an auto-reconnect if the connection to the caster is lost. I assume we should at least add some auto-reconnect attempts?

bettellam: For testing, I'm still using Veneto's NTRIP GNSS, and it's connecting now.
If there were an automatic reconnection feature, it would be great (3 attempts), but if there were something in the position interface, the user would immediately understand what's happening when RTCM is active.

nirvn: @bettellam , oh nice, so it went from "not connecting" to "connecting", that's progress :partying_face:  :wink: 

I like the idea of adding a feedback in the positioning information panel.

nirvn: @bettellam , regarding the comment on 30 sec being too high an interval, can you expand on that? 

I would like to avoid throwing too many settings if possible. If a higher frequency as actual benefits, I'm all ears.

bettellam: @nirvn
` if ( mLastNtripGgaSent != 0 && ( epoch - mLastNtripGgaSent ) < 30000 )`
is the current delay setting; it was previously 1000ms.
(25 km/h in 30 seconds, I moved 208 meters.)
I generally use 3-5 seconds as the GGA sending interval. I've never used such long intervals (30 seconds).
That's why I was doing a simultaneous test (F9P UM982).
To be more reliable, I'd have to do the test while moving (car or bicycle). I need to organize the various connections.

gdt: I do not understand the causal link between moving 200m in 30s and thinking the interval is too high.   The purpose of sending GGA is not to allow the server to track position.   It is merely to choose the nearest base (in a mode where one single base is chosen), or to initialize VRS/iMAX/etc.   We heard from someone from a corrections service and it seemed only really large (more tham km) position changes would cause anything to change.

Do you have any evidence that a caster/server is re-initializing VRS at a rate faster than 30s?   That sounds crazy to me, as I would expect the receiver to have to reinitilaize with the new base coordinates and carrier phase reference data.

If you mean that  particular caster breaks the connection when you send GGA every 30s, is that documented or observed?  It does not seem like reasonable behavior.

I do agree that there should be auto reconnect.   Perhaps no more frequently than 60s after the last connect attempt (so that the first one after a good connection is fast, and that it is moderate during network or caster failure).

bettellam: @gdt I'm not an NTRIP caster manager, I'm just a user, a client NTRIP.
If I lose connection, it's either a software issue or a server error/timeout.
What causes these discontinuities I'm experiencing? I don't know!
 [Italian  Pegaso Now network](https://www.spektra.it/servizi/servizi-di-posizionamento/ ) constantly changes VRS to keep the virtual station always close to the station point. We're talking about the Italian certified network for land registry operations: IGMI and the Revenue Agency (taxes and duties).

I think a value of 10" could be an acceptable value for both static (topography) and dynamic situations.
That's why I wanted to run some simultaneous comparative tests to understand what's happening.

bettellam: @nirvn Ciao :-)
**Test:** (Built from commit dd313734fb30c70047d1e1798c09fa53caf83365)
**Receiver:** ZED-F9P ---> QField
**Receiver:** Unicore UM982 ---> Esp32 GNSS Manager
to compare results with the same antenna with a splitter.
**Network:** connection to VRS/IMAX caster [retegnssveneto.cisas.unipd.it](https://retegnssveneto.cisas.unipd.it/Web/index.php)
**GGA sending:** 30 seconds
**Positioning:** static

**I haven't noticed any disconnections to the NTRIP caster.**
It would be nice to have the GGA/RTCM data counters in the Position interface (**NTRIP: 123/45678**). The user can immediately see if everything is working correctly, avoiding constant changes in the interface.
Please note that with **NTRIP enabled,** you will definitely need to add the value from NMEA GxGST -> **RMS** _(value of the pseudorange residuals; includes carrier phase residuals during periods of RTK (float) and RTK (fixed) processing)_ to the Position interface.
This **RMS**, combined with the accuracy values, is a very important indicator for identifying the presence of Multipath or a "noisy" measurement position.
But this will be a new issue once the NTRIP caster functions are completed.
PS: I still hope to have `int ntripGgaInterval.`
Bye ;-)

nirvn: @bettellam , that's great news, thanks for sharing. Other requests / information registered and not simply ignored :)

nirvn:
FYI, I've added a few additional mount point details when fetched from the NTRIP server source table:

image

That'll be useful to know which format is used by the server and insure the GNSS device is compatible with it.

eshattow:
"Enable NTRIP corrections" does not appear in the Positioning settings view until a connection attempt is made of the active Positioning device profile, though if adding a bluetooth connection as an active Positioning device profile there's not any way to connect if you don't have that bluetooth device fully configured; It is surely possible and reasonable to want to add a bluetooth Positioning device profile i.e. for "in the office" preparations ahead of field work.  I can for example add a TCP or UDP Positioning device profile and attempt to connect, then the NTRIP settings are visible.  I suggest simplify this so "Enable NTRIP corrections" is always visible as configuring the NTRIP connection settings does not logically (from UX) require a viable bluetooth connection or non-connection.

Additional observed inconsistency in typing a mount point:
1. Drop-down select a mount point.
2. Observe the "Source identifier: " text now represents your selection
3. Type in a valid mount point
4. The "Source identifier: " text has not changed. Did the mount point not update, or is the text not representative of the updated setting?
 
Not to the exclusion of any previous suggestion (you might want a "refresh" while the NTRIP client itself is not enabled, and simply "refresh" when the value has not changed does not constitute a need to turn off the state of the client, as yet);  The refresh-circle-arrow action adjacent following the drop-down list should overload as a confirmation of any typed mount point value. The user should be able to type in any "secret" mount point that is not returned in the index of mount points from the server;  I think this is not yet implemented?

Deciding whether to sort Mount points by distance should be user-configurable.  There could be an added circle-target iconography tri-state between "A-Z" sort, position cursor, and "locked-on-target" position cursor for the live location positioning. That would give some way to force the location sorting to be in the map area and useful thus for "in the office" without any Positioning device connected, ahead of fieldwork.

It would be an interesting flex to offer a map view zoom-to-feature on the location of a mount point when available; this action when not available would simply go to the map view cursor. Better overall would be an overview map of mount points represented on the map view. Although the index of mount points may not have location data, when the data connection is active there is (always? I think isn't that how it works?) location data and the user may want a quick zoom-to-feature to the map view of where the rtcm caster reports to be, and that should not materially be different enough from the value served in the index to avoid overloading the UI element to offer such a zoom-to-feature action.

Of course it would be also interesting to have a virtual layer of features with the caster locations but the useful of that depends on there being an index that serves these locations, and the device displays are small so having a map is not quite as friendly to workflow as having a drop-down with textual list of distances from the map cursor (locked to position or not).

gdt:
I'd like to suggest arriving at a minimum viable product and merging it, and leaving future enhancements for the future.  @eshattow has lots of great ideas, but it seems better to have something usable by most sooner than something super polished later.    Even having to type in the mount point without any help would still be awesome compared to where we are now!  (If my GNSS receiver didn't have its own NTRIP via hotspot, I'd be feeling this even more.  As it is I have been functional for years.)

eshattow:
@gdt writes:
> I'd like to suggest arriving at a minimum viable product and merging it, and leaving future enhancements for the future. @eshattow has lots of great ideas, but it seems better to have something usable by most sooner than something super polished later. Even having to type in the mount point without any help would still be awesome compared to where we are now! (If my GNSS receiver didn't have its own NTRIP via hotspot, I'd be feeling this even more. As it is I have been functional for years.)

Agreed, don't let me distract you :-)

nirvn:
@eshattow , I _think_ the latest APK should fix the out of sync mount point details when manually typing in text in the editable combo box. I've also added a slightly grayed out distance label when opening the combo box popup, which I think helps understanding the distance ordering.

eshattow:
> @eshattow , I _think_ the latest APK should fix the out of sync mount point details when manually typing in text in the editable combo box. I've also added a slightly grayed out distance label when opening the combo box popup, which I think helps understanding the distance ordering.

Typing in the full mount point was picked up, though typing in e.g. all but the last character and view the drop-down initially it was not warping to that item in the list;  later I try again and it is warping (or was already selected somehow) to that item in the list but making that same selection does not complete the text entry so nothing happens. I expect to be typing as much of the name that I would like and then using the drop-down warps to where the nearest matching item in the sort is to allow the user to complete the selection (either what is highlighted to confirm the text in the box or, something near it). Important is not to automatically select any mount point on the user having just expanded the drop-down view, if possible the warp to nearest matching item may be highlighted or visual cue but it is not necessary, and the user should have to explicitly confirm any highlighted warp to item to actually select it.

Distance sort is too much surprising UX complexity for the initial implementation as-is. It is frustrating because I really like the novelty and potential usefulness however it gets in the way of what I'm doing and there's no exposed way to avoid it. For minimum viable product just show the distance, when available, in the list items and table the user-visible sort on distance for a future iteration?  The implementation of distance sort would need to give a low friction way to set the location cursor the distance is measured from and/or have an artificial position method which is relative to the selected mount point location, and some mode that is alphabetic sort even when there is position information available.

src/core/positioning/ntripsettings.cpp
```diff
@@ -0,0 +1,115 @@
...
+         mMountPoint == other.mMountPoint &&
+         mUsername == other.mUsername &&
+         mPassword == other.mPassword;
+  // clang-format on
+}
+
+bool NtripSettings::isValid() const
+{
+  return mPort > 0 && !mHost.isEmpty() && !mMountPoint.isEmpty() && !mUsername.isEmpty() && !mPassword.isEmpty();
+}
```
MidnightCoder-m, line 40: `isValid()` requires both username and password to be non-empty. (I am not sure if there is any) NTRIP casters that support anonymous access (no credentials needed)?

This prevents users from connecting to such casters.

src/core/positioning/ntripsourcetablefetcher.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+QList<NtripMountPoint> NtripSourceTableFetcher::parseSourceTable( const QByteArray &data ) const
+{
+  QList<NtripMountPoint> mountPoints;
+
+  const int strBegin = data.indexOf( "STR;" );
+  if ( strBegin >= 0 )
+  {
+    const QByteArray body = data.mid( strBegin + 4 );
```
MidnightCoder-m: Shouldn't this be `data.mid( strBegin )` instead of `data.mid( strBegin + 4 )`? The current code skips the first `STR;` (?!)

src/core/positioning/ntripclient.cpp
```diff
@@ -0,0 +1,532 @@
...
+  if ( mSocket->isOpen() )
+  {
+    mSocket->abort();
+  }
+  mHeadersSent = false;
+}
+
+void NtripSocket::onConnected()
+{
+  qInfo() << QStringLiteral( "Connected to NTRIP caster:  host %1, port %2, mounnt point %3" ).arg( mHost, QString::number( mPort ), mMountPoint );
```
MidnightCoder-m: Small typo here `Connected to NTRIP caster: host %1, port %2, mount point %3`

src/core/positioning/ntripclient.cpp
```diff
@@ -0,0 +1,532 @@
...
+        connect( nmeaReceiver, &NmeaGnssReceiver::nmeaSentenceReceived, this, &NtripClient::nmeaSentenceReceived );
+      }
+    }
+  }
+
+  mLastNtripGgaSent = 0;
+  mBytesSent = 0;
+  mBytesReceived = 0;
+
+  qInfo() << QStringLiteral( "Starting NTRIP client: host %1, port %2, mounnt point %3" ).arg( ntripSettings.host(), QString::number( ntripSettings.port() ), ntripSettings.mountPoint() );
```
MidnightCoder-m: Same typo

src/core/positioning/abstractgnssreceiver.h
```diff
@@ -31,16 +31,19 @@ class AbstractGnssReceiver : public QObject
   public:
     enum Capability
     {
-      NoCapabilities = 0,      //!< No capabilities
-      OrthometricAltitude = 1, //!< Orthometric altitude support
-      Logging = 1 << 1,        //!< Logging support
+      NoCapabilities = 0,       //!< No capabilities
+      OrthometricAltitude = 1,  //!< Orthometric altitude support
+      Logging = 1 << 1,         //!< Logging support
+      NtripCorrection = 1 << 2, //!< Logging support
```
MidnightCoder-m: `Logging support` should change to `NTRIP correction support` right?

src/core/positioning/bluetoothreceiver.cpp
```diff
@@ -247,3 +255,67 @@ void BluetoothReceiver::pairingFinished( const QBluetoothAddress &address, QBlue
...
 }
+
+void BluetoothReceiver::onCorrectionDataReceived( const QByteArray &data )
+{
+  if ( !mSocket || !mSocket->isOpen() )
+  {
+    return;
+  }
+
+  if ( mAddress.startsWith( "C8:47:8C" ) )
```
MidnightCoder-m: Could this be made configurable? or at least documented with a comment explaining the rationale?

MidnightCoder-m (commented):
Great work on this PR! The NTRIP implementation looks solid and well-structured overall. I'm not an expert in this area, so I just have a few questions/comments

src/core/positioning/bluetoothreceiver.cpp
eshattow: such as "detect receiver model" on/off and a selection list of receiver models (including "generic" catch-all). I suspect this may be related to how SW Maps has a list of possible receiver models, are there model-specific data manipulations that need to be processed?

src/qml/PositioningNtripSettings.qml
```diff
@@ -0,0 +1,395 @@
...
+                  Layout.preferredWidth: distanceLabelMetrics.width
+                  Layout.alignment: Qt.AlignRight | Qt.AlignVCenter
+                  visible: text !== ''
+                  font: ntripMountPointComboBox.font
+                  color: Theme.secondaryTextColor
+                  horizontalAlignment: Text.AlignRight
+                  text: {
+                    if (ntripMountPointComboBox.popup.visible && positionSource.positionInformation.latitudeValid) {
+                      const pos = GeometryUtils.point(positionSource.positionInformation.longitude, positionSource.positionInformation.latitude);
+                      return GeometryUtils.formattedDistanceBetweenPoints(model.point, pos, CoordinateReferenceSystemUtils.wgs84Crs());
```
bettellam: // Web Mercator meter
return GeometryUtils.formattedDistanceBetweenPoints(model.point, pos, QgsCoordinateReferenceSystem.fromEpsgId(3857));

You could use the project's CRS, but if it were still a geographic CRS, the metric distance value would not be directly readable.

src/qml/PositioningNtripSettings.qml
```diff
@@ -0,0 +1,395 @@
...
+            if (idx > -1) {
+              if (ntripMountPointComboBox.model[idx].identifier !== '') {
+                details.push(qsTr("Identifier: ") + ntripMountPointComboBox.model[idx].identifier);
+              }
+              if (ntripMountPointComboBox.model[idx].format !== '') {
+                details.push(qsTr("Data format: ") + ntripMountPointComboBox.model[idx].format);
+              }
+              if (positionSource.positionInformation.latitudeValid) {
+                const pos = GeometryUtils.point(positionSource.positionInformation.longitude, positionSource.positionInformation.latitude);
+                const distance = GeometryUtils.formattedDistanceBetweenPoints(ntripMountPointComboBox.model[idx].point, pos, CoordinateReferenceSystemUtils.wgs84Crs());
```
bettellam: // Web Mercator meter
const distance = GeometryUtils.formattedDistanceBetweenPoints(ntripMountPointComboBox.model[idx].point, pos, QgsCoordinateReferenceSystem.fromEpsgId(3857));

You could use the project's CRS, but if it were still a geographic CRS, the metric distance value would not be directly readable.

src/core/positioning/ntripsourcetablefetcher.cpp
nirvn: Good catch

PR #7579, opened 2026-06-15
Added qml tests for the vertex editor geometry tool

kaustuvpokharel:
This pr adds a qml test file for the vertex editor tool. The vertex model itself is already covered in test_vertexmodel.cpp, so this does not retest that. It focuses on the tool layer around it: init wiring, applyChanges and the autoSave branch, cancel, remove, and undo.

The tool reads a few objects from its parent scope in the app (qfieldSettings, coordinateLocator, etc). I provide them in the test the same way tst_featureForm.qml does, as plain in-scope items. mapSettings is a real MapSettings set to the Fields layer CRS so the vertex model transform is a no-op. Tests run against the existing Fields (polygon) and Tracks (line) features in test_bees.qgz, and never commit, the edits go to the layer edit buffer and are rolled back after each test so the shared data stays clean.

>canvasClicked is left out on purpose. It turns a screen point into a map coordinate through mapSettings.screenToCoordinate, which only means something with a real rendered canvas. Without one the screen to map math is meaningless, so a click test there would just assert made up numbers. That path fits an integration (spix) test better, is what I thought, if we need it.
I did not test the remove-down-to-minimum case. The polygon has 12 real vertices so it would take about 8 removals to trip canRemoveVertex false, which could endup fragile. Instead canRemoveVertex is checked from both sides: true on a removable vertex in edit mode, false outside edit mode. That covers the guard without exhausting the ring. I also left out the save-failure toast branch. Forcing featureModel.save() to return false means making the layer non-editable mid-operation, which is fragile and risks leaving the shared layer dirty. It is a thin error branch and not worth the check, let me know if you think otherwise @nirvn . All tests pass es locally.

### DIFF
--- a/test/qml/tst_geometryeditor_vertex.qml
+++ b/test/qml/tst_geometryeditor_vertex.qml
@@ -105,18 +105,74 @@ TestCase {
     compare(vertexEditor.blocking, true);
   }
 
+  // returns how many points in list a are not present (within tolerance) in
+  // list b. Comparing both directions makes this independent of vertex order or
+  // ring rotation, which a vertex move can introduce.
+  function pointsNotIn(a, b) {
+    let missing = 0;
+    for (let i = 0; i < a.length; ++i) {
+      let found = false;
+      for (let j = 0; j < b.length; ++j) {
+        if (Math.abs(a[i].x - b[j].x) < 0.001 && Math.abs(a[i].y - b[j].y) < 0.001) {
+          found = true;
+          break;
+        }
+      }
+      if (!found)
+        missing++;
+    }
+    return missing;
+  }
+
+  // reads the vertices out of a geometry's WKT as a plain array of points, so
+  // tests can compare geometries vertex by vertex. Uses asWkt only, which is the
+  // same accessor the other tests rely on. For a polygon ring the closing vertex
+  // repeats the first one, so we drop that duplicate to count each vertex once.
+  function geometryPoints(geometry) {
+    const points = [];
+    const wkt = geometry.asWkt(6);
+    // pull the coordinate pairs from inside the parentheses
+    const inner = wkt.substring(wkt.indexOf("("), wkt.lastIndexOf(")") + 1);
+    const pairs = inner.replace(/[()]/g, " ").trim().split(",");
+    for (let i = 0; i < pairs.length; ++i) {
+      const nums = pairs[i].trim().split(/\s+/);
+      if (nums.length >= 2 && nums[0] !== "") {
+        points.push({
+          x: parseFloat(nums[0]),
+          y: parseFloat(nums[1])
+        });
+      }
+    }
+    // drop a repeated closing vertex (polygons repeat the first point at the end)
+    if (points.length > 1) {
+      const first = points[0];
+      const last = points[points.length - 1];
+      if (Math.abs(first.x - last.x) < 0.001 && Math.abs(first.y - last.y) < 0.001) {
+        points.pop();
+      }
+    }
+    return points;
+  }
+
   function test_applyChangesAppliesMovedVertexToLayerBuffer() {
     const model = makeFieldsModel();
     vertexEditor.init(model, mapSettingsItem, null, null);
 
+    // capture the polygon before the edit
+    const before = geometryPoints(model.feature.geometry);
+
+    // move the first real vertex and apply. we never commit, so the layer file
+    // is untouched.
     selectAndMoveFirstVertex();
-    const before = model.feature.geometry.asWkt();
-
-    // apply pushes the vertex model edit onto the feature. we check the feature
-    // geometry changed but never commit, so the layer file is untouched
     vertexEditor.applyChanges(true);
+    const movedGeom = model.feature.geometry;
+    verify(!movedGeom.isNull);
+    compare(movedGeom.type, Qgis.GeometryType.Polygon);
 
-    verify(model.feature.geometry.asWkt() !== before);
+    const after = geometryPoints(movedGeom);
+    compare(after.length, before.length);
+    compare(pointsNotIn(before, after), 1);
+    compare(pointsNotIn(after, before), 1);
   }
 
   function test_applyChangesWithAutoSaveOffDoesNotApply() {
@@ -168,14 +224,16 @@ TestCase {
 
     vertexModel.editingMode = VertexModel.EditVertex;
     vertexModel.currentVertexIndex = 1;
-    // the polygon has 12 real vertices, well above the minimum, so removing one is allowed
+    // the polygon has plenty of real vertices, well above the minimum, so
+    // removing one is allowed
     verify(vertexModel.canRemoveVertex);
     const countBefore = vertexModel.vertexCount;
 
     vertexModel.removeCurrentVertex();
 
-    // removing a vertex drops the row count and marks the edit dirty
-    verify(vertexModel.vertexCount < countBefore);
+    // removing one existing vertex drops it and its paired segment candidate, so
+    // the row count goes down by exactly two, and the edit is marked dirty
+    compare(vertexModel.vertexCount, countBefore - 2);
     compare(vertexModel.dirty, true);
   }
 
@@ -222,6 +280,8 @@ TestCase {
     vertexEditor.init(model, mapSettingsItem, null, null);
     const vertexModel = model.vertexModel;
 
+    const before = geometryPoints(model.feature.geometry);
+
     // on a line the first real vertex is also at an odd index
     vertexModel.editingMode = VertexModel.EditVertex;
     vertexModel.currentVertexIndex = 1;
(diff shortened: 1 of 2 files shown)


qfield-fairy:
Download a [MacOS DMG universal build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/28979717986/qfield-dmg.zip).
*(Built from commit 701edf93cb50398df21f54dd36dfb4e6f3e0bfc3)*




### 🪟 Windows builds
Download a [Windows build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/28979717970/qfield-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-windows-x64.zip) and for [arm64](https://nightly.link/opengisch/QField/actions/runs/28979717970/qfield-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-windows-arm64.zip).

*(Built from commit 701edf93cb50398df21f54dd36dfb4e6f3e0bfc3)*




### 🐧 Linux AppImage builds
Download a [Linux AppImage build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/28979717985/qfield-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-linux-x64.AppImage.zip).
*(Built from commit 701edf93cb50398df21f54dd36dfb4e6f3e0bfc3)*




### 📱 Android builds
Download an [Android arm64 build of this PR for testing](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-arm64.zip).
*(Built from commit 701edf93cb50398df21f54dd36dfb4e6f3e0bfc3)*
<details>

- [arm-neon](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-arm-neon.zip)
- [x64](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-x64.zip)
- [x86](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-x86.zip)
- [**all access arm64**](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield_all_access-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-arm64.zip)
- [all access arm-neon](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield_all_access-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-arm-neon.zip)
- [all access x64](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield_all_access-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-x64.zip)
- [all access x86](https://nightly.link/opengisch/QField/actions/runs/28979717999/qfield_all_access-dev-7579-2fc5c9e2c31148d9f57fb2e8d39855a223eafde4-android-x86.zip)
</details>

test/qml/tst_geometryeditor_vertex.qml
```diff
@@ -0,0 +1,265 @@
...
+    vertexEditor.init(model, mapSettingsItem, null, null);
+
+    selectAndMoveFirstVertex();
+    const before = model.feature.geometry.asWkt();
+
+    // apply pushes the vertex model edit onto the feature. we check the feature
+    // geometry changed but never commit, so the layer file is untouched
+    vertexEditor.applyChanges(true);
+
+    verify(model.feature.geometry.asWkt() !== before);
```
nirvn: Can we be more deterministic about this. After !== before doesn't test much other than "something changed". If the operation wrongly killed the geometry, that would still be valid.

The idea here is to check the resulting geometries against an expected geometry, not merely that it changed.

test/qml/tst_geometryeditor_vertex.qml
```diff
@@ -0,0 +1,265 @@
...
+    vertexModel.editingMode = VertexModel.EditVertex;
+    vertexModel.currentVertexIndex = 1;
+    // the polygon has 12 real vertices, well above the minimum, so removing one is allowed
+    verify(vertexModel.canRemoveVertex);
+    const countBefore = vertexModel.vertexCount;
+
+    vertexModel.removeCurrentVertex();
+
+    // removing a vertex drops the row count and marks the edit dirty
+    verify(vertexModel.vertexCount < countBefore);
```
nirvn: Again, compare expected values, not merely != or <

test/qml/tst_geometryeditor_vertex.qml
```diff
@@ -0,0 +1,265 @@
...
+    vertexModel.editingMode = VertexModel.EditVertex;
+    vertexModel.currentVertexIndex = 1;
+    const point = vertexModel.currentPoint;
+    vertexModel.currentPoint = GeometryUtils.point(point.x + 5, point.y + 5);
+    compare(vertexModel.dirty, true);
+
+    const before = model.feature.geometry.asWkt();
+    vertexEditor.applyChanges(true);
+
+    verify(model.feature.geometry.asWkt() !== before);
```
nirvn: Same

test/qml/tst_geometryeditor_vertex.qml
kaustuvpokharel: Thankyou @nirvn for the review, I reworked to assert expected outcomes, but the two move tests, asserting an exact resulting coordinate turned out to be unsafe like applying the edit rotates the polygon ring, and the moved vertex doesnt land at a simple offset of its map-coordinate input, so a hardcoded coordinate or full wkt was both sensitive to test order. So, i went over and they now check the result is the same geometry type, has the same vertex count, and differs from the original by exactly one vertex (one point removed, one added), compared as sets so ring rotation doesnt matter. That rules out the geometry being emptied or corruptd while staying deterministic

test/qml/tst_geometryeditor_vertex.qml
```diff
@@ -0,0 +1,335 @@
...
+    // the vertex model geometry reflects the move directly, one vertex relocated
+    const moved = vertexModel.geometry;
+    verify(!moved.isNull);
+    compare(moved.type, Qgis.GeometryType.Line);
+
+    const after = geometryPoints(moved);
+    compare(after.length, before.length);
+    compare(pointsNotIn(before, after), 1);
+    compare(pointsNotIn(after, before), 1);
+  }
```
nirvn: Why are you not comparing WKT vs WKT here? Again, this kind of comparison doesn't test much other than one point being different (doesn't tell us if the _right_ point has been moved and whether it's been moved where we want it to be).

What issue you are encountering here? I'm perplexed :)

test/qml/tst_geometryeditor_vertex.qml
```diff
@@ -0,0 +1,335 @@
...
+    // drop a repeated closing vertex (polygons repeat the first point at the end)
+    if (points.length > 1) {
+      const first = points[0];
+      const last = points[points.length - 1];
+      if (Math.abs(first.x - last.x) < 0.001 && Math.abs(first.y - last.y) < 0.001) {
+        points.pop();
+      }
+    }
+    return points;
+  }
```
nirvn: Was that taken from an LLM?

test/qml/tst_geometryeditor_vertex.qml
kaustuvpokharel: I actually did wkt vs wkt, this was my secondary idea, i compare it agaist the geometry after the applyChanges, and the path ran the geometry through buffer, makeValid, avoid intesections and so on, I saw the it was reoredering the ring start point and even changed the vertex count, so wkt compare was failing rather than on the move itself. I than thought this would be one way to be more determistic approach for checking the points, true it proves the point differes, not right vertex move
I am still hesitant on right approach, because I understand what you actually trying me to test here which is exact value, that is to supposedly should be to assert the vertex model directly, before applyChanges, I tried this just now and it works. That geometry is stable and index aligned, so I can select the known vertex, and move it by a fixed offeset and compare the resulting wkt to the expected one directly, which should check the right point moved to the right place. If this is right approach, what do you think @nirvn ? I will keep it on the vertex model geometry, not in the saved feature, since the save path mangles the geometry and commits to the shared test layer, please let me know how I should see this?

test/qml/tst_geometryeditor_vertex.qml
kaustuvpokharel: no, it was me working around the ring rotation I hit after applyChanges because I took this approach and not with wkt vs wkt, parsing it was the point of having it

nirvn (approved):

